PR #1, opened 2019-04-04
Add return on visit method, making possible the usage of promise.

peteleco:
Ex.:
Inertia.visit(url).then(() => alert('Do something after the page is loaded.'))


reinink:
Dig it

PR #13, opened 2019-04-17
Added array length check to navigationType

### LINKED ISSUE #12: Reloading in Safari
Tested in Safari 12.03 & 12.1

The initial page loads fine, but when you refresh, it throws the following error:

```
"TypeError: undefined is not an object (evaluating 'window.performance.getEntriesByType("navigation")[0].type')"
```

I added a length check on the array in inertia.js line 28 and it seems to work. Happy to open a PR.

scottzirkel:
Fixes #12

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -24,7 +24,7 @@ export default {
   },
 
   navigationType() {
-    if (window.performance) {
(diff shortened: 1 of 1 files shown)


reinink:
Thanks @scottzirkel!!

Issue #9, opened 2019-04-16
Allow custom loading indicators


reinink:
Currently Inertia uses [nprogress](http://ricostacruz.com/nprogress/) for its loading indicator. However, it would be nice to support other libraries. We'd simply have to expose this as a configuration option somehow.

For example:

```js
new Vue({
  render: h => h(Inertia, {
    props: {
      component: app.dataset.component,
      props: JSON.parse(app.dataset.props),      
      resolveComponent: (component) => { ... },
      beforeVisit: () => { /* start loading indicator */ },
      afterVisit: () => { /* stop loading indicator */ },
    },
  }),
}).$mount(app)
```

reinink:
I actually think a better approach is to allow Inertia level customization directly via Inertia, and not through the adapters.

```js
Inertia.beforeVisit(() => {}),
Inertia.afterVisit(() => {}),
Inertia.startProgress(() => {}),
Inertia.incrementProgress(() => {}),
Inertia.stopProgress(() => {}),
```

MichaelDeBoey:
I do think we should also support this in some way or another via the adapters, since those are really a component mapping over the Inertia API

reinink:
I'm not sure I agree. There is nothing wrong with using Inertia directly (ie. `Inertia.visit()`), and we're really just talking about configuration here. To make every adapter responsible for this just means more work at the adapter level.

MichaelDeBoey:
Just seems a bit weird to have the following code then:

```js
import InertiaVue, { Inertia } from 'inertia-vue'
import Vue from 'vue'

Inertia.beforeVisit(() => {})
Inertia.afterVisit(() => {})
Inertia.startProgress(() => {})
Inertia.incrementProgress(() => {})
Inertia.stopProgress(() => {})

Vue.use(InertiaVue)

const app = document.getElementById('app')

new Vue({
  render: h => h(InertiaVue, {
    props: {
      initialPage: JSON.parse(app.dataset.page),
      resolveComponent: name => import(`@/Pages/${name}`).then(module => module.default),
    },
  }),
}).$mount(app)

```

dbpolito:
This is something i would love to see, in my case i customised the CSS as a workaround but i would love to use something built-in i already have on my project.

Any ideas on how to solve this? I would love to contribute on this one.

Are we going with something simpler as https://github.com/inertiajs/inertia/issues/9#issuecomment-493676180 or https://github.com/inertiajs/inertia/issues/77?

[...]

PR #15, opened 2019-04-22
Add beforeVisit and afterVisit hooks

thoresuenert:
To be able to overwrite nprogess as loading indicator as mentioned in https://github.com/inertiajs/inertia/issues/9  this PR moves nprogess to `beforeVisit` and  `afterVisit` hooks and make it accessible through `Inertia.init`

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -44,9 +44,17 @@ export default {
     clearInterval(this.progressBar)
   },
 
+  beforeVisit() {
+    this.showProgressBar()
+  },
+
+  afterVisit() {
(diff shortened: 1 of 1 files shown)


thoresuenert:
maybe there is a use case to add both hooks independently from each other. for replacing nprogress both hooks have to be present.

reinink:
As noted in your Vue adapter PR ([#18](https://github.com/inertiajs/inertia-vue/pull/18)), I think I'm going to do this a little different, but thanks so much for the PR either way!

PR #19, opened 2019-05-02
Replace axios with fetch ponyfill

bradlc:
Hey @reinink and co. I’ve been checking out Inertia recently and really liking it!

One thing that I’d love to see is for the bundle cost to come down a little bit 🔬. I’d like to suggest replacing `axios` with something more lightweight: [`unfetch-abortable`](https://github.com/bradlc/unfetch-abortable) is a fork of [`unfetch`](https://github.com/developit/unfetch/) with support for [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).

To compare the size of the options I created a UMD bundle for each, using [`microbundle`](https://github.com/developit/microbundle):

```
microbundle --entry src/index.js --format umd --external none

With axios
----
inertia.umd.js       19408 B
inertia.umd.js.gz     6959 B
inertia.umd.js.br     6193 B


With unfetch-abortable
----
inertia.umd.js       10072 B (-48%)
inertia.umd.js.gz     3776 B (-46%)
inertia.umd.js.br     3306 B (-47%)
```

It’s worth noting that I have used `unfetch-abortable` for wider browser support. If you remove the import (`src/inertia.js` line 1) the rest of the code works great in browsers that natively support `fetch` and `AbortController` 🎉

Is this something you would be open to? Let me know what you think. Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
   "author": "Jonathan Reinink",
   "license": "MIT",
   "dependencies": {
-    "axios": "^0.18.0",
     "nprogress": "^0.2.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
   "author": "Jonathan Reinink",
   "license": "MIT",
   "dependencies": {
-    "nprogress": "^0.2.0"
+    "nprogress": "^0.2.0",
+    "unfetch-abortable": "0.0.1"
   },
   "devDependencies": {
     "eslint": "^5.15.3"
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -75,7 +75,9 @@ export default {
         ...this.version ? { 'X-Inertia-Version': this.version } : {},
       },
     }).then(response => {
-      if (this.isInertiaResponse(response)) {
+      if (response.status === 409 && response.headers.has('x-inertia-location')) {
+        return this.hardVisit(true, response.headers.get('x-inertia-location'))
+      } else if (this.isInertiaResponse(response)) {
         return response.json()
       } else {
         response.text().then(data => this.showModal(data))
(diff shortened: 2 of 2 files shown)


MichaelDeBoey:
@bradlc I don't think it's save to use `unfetch-abortable` at this moment, since you published it to npm with changed code from the original repo (otherwise you wouldn't make a new npm package), but without a reference to it's own repo where the changed code is visible.

So your suggestion _could_ be a fraud.  
I'm not saying it is, but I think you understand that right now there's no real way in telling so.

bradlc:
Hey @MichaelDeBoey – I _did_ reference the repo. It's here incase anybody missed it: https://github.com/bradlc/unfetch-abortable

And you can see it on unpkg here: https://unpkg.com/unfetch-abortable/

sebastiandedeyne:
While I like the idea of dropping Axios to reduce the bundle size, why does the polyfill need to be part of the core bundle? 

Why can't Inertia just use `fetch` and `AbortController`, and have developers polyfill where necessary instead of forcing a spefic implementation?

bradlc:
> While I like the idea of dropping Axios to reduce the bundle size, why does the polyfill need to be part of the core bundle?

It doesn't! This PR is really just a jumping-off point for discussion. As mentioned in the original comment, the changes to the `visit` function make it compatible with browser-native `fetch` and `AbortController`

bradlc:
Native sounds great to me!

You could even provide a way for users to inject their own `fetch`/`AbortController` implementations. Something like:

```js
<Inertia fetch={myFetch} ... />
```

I personally prefer this [ponyfill](https://ponyfill.com)-style method versus asking users to polyfill stuff directly (`window.fetch = myFetch`) – you would still be able to do it that way, of course. What do you all think of something like this?

I will update this PR (or create a new one) tomorrow, to remove the polyfill.

MichaelDeBoey:
I'm not really in favor of letting the user inject their own `fetch`,  since that's really an implementation detail the library has I think

PR #20, opened 2019-05-05
visit: return setPage

bradlc:
👋 This PR adds a `return` at the end of a `visit` so that the promise resolves only once navigation is complete.

Consider the following example:

```js
this.$inertia.visit('/').then(() => {
  console.log('Navigated!')
})
```

### Before

The `visit` promise resolves before the navigation is complete:

Kapture 2019-05-05 at 19 24 17

### After

The `visit` promise resolves once the navigation is complete:

Kapture 2019-05-05 at 19 32 04

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -85,7 +85,7 @@ export default {
       if (page) {
         this.version = page.version
         this.setState(page, replace)
-        this.setPage(page).then(() => {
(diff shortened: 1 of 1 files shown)


reinink:
Nice catch @bradlc! I swear we had this working properly at one point, but yeah, there's been lots of refactoring since. 😂

PR #21, opened 2019-05-06
Fix reload signature

MichaelDeBoey:
The `url` variable isn't used and both  the [`inertia-vue` docs](https://github.com/inertiajs/inertia-vue/blame/master/readme.md#L150) and the [`inertia-react` docs ](https://github.com/inertiajs/inertia-react/blame/master/readme.md#L152)reference the `reload` method with only an `options` object.


reinink:
Thanks @MichaelDeBoey!

PR #24, opened 2019-05-09
Set the page before hiding the progress bar

bradlc:
Hi again. This is a quick fix for an issue with the progress bar where it is hiding before navigation is complete.

### Before

Kapture 2019-05-09 at 22 23 39

### After

Kapture 2019-05-09 at 22 25 17

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -82,13 +82,15 @@ export default {
         return Promise.reject(error)
       }
     }).then(page => {
-      this.hideProgressBar()
       if (page) {
         this.version = page.version
         this.setState(page, replace)
         return this.setPage(page).then(() => {
           this.setScroll(preserveScroll)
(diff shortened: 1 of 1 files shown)


reinink:
Interesting. I can't reproduce this locally. What browser are you using?

bradlc:
Firefox and Chrome. Makes sense because in the original code `this.hideProgressBar` is being called before `this.setPage` has resolved.

Try throttling your connection in devtools if you haven’t already.

reinink:
Right! For any assets that need to be loaded yet when using code splitting.

reinink:
Thanks @bradlc! 🤟

bradlc:
No worries! 🎉 Yeah, it’s really easy to miss on a fast connection!

PR #25, opened 2019-05-09
Fix unwanted intermediate pages

bradlc:
This is a tricky one! 🤔 Might need a bit of discussion

First of all here is the problem in action:

Kapture 2019-05-09 at 22 49 31

### Steps to reproduce:

1. Open [pingcrm](https://github.com/inertiajs/pingcrm) in Firefox
2. Click on `Organizations`
3. As soon as the address bar changes to `/organizations` click on `Contacts`
4. See that the Organizations page is visible briefly

> **Note:** Weirdly, I cannot reproduce this in Chrome. Also, you may need to throttle the network to see what’s going on

### Why does this happen?

When `visit` is called any inflight request is cancelled using an axios cancel token, causing the previous `visit` to reject. All good! **But this only cancels the initial request to the backend.** If we initiate a new visit _after_ the axios request has resolved but _before_ `setPage` has resolved then the initial visit will still resolve successfully and render. In theory you could even end up on the first page that you clicked on if it takes longer to resolve than the second!

### What’s the fix?

There’s probably a few ways to fix this, and this might not be the best so other ideas are welcome.

~~When `visit` is called an object is created (`visitId`) that acts as a unique ID for the visit. This is also stored against `this.visitId`. We can then compare `visitId` to `this.visitId` to determine if the current visit is the most recent.~~

> ~~**Note:** there are a few ways this could be done. For example, instead of an object we could use the axios promise, or a timestamp, as a unique ID~~

I have updated the PR to include a wrapper around `setPage` instead, so it covers `restoreState` as well. Each `setPage` call has a token which is compared against the current "active" token to determine if the component should update or not.

> **Note:** I don’t think `token` is a good name for this but couldn’t think of anything else

Once `this.setPage` is called it’s out of our hands. That goes off, resolves the component, and renders it. So we pass a second argument, `shouldUpdate`, which can be used to verify that it’s still ok to render. There is a [companion PR](https://github.com/inertiajs/inertia-vue/pull/28) which implements this for `inertia-vue`.

Any thoughts?

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -52,6 +52,9 @@ export default {
 
     this.cancelToken = axios.CancelToken.source()
 
+    let visitId = {}
+    this.visitId = visitId
+
     return axios({
       method: method,
       url: url,
@@ -85,9 +88,11 @@ export default {
       if (page) {
         this.version = page.version
         this.setState(page, replace)
-        return this.setPage(page).then(() => {
-          this.setScroll(preserveScroll)
-          this.hideProgressBar()
+        return this.setPage(page, () => visitId === this.visitId).then(didUpdate => {
+          if (didUpdate) {
+            this.setScroll(preserveScroll)
+            this.hideProgressBar()
+          }
         })
       } else {
         this.hideProgressBar()
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -10,7 +10,9 @@ export default {
 
   init(page, setPage) {
     this.version = page.version
-    this.setPage = setPage
+    this.setPage = (page, token = this.createToken()) => {
+      return setPage(page, () => token === this.token)
+    }
 
     if (window.history.state && this.navigationType() === 'back_forward') {
       this.setPage(window.history.state)
@@ -42,6 +44,12 @@ export default {
     clearTimeout(this.progressBar)
   },
 
+  createToken() {
+    let token = {}
+    this.token = token
+    return token
+  },
+
   visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false } = {}) {
     this.hideModal()
     this.showProgressBar()
@@ -52,8 +60,7 @@ export default {
 
     this.cancelToken = axios.CancelToken.source()
 
-    let visitId = {}
-    this.visitId = visitId
+    let token = this.createToken()
 
     return axios({
       method: method,
@@ -88,7 +95,7 @@ export default {
       if (page) {
         this.version = page.version
         this.setState(page, replace)
-        return this.setPage(page, () => visitId === this.visitId).then(didUpdate => {
+        return this.setPage(page, token).then(didUpdate => {
           if (didUpdate) {
             this.setScroll(preserveScroll)
             this.hideProgressBar()


reinink:
Hmm, dark problem.

Question. Does this PR include everything? It's not making much sense to me at all. I don't see a unique token being generated anywhere. I also don't know why you're calling `this.setState` twice in a row on line 97 and 98. I feel like something is missing.

bradlc:
@reinink Sorry where’s that? I don’t see multiple `this.setState` calls:

https://github.com/inertiajs/inertia/blob/62231b4228d2ae35ef3a7956081f8e71742c3862/src/inertia.js#L97-L98

The "unique token" is the **object reference**:

https://github.com/inertiajs/inertia/blob/62231b4228d2ae35ef3a7956081f8e71742c3862/src/inertia.js#L48

I appreciate that this isn’t 100% clear. Imagine if the above line was instead `let token = Date.now()` (which is totally could be). The key is that **`token` is scoped to each "visit" and `this.token` is scoped to the Inertia instance**

This might make it clearer:

**Visit 1 starts:**  `r1.token` and `inertia.token` are set to `1`
**Visit 2 starts:** `r2.token` and `inertia.token` are set to `2`
**Visit 1 ends:** `r1.token` is `1`, but `inertia.token` is `2` so we bail out of the update
**Visit 2 ends:** `r2.token` and `inertia.token` are both `2` so we update

reinink:
> Sorry where’s that? I don’t see multiple this.setState calls:

Yeah, I was looking at this way too late last night. 😂 

Let me review this again and get back to you.

src/inertia.js
```diff
@@ -42,6 +44,12 @@ export default {
     clearTimeout(this.progressBar)
   },
 
+  createToken() {
```
reinink: I'm confused...I don't see a token actually created here. Where does the token actually come from?

bradlc: Line 48. Like I said, `token` isn’t a great name.

reinink: Hah, sorry, not trying to be difficult, but on 48 you set the token to an empty object. At what point does the token actually have a value applied to it? I don't see that anywhere. It *seems* to me that token will only ever be an empty object. Help me understand. 😂

bradlc: Sorry @reinink! 😅 So the "token" _is_ the object. Objects are compared by **reference**, so `{} !== {}`. The key is that they are unique for each "visit."

Maybe "ID" is a better word: When we start a visit we assign a unique ID to that visit, and globally we keep track of the most recent visit ID. When a visit finally resolves (for example `inertia-vue` is about to render a new component) we check that the visit ID matches the most recent one. If it doesn’t it’s because a new visit was initiated since starting this one, so we want to skip over it.

reinink: Ahh yes, that makes way more sense now. 😂 I'm obviously not a great JavaScript developer. Why am I making this framework again? 🤔

src/inertia.js
```diff
@@ -52,6 +60,8 @@ export default {
 
     this.cancelToken = axios.CancelToken.source()
 
+    let token = this.createToken()
```
reinink: Could this get inlined on line 98? Like this:

```js
return this.setPage(page, this.createToken().then(didUpdate => {
```

src/inertia.js
```diff
@@ -10,7 +10,9 @@ export default {
 
   init(page, setPage) {
     this.version = page.version
-    this.setPage = setPage
+    this.setPage = (page, token = this.createToken()) => {
+      return setPage(page, () => token === this.token)
+    }
```
reinink: I wonder, does it really make sense to even call `setPage()` if we know we don't want to update? I wonder if this would be better:

```js
this.setPage = (page, token = this.createToken()) => {
  if (token === this.token) {
    return setPage(page)
  }
}
```

This is totally untested, but it would remove the need to have this handled in the specific adapters.

src/inertia.js
reinink: Hmm, I guess not, since the token has to be create BEFORE the request.

PR #26, opened 2019-05-09
Replace instead of push initial state

bradlc:
This avoids creating an additional history entry on intial load.

For example, currently if you open an Inertia app in a new window/tab you need to hit back _twice_ to leave.

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -16,7 +16,7 @@ export default {
       this.setPage(window.history.state)
     } else {
       this.setPage(page)
-      this.setState(page)
+      this.setState(page, true)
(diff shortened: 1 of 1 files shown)


reinink:
So this might fix the issue for your situation, and it might even be a good change, but technically if the page URL isn't changing, the `setState` method should automatically do a `replace` instead of a `push` state. As noted in #27 I'd love to verify what values you're getting when the URLs are being compared.

bradlc:
Hey @reinink. You’re right, looks like I was using an old version 🤦‍♂️ sorry about that!

PR #27, opened 2019-05-10
Replace instead of push remembered state

bradlc:
Currently when you `remember` something a new history entry is created. For example, when filling out a form on pingcrm, you end up with a load of entries for the same URL 🙈

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -149,7 +149,7 @@ export default {
     this.setState({
       ...window.history.state,
       cache: { ...window.history.state.cache, [key]: data }
-    })
+    }, true)
   },
 
   restore(key = 'default') {


reinink:
I am not able to reproduce this. What browser are you using?

The reason this shouldn't happen is on line 114/115:

https://github.com/inertiajs/inertia/blob/46fe3a886e417fdac6897eeb987f344aa5f6d5c3/src/inertia.js#L114-L115

Here we check to see if the URL is the same, and if it is we always replace instead of pushing on to state.

I recently made some changes to this code to account for the fact that browsers always return a slash at the end of the domain.

Also, server-side, are you returning a full URL?

To verify, I'd suggest console logging both `page.url` and `window.location.href` in the `setState` method to see what values you're getting. If they are different, when they should be the same (ie the same page), then that's the problem.

bradlc:
Hey @reinink. You’re right, looks like I was using an old version 🤦‍♂️ sorry about that!

PR #28, opened 2019-05-13
Allow passing query parameters for GET requests

mpociot:
Right now it's not possible to pass query parameters to `Inertia.visit`, as the `data` variable gets passed as `data` to axios - which is only for actual POST/PUT/etc. data.
This PR allows making use of the data object, in case of a GET request.

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -55,7 +55,8 @@ export default {
     return axios({
       method: method,
       url: url,
-      data: data,
+      data: method.toLowerCase() === 'get' ? {} : data,
+      params: method.toLowerCase() === 'get' ? data : {},
       cancelToken: this.cancelToken.token,
       headers: {
         'Accept': 'text/html, application/xhtml+xml',


thoresuenert:
@mpociot it would be helpful to have an option for the `paramsSerializer` function to pass to axios as well.

Problem:
When we use a json:api conform  get `filter` string (for example: `filter[name]=mpociot`), we cannot use the data object as implemented.

When we allow to make use of:
```js
//axios config (pass with each request)
paramsSerializer: function (params) {
    return Qs.stringify(params, {arrayFormat: 'brackets'})
},
```

we can use for my example: 
```js
paramsSerializer: function (params) {
    return Qs.Stringify(this.data(), {
            encode: false,
            addQueryPrefix: true,
            filter: (prefix, value) => (value === "" ? undefined : value)
          });
},
```

to make json:api conform request.

reinink:
Dig it, thanks Marcel!

PR #29, opened 2019-05-13
Replace axios with window.fetch

bradlc:
As discussed in #19, this PR replaces `axios` with browser-native `fetch` and `AbortController`, significantly reducing the size of Inertia for those that don’t use `axios` elsewhere in their app.

Out-of-the-box browser support is reduced to [those that support `fetch` and `AbortController`](https://caniuse.com/#feat=abortcontroller), but users can polyfill these to increase support. If that feels like it’s burdening the user too much, you could provide a build with polyfills included, or something like that.

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -75,7 +75,9 @@ export default {
         ...this.version ? { 'X-Inertia-Version': this.version } : {},
       },
     }).then(response => {
-      if (this.isInertiaResponse(response)) {
+      if (response.status === 409 && response.headers.has('x-inertia-location')) {
+        return this.hardVisit(true, response.headers.get('x-inertia-location'))
+      } else if (this.isInertiaResponse(response)) {
         return response.json()
       } else {
         response.text().then(data => this.showModal(data))
@@ -83,12 +85,6 @@ export default {
     }).catch(error => {
       if (error.name === 'AbortError') {
         return
-      } else if (error.response.status === 409 && error.response.headers.has('x-inertia-location')) {
-        return this.hardVisit(true, error.response.headers.get('x-inertia-location'))
-      } else if (this.isInertiaResponse(error.response)) {
-        return error.response.json()
-      } else if (error.response) {
-        error.response.text().then(data => this.showModal(data))
       } else {
         return Promise.reject(error)
       }
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -1,4 +1,3 @@
-import fetch, { AbortController } from 'unfetch-abortable'
 import nprogress from 'nprogress'
 
 export default {
@@ -61,7 +60,7 @@ export default {
 
     this.abortController = new AbortController()
 
-    return fetch(url, {
+    return window.fetch(url, {
       method: method,
       ...this.hasBody(method) ? { body: JSON.stringify(data) } : {},
       signal: this.abortController.signal,
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -1,10 +1,9 @@
-import axios from 'axios'
 import nprogress from 'nprogress'
 
 export default {
   setPage: null,
   version: null,
-  cancelToken: null,
+  abortController: null,
   progressBar: null,
   modal: null,
 
@@ -30,7 +29,16 @@ export default {
   },
 
   isInertiaResponse(response) {
-    return response && response.headers['x-inertia']
+    return response && response.headers.has('x-inertia')
+  },
+
+  hasBody(method) {
+    return ['GET', 'HEAD'].indexOf(method.toUpperCase()) === -1
+  },
+
+  getCookieValue(name) {
+    let match = document.cookie.match(new RegExp('(^|;\\s*)(' + name + ')=([^;]*)'))
+    return match ? decodeURIComponent(match[3]) : null
   },
 
   showProgressBar() {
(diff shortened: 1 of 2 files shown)


src/inertia.js
```diff
@@ -46,38 +54,36 @@ export default {
...
-      cancelToken: this.cancelToken.token,
+      ...this.hasBody(method) ? { body: JSON.stringify(data) } : {},
+      signal: this.abortController.signal,
+      credentials: 'include',
       headers: {
+        'Content-Type': 'application/json;charset=UTF-8',
         'Accept': 'text/html, application/xhtml+xml',
         'X-Requested-With': 'XMLHttpRequest',
         'X-Inertia': true,
+        'X-Xsrf-Token': this.getCookieValue('XSRF-TOKEN'),
```
sebastiandedeyne: Isn't this a bit too Laravel specific?

bradlc: It’s exactly what `axios` was doing before, as far as I can tell. `axios` has options for the cookie and header name to use for XSRF tokens, but those weren’t being explicitly set; `x-xsrf-token` (header) and `XSRF-TOKEN` (cookie) are the defaults. I guess we could expose it as an option?

sebastiandedeyne: Oh, I didn't know this was a generic thing across frameworks. Suppose it doesn't hurt to have then!

sjelfull: Would be nice if you could expose that as an option, yes. Some frameworks call it XSRF, some call it CSRF, so this isn't super generic.

pedroborges:
I was trying this myself today. I only had to update your solution a bit to fix some conflicts with the latest changes. Take a look: https://github.com/pedroborges/inertia/tree/fetch

When also replacing nprogress by nanobar (#47), the [Svelte](https://github.com/pedroborges/inertia-svelte) bundle almost halved in size.

|    | Axios & nprogress   | Fetch & nanobar   |
| ------- | ------- | ------- |
| Regular | 31.9kb  | 18.6kb  |
| gzipped | 11.4kb  | 6.9kb   |

I hope to see this merged. Fetch support is great for my audience. I'm happy to provide a polyfill to unsupported browsers.

reinink:
Hmm, it's worth noting, as of three days ago `AbortController` support was added to the [fetch polyfill](https://github.com/Financial-Times/polyfill-library/pull/59). That means it's also now available on polyfill.io.

With this in place, I am now much more interested in removing the axios dependency. @bradlc, would you be willing to update this PR to remove conflicts?

reinink:
So I've updated this PR myself (#57) thanks to some code from @pedroborges. However, I've been doing some digging on the differences between fetch and xhr, and fetch has one serious limitation still: you cannot track upload progress with it.

And while we're not currently using file upload tracking in Inertia, this is something I plan to add. For example, something like this:

```js
var data = new FormData();
data.append('id', this.user.id);
data.append('file', file);

$inertia.put('/users', data, {
  onUploadProgress(event) {
    // do something
  }
})
```

Unless you guys have a good workaround to this?

bradlc:
@reinink Hey. I resolved the conflicts but didn't have a chance to push before you updated your PR. I think in your version you’re missing the URL parameters for when the `method` is set to `GET`

reinink:
Yeah, honestly, my version might still be broken. 😂

Thoughts on the file upload progress stuff?

bradlc:
It's a tricky one. I think the _main_ reason to drop Axios is to save on bundle size, not to necessarily use `fetch`. So we could perhaps use `XHR` instead, if the upload progress thing is a deal breaker?

If you remember, my [original PR](https://github.com/inertiajs/inertia/pull/19) used [`unfetch`](https://github.com/developit/unfetch) which is a minimal `fetch` polyfill that uses `XHR`.

reinink:
Yeah, I had same thought about just dropping Axios and using using `xhr` directly. I have a few concerns:

1. I don't know what sort of `xhr` gotchas Axios handles for you.
2. It's actually handy to still have Axios around, for those one-off situations where you actually need to make a non-Inertia request.
3. Axios works in both the browser, and in Node. This could be helpful in the future if we want to do server-side rendering.

Obviously all three of those concerns can probably be handled without Axios. I just wonder with how much effort.

Also, FWIW, Vue Router [ships with axios](https://github.com/vuejs/vue-router/blob/dev/package.json#L56).

🤷‍♂

reinink:
Folks, I think, for now at least, Inertia will continue using Axios.

I love the idea of just using fetch, but given the file upload progress issue, we have to use xhr, and Axios is the easiest way to do that for v0.1. We have to pick our battles at this point. Maybe we aim for a custom xhr implementation for v1.0.

I'll definitely revisit this again in the future. Thanks @bradlc and @pedroborges for your efforts here.

Yiddishe-Kop:
> So I've updated this PR myself (#57) thanks to some code from @pedroborges. However, I've been doing some digging on the differences between fetch and xhr, and fetch has one serious limitation still: you cannot track upload progress with it.
> 
> And while we're not currently using file upload tracking in Inertia, this is something I plan to add. For example, something like this:
> 
> ```js
> var data = new FormData();
> data.append('id', this.user.id);
> data.append('file', file);
> 
> $inertia.put('/users', data, {
>   onUploadProgress(event) {
>     // do something
>   }
> })
> ```
> 
> Unless you guys have a good workaround to this?

Any idea when this will this be added to Inertia?

Yiddishe-Kop:
I just created #141 to add this feature. 🎉

SergkeiM:
> Folks, I think, for now at least, Inertia will continue using Axios.
> 
> I love the idea of just using fetch, but given the file upload progress issue, we have to use xhr, and Axios is the easiest way to do that for v0.1. We have to pick our battles at this point. Maybe we aim for a custom xhr implementation for v1.0.
> 
> I'll definitely revisit this again in the future. Thanks @bradlc and @pedroborges for your efforts here.

Hi @reinink 

Fetch API  supports: `ReadableStream` (https://developer.chrome.com/articles/fetch-streaming-requests/)

Browser Support: https://caniuse.com/?search=ReadableStream


Out of Topic, but if we are talking about Inertia Size:

1. I would replace `lodash.get` with `just-safe-get`
2. `qs` don't now alternative, but for sure is and overkill.

PR #32, opened 2019-05-16
Improve visit cancellation

reinink:
This is another take on @bradlc's work in #25.

This approach uses two callbacks (`resolveComponent` and `updatePage`) from the adapters, instead of just one (`setPage`).

This PR also makes some improvements to the loading indicators.

Closes #25

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -2,25 +2,28 @@ import axios from 'axios'
 import nprogress from 'nprogress'
 
 export default {
-  setPage: null,
+  resolveComponent: null,
+  updatePage: null,
   version: null,
+  visitId: null,
   cancelToken: null,
   progressBar: null,
   modal: null,
 
-  init(page, setPage) {
-    this.version = page.version
-    this.setPage = setPage
+  init({ initialPage, resolveComponent, updatePage }) {
+    this.resolveComponent = resolveComponent
+    this.updatePage = updatePage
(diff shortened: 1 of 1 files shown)


mirago:
Hey, is there a reason why the progress indicator does not have a timeout anymore?

reinink:
@mirago Simply because I couldn't get it working properly with the page cancellation. When a page is loading, and you click to visit another page, I want the progress bar to start over. This was basically impossible to get working with Nprogress. Honestly, Nprogress is kind of a pain, and it has been updated in over 4 years.

Long term I'll likely expose a way for folks to create their own loading indicators.

mirago:
I see, thanks for taking your time to explain. I just upgraded my project to this latest version and was quite surprised it displays nprogress all the time now :) No worries though, and I agree it would be nice to add some hooks.

reinink:
For what it's worth, I found that with the 100ms delay, it basically always shows in production anyway. This isn't great, since the local experience is different than production experience.

reinink:
@mirago The progress bar delay has been brought back: https://github.com/inertiajs/inertia/commit/fc4775ce33f6e8875b76e11973aedfba209ff476

MichaelDeBoey:
One step closer to ditching the dependency 😉

reinink:
Yup.

PR #38, opened 2019-05-18
Fix resolveComponent types

MichaelDeBoey:


reinink:
I'm just going to go ahead and assume that was correct. 😂

MichaelDeBoey:
You either return a `Component` or a Promise that returns a `Component`, no? 🤔

reinink:
Yep, exactly. Previously you _had_ to return a promise from the callback. Now you can just return the component straight up.

MichaelDeBoey:
So that seems correct then the  return type 😉

PR #44, opened 2019-05-21
Return promise result

druc:
Currently whenever a request is made the `then()` method returns undefined.

Example: 
```
this.$inertia.post(this.route('projects.members.store', this.project.id), this.form)
  .then((res) => {
    console.log(res); // undefined
  });
```

When sending the request from inside a modal component, for example, I'd like to know if the request returned any errors so I can keep the modal open.

I have a button/modal component like this:
```
// inside Pages/Project/Members.vue
<add-project-member-button :project="project" :errors="errors">Add member</add-project-member-button>
```

Adding a watcher won't work because the modal will flicker.

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -86,6 +86,7 @@ export default {
       if (page) {
         this.setPage(page, visitId, replace, preserveScroll, preserveState)
       }
+      return page;
     })
   },
 


reinink:
Does it make sense to return the page here? Or would it be better to simply return true/false?

druc:
Hmm.. 
Currently I have:
```
submit() {
    this.sending = true;
    this.$inertia.post(this.route('projects.members.store', this.project.id), this.form)
        .then((res) => {
            this.sending = false;
            if (!Object.keys(res.props.errors).length) {
                this.open = false;
            }
        });
}
```

Returning a boolean would turn this into:

```
submit() {
    this.sending = true;
    this.$inertia.post(this.route('projects.members.store', this.project.id), this.form)
        .then((success) => {
            this.sending = false;
            this.open = !success;
        });
}
```

But other people might want to do stuff with the result (although probably that will end up in a mess).

Personally, I like the true/false alternative. Otherwise I'll have to `Object.keys(res.props.errors)` all over the place...

reinink:
Hmm, not sure that would work, since any valid Inertia response will return `true`, even if it's an error response.

A better approach here would be to make the modal a query param, that's a prop, controlled by the server.

@sebastiandedeyne recently sent me this:

> I have this page that opens a modal. Turns out that modal also needed its own URL so you can get redirected to it.
> 
> I was able to lift the modal open/closed state to the _server_, so you can visit the page with the modal already open. When clicking "show" or "hide" modal, I used the preserveScroll option, so the content behind the modal stays in place and you don't really realise you're navigating a new page.

sebastiandedeyne:
Still happy with that solution!

Every time I'm about to use axios, I ask myself the question if I can't just solve the problem with an Inertia request.

druc:
Maaan... 

I knew it from the start that "it doesn't _feel_ right" - that I'm just going back to how I used to do things, controlling the state on the client.

@sebastiandedeyne's solution worked like a charm. Not only my modal can stay open, now the people using the app can also share links to that specific modal - which is great!

Moving back to returning _something_, now I think it shouldn't return _anything_ - this will prevent people from falling in the same trap I did and "guide" them to do things the "inertia" way.

Thanks a lot for the help and for making inertiaJS - really love it so far!

PS: is there any forum/slack/discourse thing where people can bounce ideas around (such as above)?
Sorry for not reading "Please do not submit pull requests without first consulting me on Twitter" ..

reinink:
Awesome @druc! 👏

And I think I agree. Let's leave it as it is, without a response, as to not encourage doing things the wrong way.

reinink:
@druc And nope, no chat channel yet. Maybe I should set that up. 🤔

iainco:
> Still happy with that solution!
> 
> Every time I'm about to use axios, I ask myself the question if I can't just solve the problem with an Inertia request.

Would it be possible to share some code snippets to show how you're achieving this? Would be much appreciated.

druc:
In my case I have something like this:
- `projects/{id}/phases` - page with phases containing an "Add phase" button which will open a modal with a form to create a new phase.
- clicking that button makes a request to `projects/{id}/phases/create` which returns 
```
return Inertia::render('Projects/Phases', [
    'project' => ProjectResource::make($project),
    'showPhaseModal' => true
]);
```
- in vue (`Projects/Phases`) I show that modal when "showPhaseModal" is true
- when I click the "Save" button in the modal's form, that submits a post request to `projects/{id}/phases` which redirects to the initial `/projects/{id}/phases` route, closing the modal (since the index does not return "showPhasesModal" as being true)

The easiest way to get into this mindset is to think more in terms of views (`Projects/Phases.vue` and `Projects/Phases.vue + an opened PhaseModal` are different "views") and make more use of what the server "gives" you and less on what you can change with javascript.

iainco:
Thank you for your comment, that makes sense.

How does this work in terms of your component structure - is <modal/> inside the main Phases template? 'showPhaseModal' is a prop then? I'm just not clear on how I can pass the boolean through to the child modal and have it behave accordingly. I'm probably not making much sense but I will give it a go - currently my controller is calling Redirect::route(...) rather than Inertia::render(...) so that's the obvious first thing to change.

Thanks again.

iainco:
PS. How are you opening the modal, presumably not with an Inertia request to the server first?

tinyfly:
> The easiest way to get into this mindset is to think more in terms of views (`Projects/Phases.vue` and `Projects/Phases.vue + an opened PhaseModal` are different "views") and make more use of what the server "gives" you and less on what you can change with javascript.

@druc This is fantastic, thanks for sharing. I use modals with their own url quite a lot and not knowing how to do this has kept me from trying out Inertia. Now that you've given an example I'm off to the races.

nokios:
I think it would be helpful to have a slightly more thorough walk through of how this would be done. I'm still having difficulty figuring out how the full thing looks.

Do both endpoints return the same View component, and do you need to preserve state between calls?

In my example, I have a table with an edit button for each row. and the design calls for the ability to edit the details of that row in a modal and save the changes there.

druc:
Hey @nokios

There are three ways you can do this:
- you could pass an optional parameter to your index route and use it to find and return your record - and then use its presence to show/hide the modal

- you could call the edit route along with the `only: []` option. This will keep your current state while adding whatever you return on the edit endpoint (should return the same inertia view though). The main problem with this is that you won't be able to visit the page directly - because of missing the data needed for the index to work.

- depending on your specific case, it might make sense to go with the traditional way where you open the modal, make an ajax request to load the record, and then another one to update it.


I wanted to record a short video explaining, but I suck at talking 🤷 so I'll paste a bunch of code hoping it's good enough to get you on the right track. Feel free to ask questions if you get stuck / you come up with an even better alternative

```
// web.php
Route::get('contacts/{contact?}')->name('contacts')->uses('ContactsController@index')->middleware('remember', 'auth');

// ContactsController
public function index($contact = null)
{
    return Inertia::render('Contacts/Index', [
        'filters' => Request::all('search', 'trashed'),
        'contact' => $contact ? Contact::findOrFail($contact) : null,
        'contacts' => function () {
            // lazy evaluation so you don't query this again when you only really want the $contact
        },
    ]);
}

// Contacts/Index.vue
<template>
    <div>
        <edit-contact-modal :contact="contact"></edit-contact-modal>
        <table>
            ...
            <td>
                <a class="px-6 py-4 flex items-center focus:text-indigo" @click.prevent="edit(contact)" only="contact">
                  {{ contact.name }}
                  <icon v-if="contact.deleted_at" name="trash" class="flex-no-shrink w-3 h-3 fill-grey ml-2"/>
                </a>
            </td
            ...
        </table>
    </div>
</template>

<script>
 methods: {
  edit(contact) {
    // optional parameter for the index method
    this.$inertia.visit(this.route('contacts', {contact: contact.id}), {only: ['contact']})

    // load contact & organizations data from edit() while preserving the current records
    // however, navigating directly to /contacts/12/edit won't work because it wont have the "contacts" required by the index view
    this.$inertia.visit(this.route('contacts.edit', {contact: contact.id}), {only: ['contact', 'organizations']})
  },
</script>

// Modal looks something like
<template>
  <portal to="dropdown">
    <transition name="fade-shrink-in" appear>
      <div v-show="contact" @click="closeModal()" class="absolute top-0 left-0 h-full w-full" style="background-color: hsla(0, 0%, 0%, .5); z-index: 99998">
        <form @submit.prevent="confirm()" class="max-w-sm w-full mx-auto bg-white rounded p-4 mt-8 z-10 shadow-lg" @click.stop>
            content
        </form>
      </div>
    </transition>
  </portal>
</template>

<script>
  export default {
    props: {
      contact: Object,
    },
    data() {
      return {
        form: {
          first_name: null,
        },
      }
    },
    mounted() {
      if (this.contact) {
        this.form = this.contact
      }
    },
    methods: {
      closeModal() {
        this.$inertia.visit(this.route('contacts'))
      },
    },
  }
</script>
```

nokios:
@druc This has helped me immensely, thank you.  Following your example I was able to get the modal working. I'm not exactly a fan of having one controller do both the list and the edit modal but it seems to be the cleanest way to do it right now.  I am testing using inertia and vue to replace graphql in our stack and so far I like it far better than GraphQL (we're not really using it right)

robsontenorio:
I would like to share the "optical illusion" for future reference.

- For that form, just keep laravel default resource controller routes: index, create, edit, store, update, destroy.
- Just create a separated inertia/vue page/.
- Handle this form in traditional way, by call respective routes/actions.
- **But**, instead to show this page in traditional way, just create it like a modal that always show by default with a **fake background**.

Screen record from 2020-06-18 12 55 21

reinink:
I've created a new issue dedicated to modals in Inertia, see #249.

PR #46, opened 2019-05-31
Allow customizing request headers

nikazooz:
This will allow users to customize request headers when using Inertia to send requests.

One of uses for this is to set "Content-Type" header to "multipart/form-data" when uploading files using `FormData` object. This particular case has been mention here: https://github.com/inertiajs/inertia-vue/issues/48

Cheers 🙂

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -45,7 +45,7 @@ export default {
     return this.visitId
   },
 
-  visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false } = {}) {
+  visit(url, { method = 'get', data = {}, headers = {}, replace = false, preserveScroll = false, preserveState = false } = {}) {
     Progress.start()
     this.cancelActiveVisits()
     let visitId = this.createVisitId()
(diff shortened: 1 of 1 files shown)


reinink:
This actually isn't necessary, since Axios automatically handles setting the `Content-Type` header for you when you use `FormData`.

While I'm not completely opposed to allowing folks to manually set headers on Inertia requests, I want to try and avoid this a long as possible. My hope is this will never be necessary.

nikazooz:
Understandable 🙂
I've noticed that this is not necessary, but forgot to close the PR myself 😅

enzonotario:
Hi folks! I have an use case for this: We pass an `Accept-Language` header with the selected language (the user can switch as he/she want)... But maybe there are also another ways to handle this... What do you think? If I can customize the headers, I would do what I want as always we did it :)

sebastiandedeyne:
Okay I think I got it.

https://github.com/enzonotario/pingcrm/blob/50796da1b2bc61339fd433a1ae243869bc3c6db6/package-lock.json#L836-L857

Inertia 0.1.0 is requiring 0.18 of Axios. Could you upgrade to the latest version (0.1.3)?

enzonotario:
yes!! that solved the problem! thanks so much!

Stian-Scholtz:
@reinink Could I ask why you are trying to avoid this?  It just causes headaches for the people that need to pass custom headers. I love Inertia, but this is just unnecessary from my view point. I am Broadcasting events in my app and need to block the broadcasting for the current user. To do that I need to pass a custom header.

claudiodekker:
Heads-up: Custom headers are supported as of #202.

Issue #10, opened 2019-04-16
Inline nprogress css


reinink:
While Inertia uses [nprogress](http://ricostacruz.com/nprogress/) by default for showing page load progress, it *doesn't* actually include any CSS for it. You are still required to include this on your own ([example](https://github.com/inertiajs/pingcrm/blob/master/resources/css/nprogress.css)).

It would be nice to simply inline this somehow. It would be wise to review how [Turbolinks does this](https://github.com/turbolinks/turbolinks#displaying-progress), since their CSS is mega simple.

reinink:
I'm thinking it makes sense to just create our own progress bar. We can go off Turbolinks [implementation](https://github.com/turbolinks/turbolinks/blob/866a29c861b0e0e0625a06189a49e29abd7a344a/src/progress_bar.ts) for this.

cmizzi:
I don't think Inertia should integrate a progress bar. A hook system, before and after the visit should be implemented in order to let the user deploy his own progress bar (with a sample example in the documentation). For example, in my project, I don't use the Inertia progress bar, but only indicator based on the promise returned by the `Inertia.visit()` method. It should be great if a hook can be applied globally and manage my own progress bar, without having a progress bar dependencies I don't use.

reinink:
While that approach would certainly be the easiest, I don't think it's the best from a DX perspective. I think we need to ship with something.

cmizzi:
With a full documentation about how to register `nprogress` with hooks, it could be easy to set up ?

reinink:
This has been done, but in the https://github.com/inertiajs/progress package. 👍

PR #47, opened 2019-06-04
Replace nprogress by nanobar

pedroborges:
Nanobar solves inertiajs/inertia#10. It injects the CSS needed to display the bar. The CSS classes can be easily customized. It is just ~720b after compression.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "typings": "index.d.ts",
   "dependencies": {
     "axios": "^0.18.0",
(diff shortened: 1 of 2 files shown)


reinink:
Hey thanks for this @pedroborges! But, as we've discussed, I think we're going to actually build in a progress bar directly into Inertia for 1.0. Hopefully you can help with that. 😉

pedroborges:
No problem. Let me know what you have in mind. I'd be glad to submit a PR.

PR #48, opened 2019-06-13
Pass initial window hash to initialPage

mpociot:
Since the `initialPage` gets generated on the server side, it will never contain the current window hash. This currently results in some unwanted behavior.

When you open: `http://localhost/view#something` you will end up at `http://localhost/view` without the current hash.

This PR adds it to the initial page so that you can retain the hash.


reinink:
Nice chat @mpociot! Thanks for this. 💪

PR #59, opened 2019-08-08
Updated "visit" method to pass response further.

sandulat:
Now we can access the response directly from Inertia request promises like this:

```
this.$inertia.post('/', form).then(response => {});
```

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -87,6 +87,8 @@ export default {
       if (page) {
         this.setPage(page, visitId, replace, preserveScroll, preserveState)
       }
+      
+      return page;
     })
   },
 


reinink:
I am curious, why do you want do access the response in this way?

sandulat:
@reinink I have a form component that accepts a submit callback. If there are any errors from the backend I set errors back to the form. Example:

```vue
<template>
  <form-login :on-submit="login" />
</template>

<script>
import FormLogin from '@/Shared/Forms/FormLogin.vue';

export default {
  components: {
    FormLogin,
  },
  methods: {
    login(form, { resetField, setFieldError }) {
      return this.$inertia.post(this.$route('login.attempt'), form).then(() => {
        if (this.$page.errors.email) {
          setFieldError('email', this.$t('validations.invalidCredentials'));

          resetField('password');
        }
      });
    },
  },
};
</script>
```
The example above won't work because `this.$page` didn't update yet, so I have to workaround it like this:
```js
setTimeout(() => {
  if (this.$page.errors.email) {
    setFieldError('email', this.$t('validations.invalidCredentials'));

    resetField('password');
  }
});
```
With the tiny change from this pull request I could do:
```js
return this.$inertia
  .post(this.$route('login.attempt'), form)
  .then(response => {
    if (response.props.errors.email) {
      setFieldError('email', this.$t('validations.invalidCredentials'));

      resetField('password');
    }
  });
```

This is a very specific scenario but I believe there could be more use cases. Anyway I guess the availability of response won't be redundant.

reinink:
I thought this might be the case. So, with Inertia, this isn't really the recommended way of handling error validation. Let me copy/paste [this response](https://twitter.com/jamesburrow23/status/1158727081087447051) from Twitter.

**Question:**

> So how do you handle a 422 response in Inertia when you're posting data?  Do I need to add a custom header so inertia knows how to catch failed validation?  Thanks for your feedback

**Answer:**

> Great question! The trick is to not think of Inertia like a client-side framework, but rather a server-side framework.
> 
> So, with a client-side framework you'd catch the errors client-side, which would include some JSON error messages, and you'd update your view to show them.
> 
> With Inertia, you handle it like a server-side form submission.
> 
> You submit your form (POST). If there are errors, your server-side framework redirects back (GET) to your form page, and includes the errors in the session.
> 
> Then, you need to send those error messages to the view. Laravel automatically makes them available within the Blade views as $errors. However, we're not using Blade. So, we need to include the errors with our Inertia response, so they can be passed back to our page as a prop.
> 
> In my apps, I do this automatically on every request. If there are errors in the session, I include them as an `errors` prop.
> 
> You can see this here: https://github.com/inertiajs/pingcrm/blob/master/app/Providers/AppServiceProvider.php#L48-L50
> 
> This makes it work a lot like Blade, which always has access to the $errors object.
> 
> Your form page component has now received updated props, which include the `errors` prop, also accessible as $page.errors in Vue.js. Since Vue.js/React are reactive, the page will automatically update to show the form errors.
> 
> See here: https://github.com/inertiajs/pingcrm/blob/master/resources/js/Pages/Users/Create.vue#L10-L14
> 
> So, in summary:
> 
> 1. Submit form via $inertia.post
> 2. If errors, server-side redirect back (GET) to form page
> 3. The errors are saved in the session
> 4. Errors are passed as shared data to Vue.js
> 5. Your page displays those errors, which happens reactively
> 
> This certainly requires a shift in thinking about things, but this whole approach works REALLY well.
> 
> It's just like how we used to do classic server-side form submissions. Except it's way nicer than that, since you're not reloading the whole page, or rehydrating form input.

reinink:
Here is a similar PR: #44

I think, for now, we're going to try and avoid having to do this. I think the approach described above is better...more inline with the Inertia philosophy.

sandulat:
@reinink Yeah I do agree that sometimes it's better to "push" the user towards a certain philosophy. Sorry, should've check the closed PRs first.

reinink:
@sandulat All good! You are not the first person to wonder about how to handle errors. Honestly, check out the approach I'm suggesting. It works **awesome**.

PR #61, opened 2019-08-10
Add support for partial reloads

### LINKED ISSUE #60: Partial page reloads
Different sections of a page sometimes have different data requirements. For example a datatable with a set of dropdown filters: when you switch pages in the datatables, the dropdown data will also be refetched on the server. This isn't necessarily bad, but this can slow down your app with large datasets. 

Ideally, there would be a way to tell Inertia which data you _want_ changed, so it would only fetch and update subset of the page's data.

On the server, data would need to optionally be marked as potentially "lazy". In PHP/Laravel, the data could be wrapped in a callback:

```php
class UsersController extends Controller
{
    public function index()
    {
        return Inertia::render('Users/Index', [
            'users' => Users::paginate(),
            'countries' => function () {
                return Country::all();
            }
        ]);
    }
}
```

*(Sidenote: this only affects subsequent loads on the same page, the first visit will always fetch all of the page's necessary data)*

On the client, you'd need to be able to specify which data you want to reload. In Vue:

```js
this.$inertia.reload({ only: ['users'] });
```

This would send a request with an `X-Inertia-Only` header containing the `users` key, so Inertia can strip out the rest of the data on the server. The `Country::all()` database query would be completely skipped because it never gets evaluated.

When the request comes back, the new props will be merged with the old ones.

---

Kind of related to inertiajs/inertia-laravel#37

Thanks @reinink and @adamwathan for the feedback on this earlier 💪

reinink:
This PR adds support for partial reloads (#60). **This is still a work-in-progress.**

Partial reloads allow you to make subsequent visits to the _same page_ and only request a subset of the props (data) from the server. You do this with a new `only` option on visits.

```js
Inertia.reload({ only: ['users'] });
```

This requires coordination server-side to return only the subset of data. To do this Inertia passes along a new `X-Inertia-Only` header which includes a comma separated list of data keys that will returned. For example:

```php
X-Inertia-Only: users
```

When the server returns a subset of the data from the server, Inertia automatically merges it with the full page object from the previous request that's already in memory.

You can see the Laravel implementation of this here: https://github.com/inertiajs/inertia-laravel/pull/50

## Limiting partial page loads

Currently the `only` option is available for all visit types (not only `reload`) because it was easiest to implement it that way. I don't really see a way to make partial loading not working when going to other pages, so it probably makes sense to limit this. Or, maybe we just throw an exception if you try to make a partial request to a different URL. This still needs to be figured out.

At first I thought we'd just use `Inertia.reload()`. However, that only does `GET` requests. I could see you potentially doing partial reloads on `POST`, `PUT`, `PATCH` and `DELETE`.

Then again, doing this with non-`GET` requests is actually pretty dark territory. It means that the front-end has to know which endpoint the server will redirect to, and that actually could be different depending on the result. For example, on a form validation error it would return to the same page, but maybe on success it redirects to the index. That would mean it's now has the `X-Inertia-Only` header when navigating to a different page. Not good.

I think it probably makes the most sense to simply include the current page component name in the header as well. For example:

```php
X-Inertia-Only: [Users/Index,users]
```

My fear is that even a GET request could end up redirecting elsewhere...for whatever reason. For example, the user gets logged out. Or their permission change. Or SOMETHING. Now suddenly they are doing a partial reload on a different page.

If we include the page component name in the header, we can make sure we only implement it if the final page component is the same. That also means we don't have to limit it to `Inertia.reload()`, since it will literally only work if the final endpoint uses the same page component.

Closes #60.

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -62,7 +62,10 @@ export default {
         Accept: 'text/html, application/xhtml+xml',
         'X-Requested-With': 'XMLHttpRequest',
         'X-Inertia': true,
-        ...(only ? { 'X-Inertia-Only': only.join(',') } : {}),
+        ...(only ? {
+          'X-Inertia-Partial-Component': this.page.component,
+          'X-Inertia-Partial-Data': only.join(','),
+        } : {}),
         ...(this.version ? { 'X-Inertia-Version': this.version } : {}),
       },
     }).then(response => {
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -8,6 +8,7 @@ export default {
   version: null,
   visitId: null,
   cancelToken: null,
+  page: null,
 
   init({ initialPage, resolveComponent, updatePage }) {
     this.resolveComponent = resolveComponent
@@ -46,7 +47,7 @@ export default {
     return this.visitId
   },
 
-  visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false } = {}) {
+  visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false, only = [] } = {}) {
     Progress.start()
     this.cancelActiveVisits()
     let visitId = this.createVisitId()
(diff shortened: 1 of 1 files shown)


reinink:
Starting to wonder if this would be better done with two headers:

Option 1:

```
X-Inertia-Partial: Users/Index,users,other
```

Option 2:

```
X-Inertia-Partial-Component: Users/Index
X-Inertia-Partial-Data: users,other
```

reinink:
I decided on second header option and have implemented that.

PR #62, opened 2019-08-12
Preserve state on hard visit

reinink:
Right now when Inertia detects an asset version change, it does a hard visit (full page reload) to force update the assets. During this process it must maintain both server-side and client-side state.

Server-side "state" includes any flash messages or validation errors. The Inertia middleware solves this by [reflashing the values](https://github.com/inertiajs/inertia-laravel/blob/master/src/Middleware.php#L21) automatically.

Client-side state also needs to be preserved. This PR updates Inertia to do this. The challenge is, by default, on a hard page reload, Inertia will not restore state. This is to match default browser behaviour.

To get around this, Inertia creates a temporary `inertia.hardVisit` boolean in local storage. If that value exists on reload, it will restore remembered state just like a normal back/forward navigation.

Note in this video how Inertia is remembering 1) the flash message, 2) validation errors, 3) local form values across a **full page reload**. 🎊

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -16,6 +16,9 @@ export default {
 
     if (window.history.state && this.navigationType() === 'back_forward') {
       this.setPage(window.history.state)
+    } else if (window.localStorage.getItem('inertia.hardVisit')) {
+      window.localStorage.removeItem('inertia.hardVisit')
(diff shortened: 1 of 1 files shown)


reinink:
Good call @sebastiandedeyne. Updated.

PR #63, opened 2019-08-13
Update dependencies

MichaelDeBoey:


package.json
```diff
@@ -18,11 +18,11 @@
     "prepublishOnly": "npm run build"
   },
   "dependencies": {
-    "axios": "^0.18.0",
+    "axios": "^0.19.0",
```
sebastiandedeyne: Can this be changed to `^0.18|^0.19`?

Laravel 5.8 uses v0.18. If we suddenly start requiring v0.19 a lot of people would end up with duplicate dependencies without being aware.

MichaelDeBoey: `axios` has a security vulnerability up until `^0.18`, so `^0.19` should be the one we want to target.
https://snyk.io/blog/a-denial-of-service-vulnerability-discovered-in-the-axios-javascript-package-affecting-all-versions-of-the-popular-http-client/

MichaelDeBoey: Laravel 5.8 has been updated to `^0.19` too btw
https://github.com/laravel/laravel/pull/5038

sebastiandedeyne: 👍

PR #68, opened 2019-08-28
Modernize typings

MichaelDeBoey:
After coming across [`@types/express-serve-static-core`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-serve-static-core/index.d.ts), I finally found a way of not needing to declare a module (which is an ES5 way of doing things).

Right now we just export the same things like we do in [index.js](https://github.com/inertiajs/inertia/blob/master/src/index.js).
We also export `PagePropsBeforeTransform`, `PageProps` and `Page`, since they're used in the `inertia-react` adapter.
All other types aren't exported anymore and stay private.

`PagePropsBeforeTransform` and `PageProps` can still be overwritten globally via declaration merging/interface augmentation.

I think @sebastiandedeyne should check this out in his practice project to see if everything still works like expected (I think it does), since I don't have access to that project.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,74 +1,69 @@
-declare namespace Inertia {
-  interface PagePropsBeforeTransform {}
+declare global {
+  // These open interfaces may be extended in an application-specific manner via
+  // declaration merging / interface augmentation.
+  namespace Inertia {
+    interface PagePropsBeforeTransform {}
 
-  interface PageProps {}
-
-  interface Page<CustomPageProps extends PageProps = PageProps> {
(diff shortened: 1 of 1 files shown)


MichaelDeBoey:
Ping @sebastiandedeyne

sebastiandedeyne:
Looks a lot better, and works like a charm, thanks!

MichaelDeBoey:
With pleasure! 🙂

Will check if inertiajs/inertia-react#46 is good and make it a real PR then 🙂

Issue #65, opened 2019-08-23
Reloading with data duplicate query parameters


dsazup:
If I am on a `/users?page=1` page, and I click on a button that does 

```
Inertia.reload({
  data: {
    page: 1,
 }
```

I now see `/users?page=1&page=1`, and it keeps duplicating when I click again. Is this expected behaviour or is this a bug?

Juhlinus:
Hello @dsazup ! 

It's the expected behavior. The replace method just fetches the current url and adds your parameters to it, so if you just pass the same argument data over and over, then the same data will be added to the URL. 

This might be something that we'd want to check for upon calling the method as to not generate the result you're experiencing.

Maybe something along these lines:
```javascript
  replace(url, options = {}) {
    return this.visit(url.split('?')[0], { preserveState: true, ...options, replace: true })
  },
```

dsazup:
Ok, makes sense. It would be really nice if Inertia handled this, but I suppose this might be out of scope for this project. I'll leave it up to you guys to decide what you want to do with this issue 👍

RSpeekenbrink:
Maybe its an idea to replace 'window.location.href' in:

https://github.com/inertiajs/inertia/blob/472c1d3daa455b67df68aa71b6e2a656724d9652/src/inertia.js#L160-L162

With something like:

```
window.location.origin
```

and some kind of parser to parse the current url parameters to the data object in the options function parameter? And the parser will prefer option data over current url parameters so you can 'replace' them?

Just some brainstorming :) it would be nice indeed to fix this in Inertia.

anthony-lopez-dev:
@dsazup @Juhlinus @reinink 

It's just an idea, but don't you think it can be possible to put parameters as an object in shared data (`this.$page.parameters`) and use "[Vue Reactivity](https://vuejs.org/v2/guide/reactivity.html)" to watch when that object is modified. When the event is triggered, you rebuild url based on parameters object updated and use`Inertia.visit`.
May I have your opinion ? =)

I have a project where users can choice between showing 8 or 16 posts with a `<select>`. So I think it can be good to use that pattern and set params when my select change: `onChange : setParameters('per_page', $value)`

hotmeteor:
Any solutions on this one?

hotmeteor:
Alright, here's how I made it work (I'm using Ziggy as well):
```js
this.$inertia.replace(route(route().current(), route().params), {
    method: 'get',
    data: this.filters,
    replace: false,
    preserveState: true,
[...]

PR #69, opened 2019-08-30
Merge query parameters when calling reload

georgehanson:
This PR is a proposed way to fix the issue #65.

When a user calls the `reload` method, it calls the `mergeQueryParameters` method and sets the result to the `data` option.

The `mergeQueryParameters` simply creates an object based upon the current query parameters and merges them with the parameters provided by the user.

The `reload` method has then been changed so that rather than passing the full url including query parameters, it constructs the url itself using the origin and the pathname.

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -149,12 +149,25 @@ export default {
     }
   },
 
+  mergeQueryParameters(dataToMerge = {}) {
+    let urlSearchParams = new URLSearchParams(window.location.search)
+    let queryParameters = {}
+
+    for(let parameter of urlSearchParams.entries()) {
+      const [key, value] = parameter
+      merged[key] = value
+    }
+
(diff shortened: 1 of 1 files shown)


sebastiandedeyne:
I don't think this supports nested query parameters.

Parsing/stringifying query paramters is tough, because there's no spec of any sort to adhere too. I'm open for doing more for query parameters in Inertia, but it'd need to be something configurable.

I might be overthinking things :D Just dropping my 2 cents here

thoresuenert:
Possible break the usage of `axios paramSerializer`.
Not tested yet.

phas0r:
So what will be the way forward on this? I think this is important.

reinink:
Hey thanks for attempting this @georgehanson. I've actually fixed this in #264, which handles deep merging of query strings. 👍

PR #73, opened 2019-09-20
No inertia response handler

KABBOUCHI:
Example:

```js
new Vue({
    store,
    render: h =>
        h(InertiaApp, {
            props: {
                initialPage: initialPage,
                resolveComponent: n => import(`@/Pages/${n}`).then(m => m.default),
                noInertiaResponseHandler: ({ status, data, request }) => { 
                    if(status === 200) {
                        window.location.href = request.responseURL;
                    }else if(status === 403) {
                        Inertia.visit('/admin/403');
                    }
                }
            }
        })
}).$mount(app);
```

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -1,5 +1,5 @@
 import Axios from 'axios'
-import Modal from './modal'
+import defaultNoInertiaResponseHandler from './noInertiaResponseHandler'
 import Progress from './progress'
 
 export default {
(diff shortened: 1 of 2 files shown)


m1guelpf:
This would really help for managing redirects to non-Inertia views while still preserving the Modal functionality on exceptions

sebastiandedeyne:
Hmm, I'd rather solve this on the server with https://github.com/inertiajs/inertia-laravel/issues/57

reinink:
Yeah, I agree with @sebastiandedeyne, this is probably better handled server-side. Right now this would mean that the destination would be loaded twice, which isn't great.

A better solution is coming for this. Thanks either way for the contribution!

KABBOUCHI:
Why we can't solve it from both sides? it dosn't solve everything from laravel side. (ex: if website is down)

PR #87, opened 2019-10-14
setScroll add html to query selector

guidoerfen:
Some browsers apply the "overall" scroll to document.documentElement (the <html> element) and others to document.body (the <body> element). For compatibility with both, you have to apply the scrolling to both.


ruledio:
Can confirm this works perfectly, just came across this issue!

reinink:
Thanks @guidoerfen!

PR #91, opened 2019-10-23
Replace axio with fetch

pkboom:
This PR replaces axio with fetch. I guess fetch works great with Inertia except error handling, which is a little blurry. Any comment would be welcome. @reinink Inertia is great. :fire: Thanks.

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -1,4 +1,3 @@
-import Axios from 'axios'
 import Modal from './modal'
 import Progress from './progress'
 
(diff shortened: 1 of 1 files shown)


Juhlinus:
Hello @pkboom !

Thanks for your contribution!

There's already been a discussion regarding this on this PR: #29 and Jonathan came to the conclusion that we'll stick with axios for now.

reinink:
Thanks for the contribution @pkboom! However, as @Juhlinus noted, we did explore this once, but I decided against it due to lack of file upload progress support with Fetch.

We may revisit this in the future though! 💪

PR #96, opened 2019-11-06
NProgress show spinner

hivokas:
I need a spinner mode for my app but unfortunately, it is disabled.

### DIFF
--- a/src/progress.js
+++ b/src/progress.js
@@ -1,6 +1,6 @@
 import Nprogress from 'nprogress'
 
-Nprogress.configure({ showSpinner: false })
+Nprogress.configure()
 
 export default {
   delay: null,


ravenberg:
Hi @hivokas, thank you for contributing. 

I believe the spinner has been disabled by design. But taking your example to [the documentation](https://github.com/inertiajs/inertiajs.com) would be a great benefit.

Sinnbeck:
@hivokas This can easily be changed in your own app.js

Add this before booting inertia.
```
Nprogress.configure({showSpinner: true});

reinink:
Going to close this, since NProgress has been removed from the core, and we've added a new [progress package](https://github.com/inertiajs/progress) for this, which includes a `showSpinner` option. 👍

PR #104, opened 2019-12-12
add box-sizing: border-box to modal

timrspratt:
When using inertia with no styling framework, the width of the modal is bigger extends past the right hand side of the screen. 99% of modern web apply `box-sizing: border-box;` to all elements and so this is not noticeable, but when no CSS reset styles are used, this display issue is present.


timrspratt:
@reinink bump

reinink:
Thanks @timrspratt!

PR #105, opened 2019-12-20
Add class to modal

KABBOUCHI:
Add ability to customize the modal by targeting it using `inertia-unrecognised-response-modal` css class.

Sometime, it will triggered on production env, this way I can make it full screen instead of a modal.


bcalik:
@KABBOUCHI @reinink why was this PR closed? I was about to open the same PR, we better add a unique selector to have control over it.

Issue #109, opened 2020-01-29
Capture exception that pushState or replaceState throws


SabatinoMasala:
Firefox has a rather cryptic error message when we add a large state to the browser history. The Mozilla docs mention the [hard limit of 640k characters](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState#Parameters). When you encounter this issue you get an `NS_ERROR_ILLEGAL_VALUE` error in your console, nothing more.

Intertia could handle this exception in the [setState](https://github.com/inertiajs/inertia/blob/master/src/inertia.js#L138) method, and provide the developer with some insights on what could have been the issue.

Related issue: https://github.com/inertiajs/inertia-vue/issues/106

Would like some extra input before creating a PR.

reinink:
Thanks so much for looking into this @SabatinoMasala and sharing your findings. I'd certainly be open to a PR that adds an exception handler for this.

affektde:
That's a stupid question:

If you do transfer large data, couldn't you use an option in an `Inertia request` as a new feature that excludes certain data for `pushState`? Large amounts of data for data tables would not necessarily have to be included in the history. Or do I see it wrong?

Something like:

```javascript
Inertia.visit(url, {
  method: 'get',
  data: {},
  replace: false,
  preserveState: false,
  preserveScroll: false,
  only: [],
  excludeFromPushState: ['myLargeDataKey'],
})
```

```javascript
window.history.pushState({
        cache: {},
        ...page,
      }, '', page.url)
```

reinink:
Considering this resolved, as per my comment here:

https://github.com/inertiajs/inertia/pull/110#issuecomment-705543954

Thanks so much for reporting!

PR #110, opened 2020-01-30
Catch exceptions replacestate pushstate

SabatinoMasala:
Follow-up of https://github.com/inertiajs/inertia/issues/109

Added 2 exception handlers on both `replaceState` and `pushState` with a small warning, to help the developer troubleshoot potential issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /node_modules
+.idea
 .DS_Store
 package-lock.json
 yarn.lock
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -135,15 +135,25 @@ export default {
 
   setState(page, replace = false, preserveState = false) {
     if (replace || page.url === `${window.location.pathname}${window.location.search}`) {
-      window.history.replaceState({
-        ...{ cache: preserveState && window.history.state ? window.history.state.cache : {} },
-        ...page,
-      }, '', page.url)
+      try {
+        window.history.replaceState({
+          ...{ cache: preserveState && window.history.state ? window.history.state.cache : {} },
+          ...page,
+        }, '', page.url)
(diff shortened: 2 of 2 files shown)


SamJakob:
Can this PR be considered for merge?

briannastatic:
Same. Also looking for this to be merged.

claudiodekker:
Hmm.. Personally, I'm not in favour of this change. My reasons are simple:

- Most developers won't ever run into this problem
- It adds an extra maintenance burden / something to keep into account
- **Every user will end up with an increased build size**

Instead, simply by posting this comment and including `NS_ERROR_ILLEGAL_VALUE` in it like I'm doing now, this issue will show up at the very top of Google when someone runs into it. This, by itself, is already sufficient information for most developers to be able to figure out what's going on, especially given the context of the changes you're suggesting.

reinink:
Hey folks!

So, I've been thinking about this change for a while, and while I totally get why you want to add this, I don't think it actually makes sense:

1. I don’t think it’s our job to report browser errors better. If anything this should be reported to Firefox to report on better. 
2. Hiding the original error (`NS_ERROR_ILLEGAL_VALUE`) makes this harder to Google search (if you wanted to do more search into this problem).
3. A general try/catch could hide other errors, unrelated to the size problem.

Plus, if you can already run Google search for "NS_ERROR_ILLEGAL_VALUE inertia" and find these related issues (and the solution—don't pass so much data!). Finally, I've updated the Inertia.js docs (https://github.com/inertiajs/inertiajs.com/commit/d2719574e80b44ea8964f96d705e6ef0e86fefd5) to include this cryptic error in the note we have about response sizes.

Hope that makes sense. Thanks for your contribution either way!

Issue #101, opened 2019-11-18
Browser support


adriandmitroca:
Hey folks,

I think it would be good to have browser support specified within the documentation.

Currently, I've found that unfortunately Inertia.js isn't working out of the box in IE 11 or MS Edge, mostly because missing Promise / scrollTo polyfills. I suppose this could be easily sorted out by including those polyfills out of the box within the package bundles.

What do you think?

reinink:
Interesting, I did not realize this. I would have expected Inertia.js to work with Edge. Is that on all versions? I feel like the library should be compatible with Edge out of the box. For IE11, I agree, a note in the docs would be smart.

What's required to make it compatible with Edge?

adriandmitroca:
As for Edge, for sure scrollTo won't work.

Support: https://caniuse.com/#feat=mdn-api_scrolltooptions_top
Usage: https://github.com/inertiajs/inertia/blob/master/src/inertia.js#L132
Solution: Find out different way to scroll to the top of the page (maybe `document.body.scrollTop`? or polyfill: https://github.com/iamdustan/smoothscroll)

For IE11, simply Promise polyfill would totally do the job, so I would say that's not a big effort from package side, considering it is relatively hard to polyfill external vendor libraries and most people won't even realise this - unless support will be officially fully dropped.

Thanks for super quick response by the way :)

reinink:
> Thanks for super quick response by the way :)

It happens everyone once in a while. 😂 

Thanks so much for this @adriandmitroca, I'll add this to my higher priority list. 👍

marvinrabe:
I got this error in Edge 18.18362:

```
TypeError: Object doesn't support property or method 'scrollTo'
  at Anonymous function (/js/app.js:2:136986)
  at Array.prototype.forEach (native code)
  at t.Inertia.setScroll (/js/app.js:2:136907)
  at Anonymous function (/js/app.js:2:136760)
  at Anonymous function (/js/app.js:2:10079)
...
(1 additional frame(s) were not displayed)
```

reinink:
We have replaced `scrollTo()` with `scrollTop`, so that issue should be fixed.

We have also decided that we are not supporting IE11, which I've added to my documentation TODO list. 👍

PR #116, opened 2020-02-15
Scroll region restoration

### LINKED ISSUE #89: history.back() Scroll position
I wonder if `history.back()` can remember scroll position? Also `window.scrollY` always returns `0` and `window.addEventListener` can't be installed to a component.. probably all listeners don't work. I'm using Vue.
All I want is to remember scroll position at `history.back()` and sometimes to refetch/update page.

**EDIT:**
Also what about infinite scroll? I bet it will be re-rendered at `history.back()`. Maybe it will work if we will use `preserve-state` and `remember: 'page'`, but still no scroll remember ;) - not sure tho.

reinink:
Closes #89

This PR adds scroll region (container) restoration while navigation forward and back in history.

A walkthrough of this can be found here: https://vimeo.com/391530557

This is still a work-in-progress.

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -29,7 +29,7 @@ export default {
     this.saveScrollPositions = debounce(() => {
       this.setState({
         ...window.history.state,
-        scrollRegions: Array.prototype.slice.call(document.querySelectorAll('[scroll-region]')).map(region => {
+        scrollRegions: Array.prototype.slice.call(this.scrollRegions()).map(region => {
           return {
             top: region.scrollTop,
             left: region.scrollLeft,
@@ -45,6 +45,10 @@ export default {
     if (window.performance && window.performance.getEntriesByType('navigation').length) {
       return window.performance.getEntriesByType('navigation')[0].type
     }
+  },
+
+  scrollRegions() {
+    return document.querySelectorAll('[scroll-region]')
   },
 
   isInertiaResponse(response) {
@@ -135,7 +139,7 @@ export default {
         this.version = page.version
         this.setState(page, replace, preserveState)
         this.updatePage(component, page.props, { preserveState }).then(() => {
-          let scrollRegions = document.querySelectorAll('[scroll-region]')
+          let scrollRegions = this.scrollRegions()
 
           scrollRegions.forEach(region => {
             region.addEventListener('scroll', this.saveScrollPositions)
@@ -182,7 +186,7 @@ export default {
           this.setState(this.page)
           this.updatePage(component, this.page.props, { preserveState: false }).then(() => {
             if (this.page.scrollRegions) {
-              document.querySelectorAll('[scroll-region]').forEach((region, index) => {
+              this.scrollRegions().forEach((region, index) => {
                 region.scrollTop = this.page.scrollRegions[index].top
                 region.scrollLeft = this.page.scrollRegions[index].left
               })
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -1,7 +1,7 @@
 import Axios from 'axios'
-import Modal from './modal'
-import Debounce from './debounce'
-import Progress from './progress'
+import debounce from './debounce'
+import modal from './modal'
+import progress from './progress'
 
 export default {
   saveScrollPositions: null,
@@ -26,7 +26,7 @@ export default {
       this.setPage(initialPage)
     }
 
-    this.saveScrollPositions = Debounce(() => {
+    this.saveScrollPositions = debounce(() => {
       this.setState({
         ...window.history.state,
         scrollRegions: Array.prototype.slice.call(document.querySelectorAll('[scroll-region]')).map(region => {
@@ -65,7 +65,7 @@ export default {
   },
 
   visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false, only = [] } = {}) {
-    Progress.start()
+    progress.start()
     this.cancelActiveVisits()
     this.saveScrollPositions()
     let visitId = this.createVisitId()
@@ -90,19 +90,19 @@ export default {
       if (this.isInertiaResponse(response)) {
         return response.data
       } else {
-        Modal.show(response.data)
+        modal.show(response.data)
       }
     }).catch(error => {
       if (Axios.isCancel(error)) {
         return
       } else if (error.response.status === 409 && error.response.headers['x-inertia-location']) {
-        Progress.stop()
+        progress.stop()
         return this.hardVisit(true, error.response.headers['x-inertia-location'])
       } else if (this.isInertiaResponse(error.response)) {
         return error.response.data
       } else if (error.response) {
-        Progress.stop()
-        Modal.show(error.response.data)
+        progress.stop()
+        modal.show(error.response.data)
       } else {
         return Promise.reject(error)
       }
@@ -129,7 +129,7 @@ export default {
 
   setPage(page, { visitId = this.createVisitId(), replace = false, preserveScroll = false, preserveState = false } = {}) {
     this.page = page
-    Progress.increment()
+    progress.increment()
     return Promise.resolve(this.resolveComponent(page.component)).then(component => {
       if (visitId === this.visitId) {
         this.version = page.version
@@ -152,7 +152,7 @@ export default {
 
           this.saveScrollPositions()
         })
-        Progress.stop()
+        progress.stop()
       }
     })
   },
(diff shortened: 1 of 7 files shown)


src/inertia.js
```diff
@@ -119,20 +131,25 @@ export default {
...
+        this.updatePage(component, page.props, { preserveState }).then(() => {
+          let scrollRegions = document.querySelectorAll('[scroll-region]')
+
+          scrollRegions.forEach(region => {
+            region.addEventListener('scroll', this.saveScrollPositions)
+          })
+
+          if (!preserveScroll) {
+            window.scrollTo(0, 0)
+            scrollRegions.forEach(region => region.scrollTo(0, 0))
```
georgehanson: @reinink I think it would probably be better to use `scrollTop` rather than `scrollTo`? This will fix the cross compatibility issues with MS Edge. For example:

```javascript
window.scrollTop = 0
scrollRegions.forEach(region => region.scrollTop = 0)
```

marvinrabe:
Is it possible to use `region.scrollTop` for Edge support? See #101

klaasgeldof:
Hi guys, any plans to merge this? 🙂

src/inertia.js
```diff
@@ -24,6 +26,15 @@ export default {
       this.setPage(initialPage)
     }
 
+    this.saveScrollPositions = Debounce(() => {
+      this.setState({
+        ...window.history.state,
+        scrollRegions: Array.prototype.slice.call(
+          document.querySelectorAll('[scroll-region]')
+        ).map(region => region.scrollTop),
```
klaasgeldof: While we're at it, why not save `scrollLeft` as well?

reinink: Done. 👍

src/inertia.js
reinink: Yup, done. 👍 

//cc @marvinrabe

claudiodekker (approved):
LGTM 👍 

Initially I thought I saw a couple of things that seemed redundant or too complex, but after looking a bit further I quickly understood why they were necessary. Functionally it exactly as expected. Nice one, @reinink!

I did spot one thing that could be (visually) slightly cleaner, so I made a PR for that: https://github.com/inertiajs/inertia/pull/191

reinink:
@claudiodekker Awesome! I'm just heading out now, but will merge that PR later. Dig it 💪

Juhlinus (approved):
Looks good to me!

packages/inertia/src/inertia.js
```diff
@@ -1,8 +1,10 @@
 import Axios from 'axios'
 import Modal from './modal'
+import Debounce from './debounce'
 import Progress from './progress'
 
 export default {
+  saveScrollPositions: null,
```
sebastiandedeyne, line 7: Not 100% sure, but I _think_ you could declare the function here already.

```suggestion
  saveScrollPositions: debounce(() => { /* … */ }),
```

packages/inertia/src/inertia.js
```diff
@@ -53,6 +67,7 @@ export default {
   visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false, only = [] } = {}) {
     Progress.start()
     this.cancelActiveVisits()
+    this.saveScrollPositions()
```
sebastiandedeyne, line 70: Is this still necessary since they're being saved on scroll?

packages/inertia/src/inertia.js
```diff
@@ -117,25 +132,31 @@ export default {
...
-
         this.version = page.version
         this.setState(page, replace, preserveState)
-        this.updatePage(component, page.props, { preserveState })
-        this.setScroll(preserveScroll)
+        this.updatePage(component, page.props, { preserveState }).then(() => {
+          let scrollRegions = document.querySelectorAll('[scroll-region]')
+
+          scrollRegions.forEach(region => {
+            region.addEventListener('scroll', this.saveScrollPositions)
```
sebastiandedeyne, line 141: You could also do the `debounce` here instead of in the function declaration.
```suggestion
            region.addEventListener('scroll', debounce(this.saveScrollPositions, 100))
```

sebastiandedeyne (commented):
Looks good as is! I just added some random comments and ideas.

packages/inertia/src/inertia.js
```diff
@@ -1,8 +1,10 @@
 import Axios from 'axios'
 import Modal from './modal'
+import Debounce from './debounce'
```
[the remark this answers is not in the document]
reinink: Should `Modal` be lowercase too then? 🤔

packages/inertia/src/inertia.js
claudiodekker: No, that won't work, because there's a reference to the current object in it that will fail otherwise:
https://github.com/inertiajs/inertia/blob/5b9eab2e00db8b7b7f17d9b3d3497ee0297373bb/packages/inertia/src/inertia.js#L29-L31

reinink: I tried that originally, but got this error:

```
inertia.js?8c02:8 Uncaught TypeError: Cannot read property 'setState' of undefined
    at Object.r (inertia.js?8c02:8)
    at eval (debounce.js?1b85:5)
```

packages/inertia/src/inertia.js
reinink: I think so, lol. I wrote this months ago, and I think there was an edge case reason for this.

reinink: Maybe I'll remove it, and see if I get any issues reported.

packages/inertia/src/inertia.js
reinink: The issue here is that the old event listeners are only remove if you use the exact same function reference. We need the old event listeners to be removed when you navigate to a new page, otherwise you'll eventually get hundreds of them.

packages/inertia/src/inertia.js
claudiodekker: Did some digging into this. 

Here's an initial page load, followed by scrolling, clicking a link, and clicking the back button:

Screenshot 2020-08-20 at 13 40 59

I also tried causing a race condition by clicking a visit link (emulated 3g), scrolling, and only after hitting the next page. This also worked fine:

Screenshot 2020-08-20 at 13 43 27

My assumption here (I'll check the changelog after placing this comment) is that the this was done to ensure a state existed, so that if you go back to a previous page, the `restoreState` method doesn't accidentally throw an error.

However, because `setPage`'s nested `updatePage` also has a `saveScrollPositions` call, this doesn't happen any longer, which appears to indeed make this specific call obsolete. 

That said, I wouldn't only rely on storing on-scroll, because it's debounced and therefor doesn't always save (e.g. automated test environment, or if you manage to navigate within those 100ms)

reinink: Hey thanks @claudiodekker, I appreciate you digging into this.

> That said, I wouldn't only rely on storing on-scroll, because it's debounced and therefor doesn't always save (e.g. automated test environment, or if you manage to navigate within those 100ms)

I'm pretty sure this was my thinking. Since it really hurt to save the scroll position again on visit, it felt like a good idea. And, I feel like I did run into situations where this was actually needed.

Maybe I'll just leave it as it. 😝

Juhlinus: > Maybe I'll remove it, and see if I get any issues reported.

You madman! 😂

reinink: > You madman! 😂

I've already changed my mind lol 😅

claudiodekker: > Hey thanks @claudiodekker, I appreciate you digging into this.

Sure 👍  I'm on my lunch break anyway 😄
Re: my earlier assumption:
> .. so that if you go back to a previous page, the restoreState method doesn't accidentally throw an error.

I just noticed that there's actually already a guard-statement in place, so I guess that this shouldn't happen anyway?: https://github.com/inertiajs/inertia/blob/54d1a413e4a883afa49609b26a8dd3749e5ebb1e/packages/inertia/src/inertia.js#L188

If that works correctly, then there isn't even a need to manually call `this.saveScrollPositions()` other than in the scroll handler :) I'll make some changes on my end, and will see if that works.

packages/inertia/src/inertia.js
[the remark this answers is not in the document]
reinink: I've made `debounce`, `modal`, `progress` all lowercase, and left `Axios` uppercase. 👍

packages/inertia/src/inertia.js
claudiodekker: Pushed my changes to this in a separate PR, as this one was already merged 😅  https://github.com/inertiajs/inertia/pull/192

PR #120, opened 2020-03-02
fix: Object doesn't support property or method 'scrollTo' in Edge

marvinrabe:
See #101


Landish:
Similar issue appears in Opera browser (desktop) and some of the Android browsers. 

b

reinink:
Thanks for this @marvinrabe! I've got this fixed in #116. 💪

PR #125, opened 2020-03-07
Implements the option to make progress bar optional on requests

francoisauclair911:
Allows to use the page prop partial refreshing feature for AJAX calls that do not require the user to see the progress bar. ( ex: mark a notification as read )

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -50,8 +50,10 @@ export default {
     return this.visitId
   },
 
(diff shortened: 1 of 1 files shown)


claudiodekker:
Hi, as of yesterday's release of Inertia, this is possible using the new https://github.com/inertiajs/progress package. 👍

reinink:
Yup, no longer needed thanks to the new event system! 😍

In fact, NProgress has been completely removed from Inertia.js.

Thanks either for way for this contribution. 🤟

PR #131, opened 2020-04-03
Add Jest and some sample tests

marvinrabe:
I think its time to write tests for Inertia. This will obviously improve the overall quality.

I had to change inertia.js a bit:

- Return Promise from `init` method so that the caller can know when its initialized.
- scrollTo is not available in node.js (and IE 11 ;-) …)

If you have any wishes, please comment on this PR.

Additionally, I configured Github Actions: See https://github.com/marvinrabe/inertia/runs/557242279

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,12 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "node": "current"
+        }
+      }
+    ]
+  ]
+}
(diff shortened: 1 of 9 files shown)


reinink:
Hey thanks so much for tackling this! I'd LOOOOVE to get some tests added to Inertia core. I'm SUPER busy right now, but things open up for me in June. I hope to get back into a bunch of Inertia development then, and this will be on my todo list. 👍

reinink:
Thanks again for this @marvinrabe. However, I'm going to close this in favour of #274. 👍

PR #140, opened 2020-05-20
Allow use of dynamic method

RedFernLiam:
This allows requests where method is dynamic

So now this is possible...

```javascript
this.$inertia.request('post', '/my-endpoint', data)
```

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -180,6 +180,12 @@ export default {
     return this.visit(url, { ...options, method: 'delete' })
   },
 
+  request(method, url, data = {}, options = {}) {
+    let preserveState = method !== 'delete'
+
+    return this.visit(url, { preserveState: preserveState, ...options, method: method, data })
+  },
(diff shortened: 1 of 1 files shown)


m1guelpf:
@RedFernLiam You can currently do the following to use dynamic methods:

``` js
const method = 'post'

this.$inertia[method]('/my-endpoint', data)
```

RedFernLiam:
@m1guelpf I currently use the visit method to accomplish this but I think it would be a nice addition to the API like `axios.request()`

```javascript
this.$inertia.visit('/my-endpoint', {
    method: 'post',
    data: data
})
```

reinink:
This can already be done using the `visit()` method, which I think is sufficient here. 👍 

```js
let method = 'get|post|put|patch|delete'

this.$inertia.visit(url, {
  method: method,
})
```

Thanks either way for this contribution!

PR #141, opened 2020-05-21
Let user add any Axios config

Yiddishe-Kop:
Today I needed a way to track progress of file uploads with Inertia.
I came across @reinink's comment [here](https://github.com/inertiajs/inertia/pull/29#issuecomment-518215064), but looks like it wasn't implemented yet.

So here it is, tried and tested!

Usage:
```javascript
Inertia.post('/path/to/upload', data, {
   onUploadProgress(e) {
      const percentCompleted = Math.round((e.loaded * 100) / e.total);
      console.log(`${percentCompleted}% uploaded`);
    },
})
```

### DIFF
--- a/src/inertia.js
+++ b/src/inertia.js
@@ -50,7 +50,7 @@ export default {
     return this.visitId
   },
 
-  visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false, only = [], onUploadProgress = null } = {}) {
+  visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false, only = [], config = {} } = {}) {
     Progress.start()
     this.cancelActiveVisits()
     let visitId = this.createVisitId()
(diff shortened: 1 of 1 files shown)


src/inertia.js
```diff
@@ -61,7 +61,7 @@ export default {
       data: method.toLowerCase() === 'get' ? {} : data,
       params: method.toLowerCase() === 'get' ? data : {},
       cancelToken: this.cancelToken.token,
-      onUploadProgress,
+      ...config,
```
Yiddishe-Kop, line 64: I refactored this to let the user add any of the [axios config options](https://github.com/axios/axios#request-config).

Usage now:
```javascript
Inertia.post(this.route('files.store'), data, {
  config: {
     onUploadProgress: e => {
      const percentCompleted = Math.round((e.loaded * 100) / e.total);
      console.log(`${percentCompleted}% uploaded`);
    },
  },
})
```

Yiddishe-Kop:
Anybody here?

👀

reinink:
@Yiddishe-Kop I'm so sorry, I've been really far behind on issues and PRs, but slowly catching up. 😅

So my concern here is that there is a pretty good chance we'll eventually drop Axios from Inertia.js. Lots of folks feel like it's too heavy for Inertia, and would prefer to simply use fetch or XHR directly.

That said, file upload progress is definitely missing from Inertia.js right now and needs to be added ASAP. I'm going to keep this PR open as a reminder for that. 👍

reinink:
File upload progress and visit cancellation has been added via #208. 🙌

Going to close this in favour of that.

Thanks either way @Yiddishe-Kop, this was a good thought, but given that we're likely eventually going to remove Axios, I think it's better that we offer our own API for this stuff. 👍

PR #153, opened 2020-07-20
Remove lock files from .gitignore and add package-lock.json file

lamartire:
Hello, @reinink! I didn't find `lock` files inside this project. It is very bad, because without `lock` files it is very difficult to resolve correct dependency version. Also, it is not secure. [The first discussion about it](https://stackoverflow.com/a/39992365) from search. May be there is any reason to this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
 /node_modules
 .DS_Store
-package-lock.json
-yarn.lock
 dist
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6903 @@
+{
+  "name": "@inertiajs/inertia",
+  "version": "0.1.9",
+  "lockfileVersion": 1,
+  "requires": true,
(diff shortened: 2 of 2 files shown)


lamartire:
Of course, I can add `yarn.lock` instead `package-lock.json`

MichaelDeBoey:
I think you shouldn't add them into libraries, only in apps.

reinink:
From my understanding, Yarn and NPM won't use the library lock file. The only benefit to including this would be for library developers to ensure a specific version is being used while building the library, and even that is a debatable topic.

However, as @MichaelDeBoey said, you'll definitely want to use a lock file in your app. 👍

lamartire:
For example, you have library and `axios`'s dependency has security issue. Before `axios` update you can't get different version of vulnerable dependency. I think, it is a good opportunity to solve this moment also on library level, not only in applications.

PR #185, opened 2020-08-05
[inertia-react] Update InertiaLink to support references

hailwood:
Presently there is no way to get a reference to the underlying element created by InertiaLink.

While references are generally discouraged, they are at times necessary for example [tooltips via Tippy.js](https://github.com/atomiks/tippyjs-react#component-children) or using an InertiaLink as an item in a [dropdown menu via reach-ui](https://reach.tech/menu-button/#menulink-as).


reinink:
Pinging @sebastiandedeyne. ❤️

sebastiandedeyne:
💪 Thanks @hailwood, sorry I missed this earlier!

PR #186, opened 2020-08-09
Resolve layouts recursively 

pxlrbt:
(Resubmitted from https://github.com/inertiajs/inertia-vue/pull/135)

The current implementation of nested layouts requires using the render function. This is a) more complex than the the nice layout shorthand and b) requires importing both layouts in every file you use them.

`Page.vue`

```js
import SiteLayout from './SiteLayout'
import NestedLayout from './NestedLayout'

export default {
    layout: (h, page) => {
        return h(SiteLayout, [
            h(NestedLayout, [page]),
        ])
    },
```

For simplicity it would be nice to use the shorthand syntax for nested layouts also.

`InnerLayout.vue:`

```js
import OuterLayout from './OuterLayout'

export default {
    layout: OuterLayout,
```

`Page.vue:`

```js
import InnerLayout from './InnerLayout'

export default {
    layout: InnerLayout,
```

This PR adds the functionality to recursively resolve nested layouts using the shorthand syntax.

### DIFF
--- a/packages/inertia-vue/src/app.js
+++ b/packages/inertia-vue/src/app.js
@@ -51,7 +51,7 @@ export default {
           return this.component.layout(h, child)
         }
 
-        return h(this.component.layout, [child])
+        return this.resolveLayout(h, this.component.layout, child);
       }
 
       return child
@@ -62,5 +62,13 @@ export default {
     Object.defineProperty(Vue.prototype, '$page', { get: () => app.props })
     Vue.mixin(Remember)
     Vue.component('InertiaLink', Link)
+  },
+  methods: {
+    resolveLayout(h, layout, child) {
+        child = h(layout, [child]);
+        layout = layout.layout;
+
+        return layout ? this.resolveLayout(h, layout, child) : child;
+    }
   },
 }
--- a/packages/inertia-vue/src/app.js
+++ b/packages/inertia-vue/src/app.js
@@ -57,7 +57,7 @@ export default {
             .reduce((child, layout) => h(layout, [child]))
         }
 
-        return this.resolveLayout(h, this.component.layout, child);
+        return this.resolveLayoutRecursively(h, this.component.layout, child);
       }
 
       return child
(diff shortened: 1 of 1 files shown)


Juhlinus:
Hi @pxlrbt 👋 !

Nice job on this. (Your code examples are a bit broken, but I understand your intent 😉)

@sebastiandedeyne would this be as straightforward to implement for React?

pxlrbt:
Whoops ... Copying from the other PR messed up the comment a bit. I cleaned it up

reinink:
This is cool, thanks for submitting @pxlrbt. I'm going to do some testing with this. 👍

pxlrbt:
Since [PR 201](https://github.com/inertiajs/inertia/pull/201) was preferred and merge, I think this can be closed. Or is there still any interest in the recursive version?

pxlrbt:
I just merged the master branch and renamed the function name to reflect the recursive approach.

reinink:
Honestly I kind of dig this feature...mostly because that's how I've built up my layouts in the past when I've had multiple layouts.

One question, have you tested this with the new array shorthand (`layout: [OuterLayout, InnerLayout]`)? I'm curious what would happen. 😂

pxlrbt:
> One question, have you tested this with the new array shorthand (layout: [OuterLayout, InnerLayout])? I'm curious what would happen. 😂

I was just thinking about this myself. Not sure how or whether they should be combined?

At the moment there shouldn't be any issue as the array shorthand works on the first level/the pages level. So there shouldn't be any layout components with a layout property. Either the component is resolved using the array shorthand or the recursive approach.

I don't think it's useful to combine both approaches. What do you think?

reinink:
Yeah, mixing them is super weird. I mostly just wonder what would happen if you did. I'll do some testing. 👍

pxlrbt:
I can think of these two cases:

**Case 1:**
```vue
export {
  layout: [Layout1, RecursiveLayout]
```
Will resolve the layout with the array shorthand. Any layout defined in `RecursiveLayout` will be ignored.
**Case 2:**

```vue
export {
  layout: RecursiveLayout
```

`RecursiveLayout.vue`
```vue
export {
  layout: [MoreNestedLayout1, MoreNestedLayout1]
```
Will fail as an array is passed. But shouldn't happen, as the array approach on the second level doesn't work anyway

Can you think of anything else?

pxlrbt:
To prevent case 2 we could stop recursion and print an error message like this:
```js
if (Array.isArray(layout)) {
  console.error('InertiaJS: Array shorthand must not be combined with recursive shorthand.');
  return child;
}
```

pxlrbt:
@reinink Anything I could help you or @claudiodekker with to get this merged?

PR #189, opened 2020-08-19
Simplify debounce

claudiodekker:
Hey @reinink, just saw you commit this one & figured I'd contribute the slightly simpler one that I had laying around 😅


reinink:
Dig it! 🤟

PR #191, opened 2020-08-19
Replace Promise with async

claudiodekker:
Minor simplification for https://github.com/inertiajs/inertia/pull/116

### DIFF
--- a/packages/inertia-react/src/App.js
+++ b/packages/inertia-react/src/App.js
@@ -20,14 +20,12 @@ export default function App({
     Inertia.init({
       initialPage,
       resolveComponent,
-      updatePage: (component, props, { preserveState }) => {
+      updatePage: async (component, props, { preserveState }) => {
         setPage(page => ({
           component,
           key: preserveState ? page.key : Date.now(),
           props: transformProps(props),
         }))
-
-        return Promise.resolve()
       },
     })
   }, [initialPage, resolveComponent, transformProps])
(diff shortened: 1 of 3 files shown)


reinink:
Love it.

reinink:
One question, will this work with IE11? 😬

claudiodekker:
Oof, yeah, no, it won't. Completely forgot about IE11 🙀🙃

reinink:
I’m seriously considering dropping IE11 support. 🤔

pedroborges:
Microsoft itself is [dropping IE11](https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666) support for their services.

claudiodekker:
Yeah, it's worth considering. I was just thinking about asking your thoughts on it as well.

It's a bit of a mixed bag IMO, as Inertia is young & (unfortunately) still small enough to be able to make big decisions like this, but at the same time I can imagine people using it in apps built for institutions where IE is the only option.

Microsoft just announced they're killing support for their own IE11 & Legacy Edge as early as December: https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666

Vue 3 will still be supporting it, but will apparently ship a separate build that has "proxied" support for it built in https://github.com/vuejs/rfcs/issues/183

PR #192, opened 2020-08-20
Improvements to persistent scrolling feature

claudiodekker:
Didn't manage to get this done in time, but here you go :)

- Removes unnecessary save call
- Prevents event listener from being triggered immediately by changing the definition order
- Save current position directly before page navigation change (without a debounced delay)

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -4,7 +4,6 @@ import modal from './modal'
 import progress from './progress'
 
 export default {
-  saveScrollPositions: null,
   onRegionScroll: null,
   resolveComponent: null,
   updatePage: null,
@@ -27,21 +26,20 @@ export default {
       this.setPage(initialPage)
     }
 
-    this.saveScrollPositions = () => {
-      this.setState({
-        ...window.history.state,
-        scrollRegions: Array.prototype.slice.call(this.scrollRegions()).map(region => {
-          return {
-            top: region.scrollTop,
-            left: region.scrollLeft,
-          }
-        }),
-      })
-    }
-
-    this.onRegionScroll = debounce(this.saveScrollPositions, 100)
-
     window.addEventListener('popstate', this.restoreState.bind(this))
+    this.onRegionScroll = debounce(this.saveScrollPositions, 100)
+  },
+
+  saveScrollPositions() {
+    this.setState({
+      ...window.history.state,
+      scrollRegions: Array.prototype.slice.call(this.scrollRegions()).map(region => {
+        return {
+          top: region.scrollTop,
+          left: region.scrollLeft,
+        }
+      }),
+    })
   },
 
   navigationType() {
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -27,7 +27,7 @@ export default {
     }
 
     window.addEventListener('popstate', this.restoreState.bind(this))
-    this.onRegionScroll = debounce(this.saveScrollPositions, 100)
+    this.onRegionScroll = debounce(() => this.saveScrollPositions(), 100)
   },
 
   saveScrollPositions() {
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -185,12 +185,15 @@ export default {
           this.version = this.page.version
           this.setState(this.page)
           this.updatePage(component, this.page.props, { preserveState: false }).then(() => {
-            if (this.page.scrollRegions) {
-              this.scrollRegions().forEach((region, index) => {
+            this.scrollRegions().forEach((region, index) => {
+              region.scrollTop = 0
+              region.scrollLeft = 0
+
+              if (this.page.scrollRegions) {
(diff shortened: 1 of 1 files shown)


reinink:
Okay, you'll have to help me see the benefits here, because I am having a hard time understanding how this improved things. 😂

I am testing this, and you're right, there is no longer a call to `saveScrollPositions()` on page load—that's good. However, now when you click a link it actually calls it twice, not once.

claudiodekker:
Yeah, I see what you mean. I just figured that out as well. 
Apparently the `scrollTop = ... ` causes the event listener to still get triggered, even if it's defined only after that call.

Anyway, the other improvements still work just fine :)

---
Here's a comparison of before (left) & after (right):

#### Not scrolling (load page, click link, click browser back button):
Screenshot 2020-08-20 at 15 32 48

#### Scrolling once (load page, scroll, click link, click browser back button):
Screenshot 2020-08-20 at 15 34 09


As you can see, the functionality remains the same, but it's slightly more economical.
As for the event listener, I'll see if I can fix that as well.

packages/inertia/src/inertia.js
```diff
@@ -116,6 +118,8 @@ export default {
           page.props = { ...this.page.props, ...page.props }
         }
 
+        this.saveScrollPositions()
```
reinink, line 121: Curious, what was the benefit/purpose of moving this from line 74?

claudiodekker: This way, you can still scroll while a page is loading (e.g. on a slow connection)

claudiodekker:
I've actually since figured out that what's actually happening is that when you click the back button, the event listeners from the 'previous' page component are still handling events (and aren't unloaded).

As a result, the functionality is working fine, but it's also causing the listeners to unnecessarily trigger on the `scrollTop = ...` I mentioned earlier. I'll take a look at this later and will update this PR if I find a fix for it.

For now we can keep the implementation as-is, as the functionality is working just fine 👍

reinink:
Yeah okay cool, if you get it figured out, feel free to update. 🤟

Also, something else you can do is move the `saveScrollPositions` out of the `init` function, to its own function.

```js
this.onRegionScroll = debounce(this.saveScrollPositions.bind(this), 100)

// ...

saveScrollPositions() {
  this.setState({
    ...window.history.state,
    scrollRegions: Array.prototype.slice.call(this.scrollRegions()).map(region => {
      return {
        top: region.scrollTop,
        left: region.scrollLeft,
      }
    }),
  })
},

``

**Comment by claudiodekker** [comment]
Alright, so, even with the event listeners from the previous page unloaded, the issue still occurs, so I tried to replicate it outside of Inertia's codebase: https://codepen.io/claudiodekker/pen/pogqOQM

My assumption is that it gets triggered because the JS itself is executed quicker than the browser repaints the DOM, and as a result the listener is already there when `scrollTop = 0` actually gets painted, which then triggers the scroll event even though it is defined after it in the code.

One workaround I found is to wrap a 25ms `setTimeout` call around `addEventListener` (as I couldn't find anything similar to Vue's `$nextTick` for vanilla JS), which prevents this, although I'm not sure that this is the way to go, as it's an arbitrary value I came up with based on absolutely nothing that just so happens to work in (at least) Chrome: https://codepen.io/claudiodekker/pen/KKVbxrj

Anyway, since saving the state more often isn't really a big deal, I'm not sure if it's worth it to implement this workaround? It's fairly safe to do so (as setTimeout 'just works' in AFAIK all browsers), but it's kind of a 'weird' thing to do for this purpose.

File: packages/inertia/src/inertia.js [hunk 2]
```diff
@@ -154,7 +160,7 @@ export default {
             })
           }
 
-          this.saveScrollPositions()
+          scrollRegions.forEach(region => this.addScrollListener(region))
```
claudiodekker, line 163: This would apply the 25ms 'workaround' that prevents the scroll position to be saved instantly on page transition:

```suggestion
          setTimeout(() => scrollRegions.forEach(region => this.addScrollListener(region)), 25)
```

reinink:
Hmm, I'm worried this is getting more complicated, and not really adding a ton of value.

The current (merged) implementation works great. Yes, it does trigger a scroll save on page load, but that honestly doesn't feel like a super big deal.

Also, the currently implementation also saves the current position directly before page navigation change without a debounced delay (by making the extra save call), so that doesn't seem like a super big issue either.

Unless I'm missing something?

claudiodekker:
> Hmm, I'm worried this is getting more complicated, and not really adding a ton of value.

Yeah, I was thinking a similar thing, but wanted to push it a bit more in hopes I'd end up with an elegant solution, but alas 😅 
I'll revert this a bit, and will then re-open it.

reinink:
> Yeah, I was thinking a similar thing, but wanted to push it a bit more in hopes I'd end up with an elegant solution

I really appreciate the efforts! Often that's exactly what you need to do. 😅

> I'll revert this a bit, and will then re-open it.

Great, thanks! 👍

Issue #80, opened 2019-10-04
Move Inertia config out of adapters


reinink:
As we add more features to Inertia itself, we constantly need to update the adapters, as they serve as the entry point to Inertia. However, for most features, the adapter is simply a proxy. I'm starting to think it might be better to push more of this work to Inertia directly, and let the adapters remain a little simpler.

For example, currently we do this:

```js
import Vue from 'vue'
import { InertiaApp } from '@inertiajs/inertia-vue'

let app = document.getElementById('app')

Vue.use(InertiaApp)

new Vue({
  render: h => h(InertiaApp, {
    props: {
      initialPage: JSON.parse(app.dataset.page),
      resolveComponent: name => import(`@/Pages/${name}`).then(module => module.default),
      transformProps: props => { ... },
    },
  }),
}).$mount(app)
```

What I'm proposing:

```js
import Vue from 'vue'
import { Inertia } from '@inertiajs/inertia'
import { InertiaApp } from '@inertiajs/inertia-vue'

Inertia.init({
  initialPage: JSON.parse(app.dataset.page),
  resolveComponent: name => import(`@/Pages/${name}`).then(module => module.default),
  transformProps: props => { ... },
})

let app = document.getElementById('app')

Vue.use(InertiaApp)
new Vue({ render: h => h(InertiaApp) }).$mount(app)

**Comment by reinink** [comment]
An updated concept:

```js
import Vue from 'vue'
import { Inertia } from '@inertiajs/inertia'
import { InertiaApp, InertiaPlugin } from '@inertiajs/inertia-vue'
import { InertiaProgress } from '@inertiajs/progress'

Inertia.init('#app', { resolve: name => require(`./Pages/${name}`).default })
InertiaProgress.init()
Vue.use(InertiaPlugin)

new Vue({
  render: h => h(InertiaApp),
}).$mount('#app')
```

This add a new `InertiaPlugin`, which I think makes more sense than combining this with `InertiaApp`.

I also think we can just remove `initialPage: JSON.parse(app.dataset.page)` entirely in favour of an element selector.

PR #194, opened 2020-08-20
feat(inertia-vue): add `renderInertia` render function

innocenzi:
This PR adds a `renderInertia` export from `@inertia/inertia-vue` that is syntactic sugar for initializing Inertia with Vue.

Currently, we do it this way: 

```js
import Vue from 'vue'
import { InertiaApp } from '@inertiajs/inertia-vue'

Vue.use(InertiaApp)

new Vue({
  render: h => h(InertiaApp, {
    props: {
      // resolveComponent: name => import(`./Pages/${name}`).then(module => module.default), // code-splitting
    },
  }),
}).$mount(app)
```

This PR makes it possible to do it this way:

```js
import Vue from 'vue'
import { InertiaApp, renderInertia } from '@inertiajs/inertia-vue'

Vue.use(InertiaApp)

new Vue({
  render: renderInertia({
    resolve: (name) => require(`@/${name}`), // or "import" for code-splitting
  }),
}).$mount('#app')

```

This is a shorter way of initializing Inertia, which gets rid of things that should basically never change from app to app:

— Retrieving the `app` element
— Getting the `page` data from the `app`'s `dataset`
— Writing the actual render function

The `renderInertia` function takes an object as a parameter:

```js
const defaultConfig = {
  resolve: null,
  app: null,
  path: 'Pages',
}
```

— `resolve` must be a function that return the component
— `path` is the path to the page components
— `app` is the element from which the initial `page` dataset will be extracted

Only `resolve` is required, the other are optional.

Initially, I wanted to get rid of `require`/`import` altogether, but it doesn't seem to be easily possible, so this is the best I could do for reducing the initial boilerplate without removing `new Vue`.

### DIFF
--- a/packages/inertia-vue/.eslintrc.js
+++ b/packages/inertia-vue/.eslintrc.js
@@ -4,6 +4,7 @@ module.exports = {
     es6: true,
   },
   extends: ['eslint:recommended'],
+  parser: 'babel-eslint',
   parserOptions: {
     ecmaVersion: 2018,
     sourceType: 'module',
--- a/packages/inertia-vue/src/index.js
+++ b/packages/inertia-vue/src/index.js
@@ -1,2 +1,3 @@
 export { default as InertiaApp } from './app'
 export { default as InertiaLink } from './link'
+export { renderInertia } from './renderInertia'
--- a/packages/inertia-vue/src/renderInertia.js
+++ b/packages/inertia-vue/src/renderInertia.js
@@ -1,0 +1,29 @@
+import inertia from './app'
+
+const defaultConfig = {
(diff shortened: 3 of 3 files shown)


reinink:
So while I think this is a neat idea, I think I'm okay with a little extra boilerplate at this point, just to keep things 1) flexible, 2) transparent, and 3) with less moving parts.

Plus, I have some ideas in mind that are hopefully going to simplify the boilerplate a little more event. 👍 

Thanks either way @innocenzi!

reinink:
BTW, you can find those new config ideas here: #80. 👍

PR #198, opened 2020-09-01
Pass Vue scoped slots

jasonvarga:
Excuse the huge number of changes in this PR.

### tl;dr

This PR lets you pass along slots from your layout. [It's what this tweet was referring to](https://twitter.com/reinink/status/1300484695323443201).

### More Context

This would be useful when you need to pass along non-Inertia Vue stuff.

My use case is wanting to add Inertia to an app that's currently a standard Laravel app returning Blade views. We have an addon ecosystem where third party developers can be returning views into our admin panel. We wanted a way to convert as much as possible to Inertia, but allow non-Inertia responses to keep working as they did pre-Inertia.

There's a handful of things you need to change about how you wire up Inertia. We talked about trying to streamline this eventually, but we can leave that for a separate PR.

1. Rather than using the `@inertia` directive you'd need to manually put the div in your layout.
2. Rather than using a div, it would need to be another Vue component. When the root Vue instance mounts to the page, it would normally completely replace the element, losing any child elements. By using a component, the slot content can be maintained.
3. The component would be where your Inertia wiring would go instead of your main js file.
4. Your Vue installation needs to include the template compiler.

``` blade
{{-- layout with a single slot --}}
<inertia-wrapper id="app" :page="{{ json_encode($page) }}">
  <div>The default slot</div>
</inertia-wrapper>
```
``` blade
{{-- layout with multiple slots --}}
<inertia-wrapper id="app" :page="{{ json_encode($page) }}">
  <template v-slot:default>
    <div>The default slot</div>
  </template>
  <template v-slot:custom="slotProps">
    <div>A named slot {{ slotProps.foo }}</div>
  </template>
</inertia-wrapper>
```

``` js
// app.js
import { InertiaApp } from '@inertiajs/inertia-vue'
import Vue from 'vue'
import InertiaWrapper from './InertiaWrapper'

Vue.use(InertiaApp)

new Vue({
  components: {
    InertiaWrapper
  }
}).$mount(app)
```

``` vue
// InertiaWrapper.vue
<script>
import { InertiaApp } from '@inertiajs/inertia-vue'

export default {

    props: {
        page: Object
    },

    render(h) {
        return h(InertiaApp, {
            props: {
                initialPage: this.page,
            },
            scopedSlots: this.$scopedSlots
        });
    }

}
</script>
```

```vue
// A page, like Dashboard.vue
<template>
  <div>
    <slot />
    <slot name="custom" :foo="bar" />
  </div>
</template>

<script>
export default {
  data() {
    return {
      bar: 'baz'
    }
  }
}
</script>
```

### DIFF
--- a/packages/inertia-vue/src/app.js
+++ b/packages/inertia-vue/src/app.js
@@ -44,6 +44,7 @@ export default {
       const child = h(this.component, {
         key: this.key,
         props: this.props,
+        scopedSlots: this.$scopedSlots,
       })
 
       if (this.component.layout) {


reinink:
Very cool, thanks @jasonvarga!

pkboom:
@jasonvarga Hi, this is awesome. But while I'm trying to implement this, I got this problem. I did exactly what you showed above and I returned view like this.
```
Route::get('/', function () {
    return view('dashboard', [
        'page' => [
            'name' => 'Dashboard',
        ],
    ]);
});
```

```
// dashboard.blade.vue
<html>
    <head>
        <script src="{{ mix('js/app.js') }}" defer></script>
    </head>
    <body>
        <inertia-wrapper id="app" :page="{{ json_encode($page) }}">
            <div>The default slot</div>
        </inertia-wrapper>
    </body>
</html>
```

Problem:
```
[Vue warn]: You are using the runtime-only build of Vue where the template compiler is not available. Either pre-compile the templates into render functions, or use the compiler-included build.
```

Any help would be much appreciated. Thanks.

reinink:
@pkboom You need to remove this from your webpack config:

```js
vue$: 'vue/dist/vue.runtime.esm.js',
```

Since that doesn't include the template compiler. 👍

pkboom:
That was what I suspected. But removing it was a little scary. Thanks, man. Big help.

pkboom:
I git it working. Thanks. But somehow `undefined` is appended to the url.

Peek 2020-09-24 16-49

This is how view is returned:
```
Route::get('/', function () {
    return view('dashboard', [
        'page' => [
            'component' => 'Dashboard',
        ],
    ]);
});
```
If you could help me with this, I'd really appreciate it. :+1:

donmbelembe:
this needs to be documented

PR #202, opened 2020-09-06
Feature: Ability to define custom headers on a per-request basis

claudiodekker:
This PR allows the user to set additional (custom) request headers on a request-by-request basis.
It checks whether the 'headers' object is set on the request 'options', and if it is, combines it with Inertia's headers.

Do note that Inertia's headers take priority and cannot be overwritten.

Usage:
- `this.$inertia.visit({ headers: { 'Custom-Header': 'value' }})`
- `this.$inertia.post('/endpoint', { 'id': 1 }, { headers: { 'Custom-Header': 'value' } })`
- ... (etc.)

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -68,7 +68,7 @@ export default {
     return this.visitId
   },
 
-  visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false, only = [] } = {}) {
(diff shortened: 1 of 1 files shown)


claudiodekker:
Well this one was unexpected 😅🤼‍♂️

reinink:
Hey thanks for this @claudiodekker. While I don't personally need this, it appears others have, and since it's such a small addition to the library, it seems worth adding. 👍

PR #203, opened 2020-09-07
[Svelte] Add support for persistent layout

pedroborges:
In order to support the same flexibility as the other adapter, I had to create a basic `h` function that can be used in two ways:

```js
/* 1 */ h(Layout, { title: 'Homepage' }, [page])
/* 2 */ h(Layout, [page])
```

Merging this PR is a breaking change. It requires that the `resolveComponent` closure passed to `InertiaApp` returns the page module and not just the default export.

```diff
import { InertiaApp } from '@inertiajs/inertia-svelte'

const app = document.getElementById('app')

new InertiaApp({
  target: app,
  props: {
    initialPage: JSON.parse(app.dataset.page),
    resolveComponent: name =>
-      import(`@/Pages/${name}.svelte`).then(module => module.default),
+      import(`@/Pages/${name}.svelte`),
  },
})
```

Page components can then declare the desired layout inside the `<script context="module">` block:

```svelte
<script context="module">
  import Layout from '@/Shared/Layout.svelte'
  export const layout = (h, page) => h(Layout, { title: 'Homepage' }, [page])
</script>
```

Here the `h` function can be used in any way described above.

Alternatively you also simply assign the layout component to the `layout` export:

```svelte
<script context="module">
  import Layout from '@/Shared/Layout.svelte'
  export const layout = Layout
</script>
```

The Svelte version of the PingCRM has a [persistent-layout](https://github.com/inertiajs/pingcrm-svelte/tree/persistent-layout) branch that can be used for testing this feature.

### DIFF
--- a/packages/inertia-svelte/src/App.svelte
+++ b/packages/inertia-svelte/src/App.svelte
@@ -20,11 +20,11 @@
     },
   })
 
-  $: page = $store.component && h($store.component.default, $store.props)
+  $: child = $store.component && h($store.component.default, $store.props)
   $: layout = $store.component && $store.component.layout
   $: components = layout
-    ? layout.name === 'layout' ? layout(h, page) : h(layout, [page])
-    : page
+    ? h(layout, {}, [child])
+    : child
 </script>
 
 <Render {...components} />
--- a/packages/inertia-svelte/src/App.svelte
+++ b/packages/inertia-svelte/src/App.svelte
@@ -23,7 +23,9 @@
   $: child = $store.component && h($store.component.default, $store.props)
   $: layout = $store.component && $store.component.layout
   $: components = layout
-    ? h(layout, {}, [child])
+    ? Array.isArray(layout)
+      ? layout.concat(child).reverse().reduce((child, layout) => h(layout, {}, [child]))
+      : h(layout, {}, [child])
     : child
 </script>
 
--- a/packages/inertia-svelte/src/App.svelte
+++ b/packages/inertia-svelte/src/App.svelte
@@ -1,6 +1,7 @@
 <script>
   import { Inertia } from '@inertiajs/inertia'
   import store from './store'
+  import Render, { h } from './Render.svelte'
 
   export let
     initialPage,
@@ -18,8 +19,12 @@
       }))
     },
   })
+
+  $: page = $store.component && h($store.component.default, $store.props)
+  $: layout = $store.component && $store.component.layout
+  $: components = layout
+    ? layout.name === 'layout' ? layout(h, page) : h(layout, [page])
+    : page
 </script>
 
-{#each [$store.key] as key (key)}
-<svelte:component this={$store.component} {...$store.props} />
-{/each}
+<Render {...components} />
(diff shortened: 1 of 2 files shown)


reinink:
I wonder @pedroborges, what if we simply didn't offer the `h()` function like we do in Vue? What if we simply offered the shorthand versions:

```js
export const layout = Layout
export const layout = [SiteLayout, NestedLayout] // just added to Vue
```

That seems reasonable to me, since Svelte doesn't even ship with an `h()` function to begin with. I feel like this would support 99% of the persistent layout use-cases. Like, I doubt I'll ever even use the `h()` function myself in my apps, but will simply use these shorthand versins.

Does that simplify things a little?

pedroborges:
I didn't have a good feeling about adding the `h` function to the Svelte adapter, to be honest. I really liked the short array syntax when I saw it this week. Supporting only the direct assignment and the short array syntax will simplify things a lot.

The only downside is not being able to pass props to the layout for things like the title, as an example. Should we automatically pass all props coming from the server adapter to the layout?

I'll update this PR and test it.

reinink:
@pedroborges So, I don't know about Svelte, but with Vue.js (and maybe React too), you can't pass the props reactively from the page component to the parent component when working with persistent layouts...it just doesn't work. That's only possible with the layout is a child component of the page component.

Instead you need some type of event system. For that reason, I kind of think this is a non-issue, and doing that should be discouraged, unless, of course, that's actually possible in Svelte...although I suspect it's not.

Part of me almost wants to remove the callback option from the Vue.js and React adapters, but that might be going too far.

pedroborges:
Forget what I said, it is not an issue at all. Turns out the layout can export a store that can be used by page components to pass data back to the layout.

```svelte
<!-- Layout.svelte -->

<script context="module">
  import { writable } from 'svelte/store'
  export const title = writable(null)
</script>
```

```svelte
<!-- SomePage.svelte -->

<script context="module">
  import Layout, { title } from '@/Shared/Layout.svelte'
  export const layout = Layout
</script>

<script>
  $title = 'Dashboard'
</script>
```

This is not related to Inertia, so I'm not sure we need to document it.

reinink:
That's awesome! 👏

Might be worth documenting, just to help other Svelte users, but I'll leave that up to you. 👍

pedroborges:
@reinink I used [this approach](https://github.com/inertiajs/pingcrm-svelte/commit/42564ae854ba5b1fd47a8a1d9e1615e68340363f) on the Svelte PingCRM 💪

I have removed the layout function syntax and added the short array one. The assignment syntax remains the same.

The `h` is still useful for internal use. I have simplified it to handle just our use case.

Everything looks good on my side. The Svelte PingCRM [`persistent-layout branch`](https://github.com/inertiajs/pingcrm-svelte/tree/persistent-layout) branch has been updated. Feel free to use it for testing.

reinink:
Awesome Pedro! 👏 

And just to be absolutely sure, there is no way to do this without removing the `.then(module => module.default)` part?

The only reason I ask is because this, right now, is consistent between Vue.js, React and Svelte. It would be *really* nice if it could remain that way...plus then no breaking changes.

Why exactly is this change required again? 😬

pedroborges:
In Svelte, a component script/html/css is the `default` export. The `<script context="module">` block allows us to export more stuff like we are doing with `layout`.

If we return the component `default` export from `resolveComponent` there's no way to access other exports later in the adapter `App.svelte` component.

PR #204, opened 2020-09-09
Add headers prop to InertiaLink components

pedroborges:
Declare new `headers` prop on the `InertiaLink` component for all adapters.

### DIFF
--- a/packages/inertia-vue/src/link.js
+++ b/packages/inertia-vue/src/link.js
@@ -30,6 +30,10 @@ export default {
     only: {
       type: Array,
       default: () => [],
+    },
+    headers: {
+      type: Object,
+      default: () => {},
     }
   },
   render(h, { props, data, children }) {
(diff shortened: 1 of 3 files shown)


claudiodekker (approved):

reinink:
Thanks @pedroborges! 👍

packages/inertia-vue/src/link.js
```diff
@@ -30,6 +30,10 @@ export default {
     only: {
       type: Array,
       default: () => [],
+    },
+    headers: {
+      type: Object,
+      default: () => {},
```
reinink, line 36: Realize now that this is actually wrong. It needs to be:

```js
default: () => ({}),
```

Right now this results in a default of `undefined`.

I'll fix. 👍

pedroborges: Ops, sorry about that 🥺

reinink: Haha all good brother! I didn't catch it either. 😊

PR #206, opened 2020-09-09
[Svelte] Add use:inertia directive

pedroborges:
The `use:inertia` directive can be used instead of the `InertiaLink` component to add Inertia behavior to any element in the DOM.

```svelte
<script>
  import { inertia } from '@inertiajs/inertia-svelte'
</script>

<a href="/users" use:inertia>Users</a>

<button
  use:inertia={{ href: '/users/100', method: 'post', data: { '_method': 'delete' }}}
  class="btn-indigo"
  on:click={() => alert('User deleted')}>
  Delete
</button>

<a
  href="/users/100"
  use:inertia={{ method: 'post', data: { '_method': 'put', 'visited': true }}}>
  Visited
</a>
```

### DIFF
--- a/packages/inertia-svelte/package.json
+++ b/packages/inertia-svelte/package.json
@@ -17,6 +17,6 @@
   },
   "peerDependencies": {
     "@inertiajs/inertia": "^0.2.0",
-    "svelte": "^3.8.0"
+    "svelte": "^3.20.0"
   }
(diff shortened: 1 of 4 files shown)


reinink:
Awesome work brother! You're making the Svelte adapter amazing! 🙌

PR #207, opened 2020-09-09
Add v-inertia directive

claudiodekker:
Screenshot 2020-09-09 at 20 09 21

Usage:

- `<a href="/link" v-inertia>Click me!</a>`
- `<a href="/logout v-inertia:post>Logout</a>`
- `<a href="/update" v-inertia:patch="{ title: form.title }>Save Changes</a>`
- `<a href="/" v-inertia="{ options: { ... } }">Visit with request options</a>`

Under the hood, this feature uses the existing [request methods](https://inertiajs.com/requests), and acts as an alternative for `<inertia-link>`.

### DIFF
--- a/packages/inertia-vue/src/directive.js
+++ b/packages/inertia-vue/src/directive.js
@@ -11,7 +11,7 @@ export default {
 
       const method = binding.arg || 'get'
       const options = (binding.value || {}).options || {}
-      const data = (binding.value || {}).data || (! options ? binding.value : {})
+      const data = (binding.value || {}).data || (Object.keys(options).length === 0 ? binding.value : {})
       const url = ((vnode.data || {}).attrs || {}).href || (binding.value || {}).href || (binding.value || {}).url
 
       if (method === 'replace') {
--- a/packages/inertia-vue/src/directive.js
+++ b/packages/inertia-vue/src/directive.js
@@ -1,0 +1,36 @@
+import { Inertia, shouldIntercept } from '@inertiajs/inertia'
+
+export default {
+  bind(el, binding, vnode) {
+    el.addEventListener('click', event => {
+      if (!shouldIntercept(event)) {
+        return
+      }
+
+      event.preventDefault()
+
+      const method = binding.arg || 'get'
+      const options = (binding.value || {}).options || {}
+      const data = (binding.value || {}).data || (! options ? binding.value : {})
+      const url = ((vnode.data || {}).attrs || {}).href || (binding.value || {}).href || (binding.value || {}).url
+
+      if (method === 'replace') {
+        Inertia.replace(url, options)
+      } else if (method === 'reload') {
+        Inertia.reload(options)
+      } else if (['post', 'put', 'patch'].indexOf(method) > -1) {
+        Inertia[method](url, data, options)
+      } else if (method === 'delete') {
+        Inertia.delete(url, options)
+      } else {
+        Inertia.visit(url, {
+          ...options,
+          method,
+          data,
+        })
+      }
+
+      return false
+    })
+  },
+}
--- a/packages/inertia-vue/src/app.js
+++ b/packages/inertia-vue/src/app.js
@@ -1,4 +1,5 @@
 import { Inertia } from '@inertiajs/inertia'
+import Directive from './directive'
 import Link from './link'
 import Remember from './remember'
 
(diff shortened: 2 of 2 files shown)


lostdesign:
I am wondering what the use case would be for this? Or why is there a need to use the directive over `inertia-link`?

claudiodekker:
@lostdesign Essentially, it works exactly the same but just uses a different (more vue-ish/non-component) syntax.

lostdesign:
Well it's a nice addition for sure I guess. The patch one baffles me a bit though, when would you need to pass data directly into the link to patch? Wouldn't you have a method rather that does an $inertia.patch?

claudiodekker:
No idea, but it exists for inertia-link as well, right? I'm just trying to keep the API's the same.

lostdesign:
OH TIL then :D Well it's a nifty addition then as I mentioned.

shaffe-fr:
It's a nice addition. Inertia svelte as a similar feature https://github.com/inertiajs/inertia/pull/206.

claudiodekker:
Thanks for the feedback everyone!

So far, more than half of the responses (including Twitter and our Discord) mention they would not use this feature, with roughly ~25% giving positive feedback. 

Because of this reason, as well as to focus on more important things and to not end up in a situation where we'd be maintaining 5 different ways to do the same thing, we've decided to place this idea back on the shelf for now.

Zzombiee2361:
@claudiodekker Please reconsider reopening this pr. This is the perfect way to use inertia with vuetify. I can't just drop `<inertia-link>` anywhere between vuetify tag, it will break the visuals.

For example, putting inertia-link inside list:
```vue
<v-list dense>
  <inertia-link :href="route('admin.dashboard')">
    <v-list-item>
      <v-list-item-action>
        <v-icon>mdi-view-dashboard</v-icon>
      </v-list-item-action>
      <v-list-item-content>
        <v-list-item-title>Dashboard</v-list-item-title>
      </v-list-item-content>
    </v-list-item>
  </inertia-link>

  <inertia-link :href="route('admin.news.index')">
    <v-list-item>
      <v-list-item-action>
        <v-icon>mdi-newspaper</v-icon>
      </v-list-item-action>
      <v-list-item-content>
        <v-list-item-title>News</v-list-item-title>
      </v-list-item-content>
    </v-list-item>
  </inertia-link>
</v-list>
```
image
The list is underlined, the hover and ripple effect is gone.

Again, please reconsider reopening this pr.

reinink:
Happy to review this.

@claudiodekker Would it make sense to abstract some of this behaviour, to be shared by both the Inertia link component and this new directive?

We'll also need to add the new event callbacks.

packages/inertia-vue/src/directive.js
```diff
@@ -0,0 +1,36 @@
...
+      if (!shouldIntercept(event)) {
+        return
+      }
+
+      event.preventDefault()
+
+      const method = binding.arg || 'get'
+      const options = (binding.value || {}).options || {}
+      const data = (binding.value || {}).data || (Object.keys(options).length === 0 ? binding.value : {})
+      const url = ((vnode.data || {}).attrs || {}).href || (binding.value || {}).href || (binding.value || {}).url
```
Zzombiee2361, line 15: One thing to add, I manually import this directive and this returns `undefined`. I had to change this to `el.href`

claudiodekker: Can you post an example of what you were trying exactly? I'll try to replicate it later, thanks!

claudiodekker:
@reinink Yeah, I can definitely look at doing that..!

packages/inertia-vue/src/directive.js
nickknissen: `el.href` needs to be added to line 15 or it won't work for this example

```
 <template>
    <v-navigation-drawer
      class="teal darken-3"
      dark
      permanent
    >
      <v-list>
        <v-list-item
          v-for="item in items"
          :key="item.title"
          link
          :href="item.route"
          v-inertia
        >
          <v-list-item-icon>
            <v-icon>{{ item.icon }}</v-icon>
          </v-list-item-icon>

          <v-list-item-content>
            <v-list-item-title>{{ item.title }}</v-list-item-title>
          </v-list-item-content>
          
        </v-list-item>
      </v-list>
    </v-navigation-drawer>
</template>
<script>
export default {
  data: () => ({
    items: [
      { title: "Dashboard", route: this.$route('dashboard.index'), icon: 'mdi-view-dashboard'},
      { title: "Organizations", route: this.$route('organizations.index'), icon: 'mdi-file-tree'},
    ]

  }),
};
</script>
```

`const url = ((vnode.data || {}).attrs || {}).href || (binding.value || {}).href || (binding.value || {}).url || el.href;`

reinink:

Issue #169, opened 2020-03-10
[inertia-vue] How to configure Google Analytics in Laravel Vue Inertia SPA?


nextifier:
This is my first time building a Laravel-Vue SPA with Inertia JS.
Because this is a SPA, Google Analytics can't detect any page visits when user navigating the pages.

In Vue Router, the configuration is something like this:

```
import Vue            from 'vue';
import router           from './routes.js'

new Vue({
    router,
}).$mount('#app')


ga('set', 'page', router.currentRoute.path);
ga('send', 'pageview');

router.afterEach(( to, from ) => {
  ga('set', 'page', to.path);
  ga('send', 'pageview');
});
```

How do you do this in Inertia JS?

Note: I'm using Ziggy library to use Laravel routes in my Vue app, so I can use route() helper, but I have no idea how to use it to send the pageview to Google Analytics.

reinink:
The new event system (#208) has been merged in, which makes this a piece of cake using the `navigate` event!

```js
Inertia.on('start', event => {
  // Track visit via Google Analytics
})
```

I hope to tag a release tomorrow, and I also plan to add documentation to the website explaining how to use it. 🙌

Issue #159, opened 2019-09-27
[inertia-vue] Listening to & intercepting visit (navigation) events


baaskoen:
If you were to implement a feature like warning a user before exiting a page, how would you go about and implement this in Inertia?

If the user navigates back or clicks on a `<inertia-link />`, the "onbeforeunload" event is not actually called. When a user navigates back in history, the "popstate" event seems to be called indefinitely.

Are there any events that I can subscribe my listeners to?

reinink:
The new event system (#208) has been merged in, which allows you to listen for visit events and cancel them. I hope to tag a release tomorrow, and I also plan to add documentation to the website explaining how to use it. 🙌

Issue #66, opened 2019-08-25
Inertia emitted events


theblindfrog:
More of a question, than an issue per se... 

A couple of previous apps I've worked on have used turbolinks and vue - pretty much as @reinink described in his [article about backend apps with client rendering](https://reinink.ca/articles/server-side-apps-with-client-side-rendering)

What's useful about using turbolinks in this approach, is that it [emits a heap of events](https://github.com/turbolinks/turbolinks/blob/17ffb3ad2a7cf1987edacab65ea7339351193ce7/src/controller.ts#L236-L262), that your application can hook in to - for example - fire off analytics tracking, etc.

I think there might be merit for inertia.js to emit some of its own events for other parts of applications to latch on to.

For example, you could listen for events like `click`, `visit`, `hardvisit`, `restore`, under an `inertia` namespace:

```javascript
document.addEventListener('inertia:visit', (event) => {
    // do something, for example: 
    dataLayer.push({
      'url': event.data.path, // e.g. `/user/1/update`
      'event': event.data.trigger // e.g `updateProfile`
    });
});
```
If this is something that folks would find useful, I might be able to hack on a PR to add such a feature. 

Any thoughts?

Juhlinus:
@theblindfrog This seems like cool feature indeed!

We should wait for feedback from @reinink and @sebastiandedeyne on this, but I don't see any reason why this wouldn't be good feature for the framework. 😄

sebastiandedeyne:
I like this idea, maybe we could even use it internally to deal with the loading indicator (it would also provide a way for users to implement custom loading indicators)

theblindfrog:
@sebastiandedeyne That's an _even better_ use-case for something like this. 

After more thought, what I'm trying to do could be achieved in the 'layout' layer of my app (or most apps, fwiw) - firing off a new pageview to analytics as new frontend component(s) are created.

Depending on what @reinink's view is, the original thrust of this issue might not be relevant; even if the 'loading indicator' use-case may warrant another look.

reinink:
So, events are certainly on my radar. However, I can't decide if they offer any real benefit over simply providing additional configuration options, either at the adapter level, or even with Inertia core.

theblindfrog:
Fair point @reinink. Perhaps they're the 'easy' way to offer some of this functionality; in contrast to more elegant implementations.

I'll close this issue. Thanks!

ellefsen:
[...]

PR #208, opened 2020-09-09
New event system

reinink:
This PR adds a new event system to Inertia.js, allowing developers to "hook into" the various lifestyle events of the library. Most events are cancelable, letting you override Inertia's default behaviour.

## The API

```js

Inertia.on('start', event => {
  console.log(`starting a visit to ${event.detail.request.url}`)
})
```

What's really cool here is that when you register an event, Inertia automatically returns you a callback to unregister the event. Combined with an unmounted/destroyed hook, you can do something like this to automatically unregister the event when the page component is removed.

```js
export default {
  mounted() {
    this.$once('hook:destroyed', Inertia.on('start', event => {
      return confirm('Are you sure you want to navigate away?')
    }))
  },
}
```

You can register multiple event listeners of the same type. They all will run, and if just one of them calls `event.preventDefault()`, the visit will be stopped.

Also worth noting that under the hood this uses native browser events, so you can also work with them that way. For example:

```js
document.addEventListener('inertia:start', event => {
  console.log(`starting a visit to ${event.detail.request.url}`)
})
```

## Goals

- Allow developers to use their own loading indicators.
- Navigation tracking (ie. for analytical purposes).
- Request error handling (ie. network interruption).

## Global events

| Name        | Description                                                                                                                                                                                                                                                                                   | Detail              | Cancelable                                 |
|-------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------|--------------------------------------------|
| `start`     | Fires before an XHR request is made to the server. Useful for displaying loading indicators (#9, #66) and for listening to and intercepting visits (#152, #159).                                                                                                                              | `e.detail.request`  | Yes, prevents visit from being made.       |
| `progress`  | Fires whenever a progress event is emitted during a file upload.                                                                                                                                                                                                                              | `e.detail.progress` | No                                         |
| `success`   | Fires on successful Inertia responses.                                                                                                                                                                                                                                                        | `e.detail.response` | Yes, prevents the page from being swapped. |
| `invalid`   | Fires when a non-Inertia response is received, for all response levels (200, 400 and 500).                                                                                                                                                                                                    | `e.detail.response` | Yes, prevents the modal from being shown.  |
| `error`     | Fires on unexpected XHR errors (such as network interruptions) and for errors generated in the `resolveComponent()` callback. Will NOT fire for XHR requests that receive a response (even if they are 400/500 level responses).                                                              | `e.detail.error`    | Yes, prevents the error from being thrown. |
| `finish`    | Fires after an XHR request is made to the server, for both successful and unsuccessful responses. Useful for hiding loading indicators.                                                                                                                                                       | N/A                 | No                                         |
| `navigate`  | Fires when a new state is pushed into the history (`history.pushState`), or when navigating history (`popstate`). Does not fire for `history.replaceState` events. These are considered "successful" page navigation events, useful for tracking analytics, and things like that (#66, #169). | `e.detail.page`     | No                                         |

## Per-visit event callbacks

In addition to the global events, this PR adds a number of per-visit event callbacks.

```js
Inertia.post('/users', data, {
  onCancelToken: cancelToken => {},  // Called before the request starts.
  onStart: () => {},                 // Called when the request starts.
  onProgress: progress => {},        // Called when the progress event is fired.
  onFinish: () => {},                // Called when the request is done.
  onCancel: () => {},                // Called when the request is cancelled.
  onSuccess: () => {},               // Called when the request successfully finishes.
})
```

## Other changes

- This PR deprecates the promise that is returned from `Inertia.visit()`, and replaces it with a number of event callbacks that are configurable on a visit. If you try calling `then()` or `catch()` on an Inertia visit, you'll get the following warning:
    > Inertia.js visit promises have been deprecated and will be removed in a future release. Please use the new visit event callbacks instead.
    > 
    > Learn more at https://inertiajs.com/events
- This PR removes the direct NProgress integration, since the new event system makes this so easy to do outside of Inertia core. This behaviour is now available via a first-class package called [@inertiajs/progress](https://github.com/inertiajs/progress).
- While working on this PR, I inadvertently found a way to improve scroll region handling, so this PR includes some refactoring around that.


packages/inertia/src/inertia.js
```diff
@@ -68,7 +68,16 @@ export default {
...
 
-  visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false, only = [], headers = {}} = {}) {
+  visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false, only = [], headers = {} } = {}) {
+    let event = new CustomEvent('inertia:start', {
+      cancelable: true,
+      detail: { url: url, ...arguments[1] }
+    })
+
+    if (!document.dispatchEvent(event)) {
+      return
```
claudiodekker, line 78: Not sure if this can really be considered a bug, but, returning void causes an inconsistent return type.

Example: PingCRM's login page chains a `then` on the end of `this.$inertia.post(..)`, which (when interrupted by an event listener) causes the following error:
```js
"TypeError: Cannot read property 'then' of undefined"
```

```suggestion
      return Promise.reject()
```

With this, it'd still fail the `.then`, but at least it would consistently return a promise.

packages/inertia/src/inertia.js
```diff
@@ -89,28 +126,35 @@ export default {
...
+      }
+      return response.data
+    }).catch(error => {
+      if (this.isInertiaResponse(error.response)) {
         return error.response.data
+      } else if (Axios.isCancel(error)) {
+        return
+      } else if (this.isHardVisit(error.response)) {
+        this.hardVisit(error.response.headers['x-inertia-location'])
+        return
```
claudiodekker, line 146:
```suggestion
```

packages/inertia/src/inertia.js
```diff
@@ -227,12 +246,19 @@ export default {
     let newState = { ...window.history.state }
     newState.cache = newState.cache || {}
     newState.cache[key] = data
-    this.setState(newState)
+    this.setState(newState, true)
```
claudiodekker, line 249: Maybe we can create an replaceState alias for this?

claudiodekker (approved):

claudiodekker (approved):

packages/inertia/src/inertia.js
```diff
@@ -68,132 +102,145 @@ export default {
...
-    progress.start()
+  visit(url, {
+    method = 'get',
+    data = {},
+    replace = false,
+    preserveScroll = false,
+    preserveState = false,
+    only = [],
+    headers = {},
+    onCancelToken = () => ({}),
```
claudiodekker, line 113: If we allow the cancel token to be retrieved, wouldn't it also make sense to allow setting it?

reinink: What's the use-case here?

claudiodekker (approved):

packages/inertia/src/inertia.js
claudiodekker: Preloading ;P

reinink: Ahh yes, understood. Do you want to PR that maybe once this is done...since it's lower urgency?

claudiodekker: Sure 👍

PR #211, opened 2020-09-11
Fix wrong typescript definition for the delete method (no data)

### LINKED ISSUE #197: Wrong syntax for delete?
I just noticed that the Inertia.delete method expects (url, data, options). But when actually using it it, is instead  (url, options). The docs even state this as well

```
Inertia.delete(url, {
  replace: false,
  preserveState: false,
  preserveScroll: false,
  only: [],
})
```
Here is how it is shown in phpstorm 
image

And index.d.ts
```
delete: (url: string, data?: object, options?: VisitOptions) => Promise<void>

wazum:
Fixes #197

### DIFF
--- a/packages/inertia/index.d.ts
+++ b/packages/inertia/index.d.ts
@@ -53,7 +53,7 @@ interface Inertia {
 
   put: (url: string, data?: object, options?: VisitOptions) => Promise<void>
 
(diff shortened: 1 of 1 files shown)


reinink:
Thanks @wazum!

PR #214, opened 2020-09-12
[inertia-vue] Fix remember key of this binding

ycs77:
I follow the documentation [Multiple components](https://inertiajs.com/local-state-caching#multiple-components) to set the remember key:

```js
  remember: {
    data: ['form'],
    key: () => `Post/Edit:${this.post.id}`,
  },
```

But it return an error:

inertia-remember-bug

I found is the `this` binding problem, this PR fixes it.

After using this PR to fix, arrow function cannot be used, can work with the following usage:

```js
  remember: {
    data: ['form'],
    key() {
      return `Post/Edit:${this.post.id}`
    },
  },
```

### DIFF
--- a/packages/inertia-vue/src/remember.js
+++ b/packages/inertia-vue/src/remember.js
@@ -19,7 +19,7 @@ export default {
     }
 
     const stateKey = this.$options.remember.key instanceof Function
-      ? this.$options.remember.key()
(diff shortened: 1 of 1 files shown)


reinink:
Nice catch @ycs77! ❤️

ycs77:
@reinink Document need to update

PR #215, opened 2020-09-13
feat(vue): Inject Props

KABBOUCHI:
Add option to auto inject props on pages.

```js
// ./app.js
new Vue({
    render: h => h(InertiaApp, {
        props: {
            injectProps: true
        },
    }),
}).$mount(app)
```

image

image

### DIFF
--- a/packages/inertia-vue/src/app.js
+++ b/packages/inertia-vue/src/app.js
@@ -19,6 +19,10 @@ export default {
       type: Function,
       default: props => props,
     },
+    injectProps: {
+      type: Boolean,
+      default: () => false,
+    }
   },
   data() {
     return {
@@ -41,6 +45,10 @@ export default {
   },
   render(h) {
     if (this.component) {
+
+      if (this.injectProps)
+        this.component.props = Object.keys(this.props);
+
       const child = h(this.component, {
         key: this.key,
         props: this.props,


reinink:
Hey @KABBOUCHI, thanks for the contribution!

So, my gut on this is "no", simply because this doesn't feel like the Vue.js way of doing things. I think it's better to explicitly set props, and I think Vue developers are used to doing it that way.

Do you have any precedent from elsewhere in the Vue community indicating that this is a valid approach?

KABBOUCHI:
Pages are controlled by Inertia, pages are not reusable components, for me it's safe to use and it's optional, it's not forced by default.

reinink:
Yeah, I know...and for that reason it's tempting to merge this in. However, my gut is telling me "no". Vue already has an established way of registering props, and I want to be careful not to introduce new "magical" ways of doing something that is already a "solved thing" in Vue.

In addition to that, this top-level option won't apply to the other adapters (React and Svelte). Not that all adapters need feature parity, but it is another consideration.

I'm going to close this for now, but am (obviously) willing to reconsider in the future if we get this feature request more often.

Thanks again @KABBOUCHI! 🤟

KABBOUCHI:
> In addition to that, this top-level option won't apply to the other adapters (React and Svelte). Not that all adapters need feature parity, but it is another consideration.

but react inject all the props? they already available.

reinink:
> but react inject all the props? they already available.

Correct, which is why the `injectProps` feature wouldn't apply there.

PR #216, opened 2020-09-14
Replace Axios with native XMLHttpRequest

claudiodekker:
This PR serves as a basis / starting point for eventually migrating away from Axios. 
The basic functionality presented works, but file uploads have not been tested.

This PR will remain a WIP as to not accidentally get pulled in to core.

### DIFF
--- a/packages/inertia/package.json
+++ b/packages/inertia/package.json
@@ -19,7 +19,6 @@
     "watch": "microbundle watch --format cjs"
   },
   "dependencies": {
-    "axios": "^0.19.0",
     "nprogress": "^0.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


clementmas:
I think removing the dependency on `axios` is a good move. However why not use `fetch` instead of `XMLHttpRequest`?

If browser compatibility is the issue, then you can't use `URLSearchParams` either.

citricguy:
Would there still be some sort of API available to mimic axios interceptor behavior after this change? 

For example, could this still be accomplished without any odd hacks: 

```
axios.interceptors.response.use(response => response, async err => {
  const status = get(err, 'response.status')

  if (status === 419) {
    // Refresh our session token
    await axios.get('/csrf-token')

    // Return a new request using the original request's configuration
    return axios(err.response.config)
  }

  return Promise.reject(err)
})
```

https://dev.to/grantholle/better-csrf-refreshing-in-laravel-and-axios-177c

reinink:

PR #218, opened 2020-09-15
accept new returnsJson option and return the axios response back

kingeke:
This is not a major feature, I think its more of a nice to have.

Added a new option parameter called returnsJson, if passed as true, it'll return the axios response object back to the initiator which then can be used to perform other actions.

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -68,7 +68,7 @@ export default {
     return this.visitId
   },
 
-  visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false, only = [], headers = {}} = {}) {
+  visit(url, { method = 'get', data = {}, replace = false, preserveScroll = false, preserveState = false, only = [], headers = {}, returnsJson = false} = {}) {
     progress.start()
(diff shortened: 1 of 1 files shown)


claudiodekker:
Hi @kingeke, thanks for contributing!

Unfortunately, however, we _probably_ won't be accepting this one for a couple of reasons. Most importantly, the `$inertia.visit` method is really only meant for handling Inertia requests, and extending it's behaviour into becoming capable of also handling API/custom JSON calls just doesn't feel right. 

Another reason is that if this were to be accepted, it would most likely eventually cause a breaking change due to our [plans to drop Axios from Inertia eventually](https://github.com/inertiajs/inertia/pull/141#issuecomment-688962312).

Instead, my recommendation would be to manually create an Axios request in your codebase instead, and to use that to perform the other actions you want to do. 👍

kingeke:
Hello @claudiodekker I understand. Closing pull request.

reinink:
Well said @claudiodekker. 👍 

@kingeke It's totally common to create some manual XHR requests within an Inertia.js app, but in those situations we recommend just using Axios, XHR, or even fetch directly.

PR #222, opened 2020-09-18
Check for sessionStorage before getItem from it

niveshsaharan:
sessionStorage.getItem() throws error when the app is loaded inside an iFrame

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -18,7 +18,7 @@ export default {
 
     if (window.history.state && this.navigationType() === 'back_forward') {
       this.setPage(window.history.state)
-    } else if (window.sessionStorage.getItem('inertia.hardVisit')) {
+    } else if (window.sessionStorage && window.sessionStorage.getItem('inertia.hardVisit')) {
       window.sessionStorage.removeItem('inertia.hardVisit')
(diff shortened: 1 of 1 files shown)


ninjaparade:
@reinink I tested this out, you can see the result below. This is in [ungoogled chrome](https://ungoogled-software.github.io/ungoogled-chromium-binaries)

image

I want to ensure the problem with these browsers is they're blocking and throwing exceptions for even "trying" to access `window.sessionStorage`. 

with my changes from #234 you can see it loads without any errors.

image

reinink:
Okay, so just calling `window.sessionStorage` causes this error. Understood.

Does this only happen in Ungoogled Chrome? Or does this happen elsewhere as well (ie. Safari)?

niveshsaharan:
@reinink I am using Inertia with simple window.sessionStorage and it is working in Chrome, Brave, Firefox, Safari, Opera. I haven't tested it in Ungoogled Chrome yet.

niveshsaharan:
Yes, the issue is with Ungoogle Chrome for sure.

niveshsaharan:
How about adding something like:

`window.self === window.top`

or 

`window === window.top`

reinink:
This has been fixed in #257. 👍

Issue #45, opened 2019-05-28
Preloading visits


sebastiandedeyne:
It would be great if we would be able to preload visits before the user clicks a link. This would be opt-in behaviour per link.

Next.js does this pretty good:

```jsx
<Link prefetch href="/">
  <a>Home</a>
</Link>
```

https://github.com/zeit/next.js#prefetching-pages

With the way Inertia works this could be very difficult to implement. I currently see two things we could preload:

1) Preload the entire page

We could build something like [InstantClick](http://instantclick.io/), that prefetches the next page in the background when the user hovers over a link, or when we add a `prefetch` attribute to the link.

The problem here is that we'd be prefetching the html, but not the JavaScript chunk needed for the next page.

2) Preload the JavaScript chunk that's needed for the page behind the link

Ideally we would be able to preload the script for the next page's component. However, Inertia has no knowledge which URL's map to which chunks, so I have no idea how this could work.

Any input on this matter is appreciated!

ecreeth:
@sebastiandedeyne see [Google quicklink](https://github.com/GoogleChromeLabs/quicklink)

sebastiandedeyne:
I like the idea of prefetching page states when an Inertia link with a `prefetch` attribute is in the viewport. On click, the page state would be applied, instead of doing a full visit.

What do you think @reinink? I could whip up a proof of concept.

ecreeth:
> We could build something like [InstantClick](http://instantclick.io/), that prefetches the next page in the background when the user hovers over a link, or when we add a `prefetch` attribute to the link.
 The problem here is that we'd be prefetching the html, but not the JavaScript chunk needed for the next page.

@sebastiandedeyne see [simple-pjax](https://mitranim.com/simple-pjax/react/)

PR #223, opened 2020-09-18
Example: Link preloading (GET-only)

claudiodekker:
This functionality adds a 'request manager' of sorts, which has the following behavior:

1) If there is no request, it starts the request and returns a promise that will eventually resolve (or fail)
2) If the request is pending, it hooks a promise-resolver into the existing request, which will resolve (or fail) once completed.
3) If the request has successfully finished, it returns a promise that instantly resolves to the (cached) request.
4) If the prefetch request failed, it'll start a new request (see 1)

In all situations, in addition to the promise, you'll receive a cancel hook, so that if you want to cancel the individual visit.

This request manager has been implemented to pre-load GET requests on-hover (see https://instant.page/ for what it's supposed to do), as well as to cancel all 'pending' requests once the page switches.

Solves #45


ArondeParon:
Is this still being considered? Sounds like a great improvement!

reinink:

TheBlckbird:
@claudiodekker I think you could update this PR so it works with the current version of Inertia and re-submit it.

TheBlckbird:
although I would use mousedown and not hover

aarondfrancis:
Any chance we can pick this one back up? I really like Livewire's prefetching, it'd be nice to have it here too

reinink:
> Any chance we can pick this one back up? I really like Livewire's prefetching, it'd be nice to have it here too

It's something we're looking at for v2 👍

PR #224, opened 2020-09-18
Catch Inertia (error) Response early

claudiodekker:


reinink:
I like what you're getting at here @claudiodekker, and is some ways I like it better (separating the two different error types), but in some ways I don't (having an extra `catch` statement, and putting the `catch` before the `then` feels odd). Since I can't decide which approach I like better, I'm going to leave it as is. Thanks anyway! 🤟

PR #228, opened 2020-09-22
Events: Re-key 'completed' to 'percentage'

claudiodekker:


packages/inertia/src/inertia.js
```diff
@@ -146,7 +146,7 @@ export default {
           ...(this.page.version ? { 'X-Inertia-Version': this.page.version } : {}),
         },
         onUploadProgress: progress => {
-          progress.completed = Math.round(progress.loaded / progress.total * 100)
+          progress.percentage = Math.round(progress.loaded / progress.total * 100)
```
claudiodekker, line 149: Can also add 'completed' as a boolean if we want:

```suggestion
          progress.percentage = Math.round(progress.loaded / progress.total * 100)
          progress.completed = progress.percentage === 100
```

PR #230, opened 2020-09-24
[Svelte] Add auto update method to use:inertia

pedroborges:
Add update method to the use directive. This ensures options are updated when the directive parameters are dynamic.


reinink:
I don't know what this means, but trust you know what you're doing here @pedroborges. 😂

pedroborges:
Yeah, it makes sure `options` inside the directive action can be updated in case it's dynamic and not a static object.

PR #234, opened 2020-09-24
Add window.sessionStorage check

ninjaparade:
add try catch to init to make sure `window.sessionStorage` is accessible

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -12,15 +12,19 @@ export default {
   init({ initialPage, resolveComponent, updatePage }) {
     this.resolveComponent = resolveComponent
     this.updatePage = updatePage
-
-    if (window.history.state && this.navigationType() === 'back_forward') {
-      this.setPage(window.history.state)
-    } else if (window.sessionStorage.getItem('inertia.hardVisit')) {
-      window.sessionStorage.removeItem('inertia.hardVisit')
-      this.setPage(initialPage, { preserveState: true })
-    } else {
-      initialPage.url += window.location.hash
-      this.setPage(initialPage)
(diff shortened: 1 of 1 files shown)


claudiodekker (changes requested):
I'm not sure whether we _want to_ support situations where the session storage isn't available, but if @reinink decides to pull this in, I'd like to point out that we would still have to think regarding what to do with `hardVisit`, as it still accesses the sessionStorage directly:

https://github.com/inertiajs/inertia/blob/0e43679e58e16f798e551acf3b694fe399134108/packages/inertia/src/inertia.js#L193-L195

ninjaparade:
The problem I an trying to solve is that, in an iframe of certain browsers, the call to `window.sessionStorage` is not allowed. With the changes I've suggested above, I haven't encountered any other issues.

reinink:
@ninjaparade I am trying to reproduce this issue, but I can't seem to. I've tried in Chrome, Safari and Firefox, and it seems to work in all. What are the conditions needed to cause this?

ninjaparade:
You should see the error if you embed an inertia app in a iframe in Safari, Firefox, ungoogled chrome or Brave. All of them have very strict protection from frames accessing window and cookies.

reinink:
Hey Yaz, would you mind testing out this solution? #222

I've been unable to reproduce this issue, and kind of like this solution, since it avoids adding a try/catch (which could be thrown for other things, in which we don't actually want to catch it.

Thanks!

ninjaparade:
@reinink I'll test it out today and let you know, thanks for the ping 👍

reinink:
This has been fixed in #257. 👍

PR #235, opened 2020-09-25
Inertia-link: Local event support

claudiodekker:
```vue
<InertiaLink :href="`/posts/${post.id}`" method="delete" @start="() => confirm('Are you sure you delete this post?')">
  Delete post
</InertiaLink>
```

### DIFF
--- a/packages/inertia-vue/src/link.js
+++ b/packages/inertia-vue/src/link.js
@@ -34,9 +34,20 @@ export default {
     headers: {
       type: Object,
       default: () => ({}),
-    }
+    },
   },
   render(h, { props, data, children }) {
+    data.on = {
+      click: () => ({}),
+      cancelToken: () => ({}),
+      start: () => ({}),
+      progress: () => ({}),
+      finish: () => ({}),
+      cancel: () => ({}),
+      success: () => ({}),
+      ... (data.on || {}),
+    }
+
     return h('a', {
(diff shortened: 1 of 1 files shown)


reinink:
Awesome, thanks @claudiodekker. I'd love to get this added to the React and Svelte adapters at the same time. Do you feel comfortable making that change as well? Maybe @sebastiandedeyne and @pedroborges can provide direction (or PR it themselves).

claudiodekker:
I gave it a shot, but React & Svelte do things quite differently, and since I really don't know either I figured I'd just leave it be for the others. Sorry!

pedroborges:
I started looking into it. I report back when I find a straight-forward way to implement it in the Svelte adapter.

pedroborges:
What does this PR change in userland?

`<inertia-link data="{ on: click: () => { /* ... */ } }" href="/">Home</inertia-link>`

or

`<inertia-link href="/" @click="() => { /* ... */ }">Home</inertia-link>`

claudiodekker:
There's no change in the userland part, really (so the 2nd example).

Rather, it's just that the new listeners get 'forwarded' to the underlying `Inertia.visit` call if provided (otherwise a noop method gets passed)

pedroborges:
You answered my question. I used `@click`, should have used one of the new events...

PR #239, opened 2020-09-30
Add get() method

reinink:
We currently have methods for `post()`, `put()`, `patch()` and `delete()`, but we don't have one for `get()`. Right now to do a GET visit you simply use `Inertia.visit()`. However, since people might expect a `get()` method to exist, this PR adds it. 👍 

```js
Inertia.get(url, data, options)
```

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -244,6 +244,10 @@ export default {
     }
   },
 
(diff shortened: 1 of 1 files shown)


claudiodekker (approved):

stevebauman:
Just ran into this myself yesterday, thanks for putting this in! 👍

PR #240, opened 2020-09-30
[React] InertiaLink local event props

sebastiandedeyne:
The React version of #235

### DIFF
--- a/packages/inertia-react/index.d.ts
+++ b/packages/inertia-react/index.d.ts
@@ -17,7 +17,7 @@ type App<
   transformProps?: (props: PagePropsBeforeTransform) => PageProps
 }>
 
-interface InertiaLinkProps extends React.HTMLAttributes<HTMLAnchorElement> {
+interface InertiaLinkProps {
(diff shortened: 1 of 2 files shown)


packages/inertia-react/index.d.ts
```diff
@@ -30,9 +30,17 @@ interface InertiaLinkProps extends React.HTMLAttributes<HTMLAnchorElement> {
   preserveState?: boolean | ((props: Inertia.PageProps) => boolean)
   replace?: boolean
   only?: string[]
+  onCancelToken: (cancelToken: string) => void
```
claudiodekker, line 33: Are you sure cancelToken is a string? 
If memory serves me correctly, it's an object on which you can call `.cancel()`

sebastiandedeyne: Good catch! Had it confused with something else. I think it's a `CancelTokenSource` object.

claudiodekker (approved):
Looks good to me visually :)

PR #243, opened 2020-10-02
Updates visit options types and adds get() method type definition

dulldave:
This PR updates the "manual visits" (https://inertiajs.com/manual-visits) options and method type definitions to match what's currently deployed.

### DIFF
--- a/packages/inertia/index.d.ts
+++ b/packages/inertia/index.d.ts
@@ -1,3 +1,5 @@
+import { CancelTokenSource } from "axios"
+
 declare global {
   // These open interfaces may be extended in an application-specific manner via
   // declaration merging / interface augmentation.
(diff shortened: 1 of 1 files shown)


packages/inertia/index.d.ts
```diff
@@ -20,10 +22,17 @@ export interface Page<CustomPageProps extends PageProps = PageProps> {
...
   method?: string
+  replace?: boolean
   preserveScroll?: boolean | ((props: Inertia.PageProps) => boolean)
   preserveState?: boolean | ((props: Inertia.PageProps) => boolean)
-  replace?: boolean
   only?: string[]
+  headers?: object
+  onCancelToken?: (cancelToken: CancelTokenSource) => void
+  onStart?: (visit: VisitOptions & {url: string}) => void | boolean
+  onProgress?: (progress: any) => void // Cast to any, Axios doesn't have a type definition for this
```
reinink, line 32: I'm pretty certain this will always be a native [ProgressEvent](https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent). Does that help at all?

reinink:
Thanks for this! 🙌

dulldave:
@reinink Sorry, this is my bad, I missed your reply. Thanks for getting this pushed over! 🎉

Issue #85, opened 2019-10-09
Preserve #id_of_element in route after redirection


smbdelse:
**Problem:**
When inertia-link href attribute contains #some_id it is not preserved.

**Example:**
After click inertia-link with href="homepage#contact" it send us to //example.com/homepage, dropping "#contact" from url.

**Expected behavior:**
To keep #id in url and scroll to it's position after redirect

smbdelse:
For now I am marking proper behavior by getting hash before redirection, transform it to standard GET parameter, send a request with it and attach it to $page object with response, so I can scroll to desired location then. But I think that it should be somehow handled by Inertia itself.

reinink:
Yeah, totally, this is something I need to figure out still. 👍

hivokas:
My proposal is to send `X-Inertia-Hash` header alongside with other headers.
And append the value of this header at the end of `url` which is sent to the front end on each Inertia response.

hivokas:
My another proposal would be to get the hash of the target URL before visiting it. 
And execute `location.href = hash` after the axios request.

hivokas:
@reinink your thoughts?

anatoliyarkhipov:
@hivokas 

> My another proposal would be to get the hash of the target URL before visiting it.
> And execute location.href = hash after the axios request.

It's not so simple: if you manipulate the hash bypassing Inertia (i.e. through `location.href = "#hash"` like in your proposal, or by click on a `<a href="#hash" />`), and then navigate somewhere through `Inertia.visit` (or `InertiaLink`, which uses `Inertia.visit` under the hood), the Back button will not work properly - it will change URL in the address bar back to `#hash`, but will not get you back to the page from which you `visit`ed the current one.

reinink:
This has been fixed in #257, which just got merged in, and will be released soon. 🙌

ldanielduarte:
Hello guys,

I'm sorry to write this here because it is not an issue or a bug report, just some support request - but it seems that the topic fits very well in the subject and I am sure that @reinink will be able to help me 😄 

When redirecting from the controller:
```
class EntityController extends Controller
{
    public function store(Request $request): RedirectResponse
    {
        ....
        return redirect()->route('entities.list')->withFragment('manage');
    }
}
```

How can I keep the hash in the URL?

For instance, in this case, return to the page `https://domain.com/entities#manage`

[...]

PR #244, opened 2020-10-03
Preserve url hash on navigation

rgehan:
This PR is an attempt at fixing https://github.com/inertiajs/inertia/issues/85.

It goes with its inertiajs/inertia-laravel counterpart: https://github.com/inertiajs/inertia-laravel/pull/155

It adds a new `X-Inertia-Hash` header which contains the hash part of the URL.

That header can later be re-appended to the URL by the API, so that the client-side URL is correct.

---

If you are happy with the proposed changes, here are a few things that will need to be done too:
- [ ] Support the same behavior in inertiajs/inertia-rails
- [ ] Update PingCRM to demonstrate this behavior

---

Feel free to suggest any improvement :)

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -126,10 +126,13 @@ export default {
     let visitId = this.createVisitId()
     onCancelToken(this.cancelToken)
 
+    const stringifiedUrl = url.toString()
+    const [, hash] = stringifiedUrl.split('#')
+
     return new Proxy(
       Axios({
         method,
-        url: url.toString(),
+        url: stringifiedUrl,
         data: method.toLowerCase() === 'get' ? {} : data,
         params: method.toLowerCase() === 'get' ? data : {},
         cancelToken: this.cancelToken.token,
@@ -143,6 +146,7 @@ export default {
             'X-Inertia-Partial-Data': only.join(','),
           } : {}),
           ...(this.page.version ? { 'X-Inertia-Version': this.page.version } : {}),
+          ...(hash ? { 'X-Inertia-Hash': '#' + hash } : {}),
         },
         onUploadProgress: progress => {
           progress.percentage = Math.round(progress.loaded / progress.total * 100)


reinink:
Cool, thanks for attempting to fix this @rgehan! This looks like a fairly straightforward solution. Let me do some testing with this next week and see if I can find any issues with it.

Do you know of any gotchas with this solution?

rgehan:
I don't know about any particular gotchas no, but I could totally be missing something (that's where I'd like your opinion/experience).

That said, considering how small the changes are, I don't think it has a high potential of breaking things.

reinink:
Good stuff @rgehan.

I've been messing around with this tonight, and while this did preserve the hash, the browser wasn't actually following the anchor link. I'm still not totally sure if this is just something I am doing wrong. In your testing, did the scroll position actually change to the location of the anchor link element (`id="whatever"`)?

To get it working, I had to  update the Inertia `resetScrollPositions()` method to do this manually:

```js
if (window.location.hash) {
  const el = document.getElementById(window.location.hash.slice(1))
  if (el) {
    el.scrollIntoView()
  }
}
```

And now it works great.

I'm going to test this a little more. 👍

reinink:
@rgehan Also, regarding this:

> We'd have to check whether the URL matches what was requested, and isn't a redirection though.

Maybe I'm missing something, but this already an issue with your solution. For example, if I request `/users#something`, but my session has expired, this implementation causes me to land on `/login#something`.

I think we might be better just doing this client-side in Inertia core:

- Before the request starts, save the hash.
- When the requests finishes, check to see if the intended URL equals the response URL.
    - If yes, add the hash.
    - If no, ignore the hash.

reinink:
I just put together an alternative approach to this, see #257.

reinink:
Closing this in favour of #257 (which is almost done).

Thanks so much for bringing this to my attention, and pushing me to fix this @rgehan!

PR #245, opened 2020-10-05
Add Vue 3 adapter

reinink:
This adds a new Vue 3 adapter package (`@inertiajs/inertia-vue3`) to Inertia.js. 🎉

I considered trying to make this work within the existing `@inertiajs/inertia-vue` package, but decided it was better to offer this as a separate package, since quite a lot has changed in Vue 3. Further, I suspect that we're going to be supporting the existing Vue (2) adapter for a long time, so having two packages is just easier to work with.

## 🚧 How to use

To use this package, you'll need to use a Vue 3 compatible bunder. I am using the new version of Laravel Mix (6) that @jeffreyway has been busy building recently (`"laravel-mix": "6.0.0-beta.3"`).

The app setup has changed a little in Vue 3, however it's still very familiar. Here is what your `app.js` file looks like now:

```js
import { app, plugin } from '@inertiajs/inertia-vue3'

const el = document.getElementById('app')

createApp({
  render: () => h(app, {
    resolveComponent: name => import(`@/Pages/${name}`).then(module => module.default),
  })
}).use(plugin).mount(el)
```

## ⚠️  Vue plugin changes

This PR moves the Vue plugin to a separate import, instead of packaging this in with `app`.

```js
import { app, plugin } from '@inertiajs/inertia-vue3'

app.use(plugin)
```

I have made the same change in the `@inertiajs/inertia-vue` package, but in a [non-breaking way](https://github.com/inertiajs/inertia/blob/7ff99b959e96ce8731ecb8ba9d749dce692f6f20/packages/inertia-vue/src/app.js#L64-L66).

I think this is an important change, since folks don't realize that the Inertia Vue plugin is actually totally optional.

## 🚨 Global page access (breaking change)

This PR also changes the global `page` property to be the full [page object](https://inertiajs.com/the-protocol#the-page-object), instead of just the props (`page.props`). This is really handy, since it gives you access to `page.url` and `page.version`, which are both reactive, useful for things like setting active states on links. This is a breaking change I have made to all the existing adapters as well:

- `@inertiajs/inertia-vue`: Update the `$page` property to return the full `page` object.
- `@inertiajs/inertia-react`: Update the `usePage()` hook to return the full `page` object.
-  `@inertiajs/inertia-svelte`: Update the `$page` store to return the full `page` object.


packages/inertia-vue/src/plugin.js
```diff
@@ -0,0 +1,13 @@
+import link from './link'
+import remember from './remember'
+import { Inertia } from '@inertiajs/inertia'
+
+export default {
+  instance: null,
+  install(Vue) {
+    Object.defineProperty(Vue.prototype, '$inertia', { get: () => Inertia })
```
JakubRobotka, line 8: According to documentation, you can use globalProperties instead of prototype.
https://v3.vuejs.org/api/application-config.html#globalproperties

packages/inertia-vue3/src/link.js
```diff
@@ -0,0 +1,67 @@
+import { h } from 'vue'
+import { Inertia, shouldIntercept } from '@inertiajs/inertia'
+
+export default {
+  functional: true,
```
JakubRobotka, line 5: Functional attribute is removed from Vue3
https://v3.vuejs.org/guide/migration/functional-components.html#functional-components

packages/inertia-vue/src/plugin.js
reinink: This is actually the Vue 2 plugin (I made some changes to the other adapters while making the Vue 3 plugin).

packages/inertia-vue3/src/link.js
reinink: Good catch, this has been removed! 👍

nathandaly (approved):

ExpDev07:
Really nice work! Any idea when it’ll be released 🥳🥳🥳?

nathandaly:
:shipit:

packages/inertia-react/src/App.js
```diff
@@ -53,8 +54,8 @@ export default function App({
 
   return createElement(
     PageContext.Provider,
-    { value: page.props },
-    renderChildren({ Component: page.component, key: page.key, props: page.props })
+    { value: current.page },
+    renderChildren({ Component: current.component, key: current.key, props: current.page.props }),
   )
 }
```
reinink, line 61: @sebastiandedeyne Ping! Just wanted you to be aware that I made these changes to the React adapter. 👍

sebastiandedeyne: LGTM! `swapComponent`'s argument list looks a lot cleaner than `updatePage`!

reinink: Awesome! ❤️

nathandaly:
🎉 Exciting stuff, thanks @reinink 😃

ExpDev07:
Hello, why can i not ```npm install @inertiajs/inertia-vue3`` :O

reinink:
@ExpDev07 It's only been merged. I still need to do the release, which takes some time. It should be done within the next week.

PR #248, opened 2020-10-06
Updated $page to access $page directly insted of $page.props

FaizAhmadSE:


reinink:
Hello, thanks for the contribution, but this change has already been made in #245, and not only for the Vue adapter, but for the React and Svelte adapters too. That PR should be merged and tagged within the week. 👍

aske-cph:
I think the examples here should be updated then:
https://inertiajs.com/shared-data

reinink:
@aske-cph It has been...where are you seeing the old behaviour in the docs?

To be clear, the new behaviour is `$page.props`, which might not be very clear from this issue title.

aske-cph:
Ah i thought the header set the complete opposite - reason i am confused is that i don't get a "props" property in 0.3.3 for some reason - if i dump the $page obj flash for example just lies next to Auth, Errors, Jetstream etc..

PR #254, opened 2020-10-13
Add typings for Inertia.on

AdrianMrn:


reinink:
Nice, thanks @AdrianMrn!

PR #257, opened 2020-10-14
Preserve url hash on navigation

reinink:
This PR adds support for URL hashes in Inertia.js (which are currently very much broken).

## URL hashes when making Inertia visits

Currently it's not possible to make an Inertia visit that includes a hash in the URL. For example:

```html
<inertia-link url="/users#create">Create user</inertia-link>
```

This is because hashes are never sent from the browser to the server, which means when an Inertia response is received (which includes the `url` as part of the `page` object) and Inertia updates the history state with the new URL, the hash is gone.

We've been able to fix this by keeping track of the hash client-side, and appending it back on the URL when the Inertia response is received. Here's how:

1. When a request is made, we parse out the hash from the URL before making the request to the server.
2. When the response is received, we compare the intended URL with the response URL. If they match (meaning we ended up at the intended URL, and didn't get logged out or something), then we append the hash back to the URL (which causes it to be included in the URL when the history state is updated).
3. When we reset the scroll position, we check if a hash is set, and manually scroll to that location using the `Element.scrollIntoView()` method.

The nice thing about this change is that it's entirely made in Inertia core, and requires no adapter changes.

This is an alternative solution to #244.

## Same-page anchor links

Currently same-page anchor links are broken in Inertia. For example:

```html
<a url="#create">Create user</a>
<!-- ... -->
<h2 id="create">Create User</h2>
```

This is because anchor links add a new entry to the history state, however our current `popstate` handler doesn't account for them—rather it assumes that there will always be an `event.state` property present. This causes an error to be thrown when navigating history "anchor visits".

To solve this, I've updated the Inertia `popstate` handler to detect same-page anchor visits. When these occur, Inertia adds the current `page` from local memory to the history state. This allows anchor visits to have their own remember state (local state caching) and scroll positioning.

## Other changes

Here is a list of other changes that came as a result of the refactor to support hashes:

- Fixed scroll restoration for offsite back/forward visits.
- Fixed `sessionStorage` issues for browsers that have it disabled.
- Renamed "hard visits" to "location visits".
- Updated location visits to respect the `preserveScroll` option.
- Made the URL check when determining the history "replace" value less brittle.
- Removed prop transforming from `popstate` handler (to prevent double transforming).
- Split up large methods into smaller, better named methods.
- Improved overall code formatting.
- Fixed bug where same page location visits wouldn't cause a full page reload.

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -126,8 +126,7 @@ export default {
   },
 
   urlWithoutHash(url) {
-    const parts = url.href.split('#')
-    return parts[0]
+    return url.href.split('#')[0]
   },
 
   visit(url, {
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -64,6 +64,7 @@ export default {
       region.scrollTop = 0
       region.scrollLeft = 0
     })
+    this.saveScrollPositions()
 
     if (window.location.hash) {
       const el = document.getElementById(window.location.hash.slice(1))
@@ -265,7 +266,7 @@ export default {
   },
 
   restoreState(event) {
-    if (event.state) {
+    if (event.state && event.state.component) {
       const page = this.transformProps(event.state)
       let visitId = this.createVisitId()
       return Promise.resolve(this.resolveComponent(page.component)).then(component => {
@@ -276,6 +277,8 @@ export default {
           })
         }
       })
+    } else {
+      this.resetScrollPositions()
     }
   },
 
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -20,7 +20,7 @@ export default {
 
   handleInitialPageVisit(page) {
     if (this.isBackForwardVisit()) {
-      this.handleBackForwardVisit()
+      this.handleBackForwardVisit(page)
     } else if (this.isLocationVisit()) {
       this.handleLocationVisit(page)
     } else {
@@ -87,7 +87,8 @@ export default {
       && window.performance.getEntriesByType('navigation')[0].type === 'back_forward'
   },
 
-  handleBackForwardVisit() {
+  handleBackForwardVisit(page) {
+    window.history.state.version = page.version
     this.setPage(window.history.state, { preserveScroll: true }).then(() => {
       this.restoreScrollPositions()
     })
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -302,7 +302,7 @@ export default {
         }
       })
     } else {
-      let url = this.normalizeUrl(this.page.url)
+      const url = this.normalizeUrl(this.page.url)
       url.hash = window.location.hash
       this.replaceState({ ...this.page, url: url.href })
       this.resetScrollPositions()
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -115,6 +115,20 @@ export default {
     return this.visitId
   },
 
+  normalizeUrl(url) {
+    url = url.toString()
+    try {
+      return new URL(url)
+    } catch (error) {
+      return new URL(`${window.location.origin}${url}`)
+    }
+  },
+
+  urlWithoutHash(url) {
+    const parts = url.href.split('#')
+    return parts[0]
+  },
+
   visit(url, {
     method = 'get',
     data = {},
@@ -130,6 +144,7 @@ export default {
     onCancel = () => ({}),
     onSuccess = () => ({}),
   } = {}) {
+    url = this.normalizeUrl(url)
     let visit = { url, ...arguments[1] }
     if (onStart(visit) === false || !this.fireEvent('start', { cancelable: true, detail: { visit } } )) {
       return
@@ -139,12 +154,10 @@ export default {
     let visitId = this.createVisitId()
     onCancelToken(this.cancelToken)
 
-    const [urlWithoutHash, hash] = url.toString().split('#')
-
     return new Proxy(
       Axios({
         method,
-        url: urlWithoutHash,
+        url: this.urlWithoutHash(url),
         data: method.toLowerCase() === 'get' ? {} : data,
         params: method.toLowerCase() === 'get' ? data : {},
         cancelToken: this.cancelToken.token,
@@ -171,8 +184,8 @@ export default {
         if (only.length && response.data.component === this.page.component) {
           response.data.props = { ...this.page.props, ...response.data.props }
         }
-        if (hash && urlWithoutHash === response.data.url) {
-          response.data.url = `${response.data.url}#${hash}`
+        if (url.hash && this.urlWithoutHash(url) === this.normalizeUrl(response.data.url).href) {
+          response.data.url = `${response.data.url}${url.hash}`
         }
         return this.setPage(response.data, { visitId, replace, preserveScroll, preserveState })
       }).then(() => {
@@ -184,14 +197,11 @@ export default {
         } else if (Axios.isCancel(error)) {
           onCancel()
         } else if (this.isHardVisit(error.response)) {
-          let location = new URL(error.response.headers['x-inertia-location'])
-          if (hash &&
-              window.location.origin === location.origin &&
-              urlWithoutHash === `${location.pathname}${location.search}`
-          ) {
-            location.hash = `#${hash}`
+          let locationUrl = this.normalizeUrl(error.response.headers['x-inertia-location'])
+          if (url.hash && this.urlWithoutHash(url) === this.urlWithoutHash(locationUrl)) {
+            locationUrl.hash = url.hash
           }
-          this.hardVisit(location.href)
+          this.hardVisit(locationUrl.href)
         } else if (error.response) {
           if (this.fireEvent('invalid', { cancelable: true, detail: { response: error.response } })) {
             modal.show(error.response.data)
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -145,7 +145,6 @@ export default {
   },
 
   normalizeUrl(url) {
-    url = url.toString()
     try {
       return new URL(url)
     } catch (error) {
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -128,7 +128,7 @@ export default {
   },
 
   isInertiaResponse(response) {
-    return response && response.headers['x-inertia']
+    return response.?headers['x-inertia']
   },
 
   cancelActiveVisits() {
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -31,7 +31,7 @@ export default {
 
     this.fireEvent('navigate', { detail: { page: initialPage } })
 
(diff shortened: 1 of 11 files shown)


packages/inertia/src/inertia.js
```diff
@@ -63,6 +64,13 @@ export default {
       region.scrollTop = 0
       region.scrollLeft = 0
     })
+
+    if (window.location.hash) {
+      const el = document.getElementById(window.location.hash.slice(1))
+      if (el) {
+        el.scrollIntoView()
+      }
```
innocenzi, line 72: You should be able to remove the condition thanks to the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) :+1: 

```suggestion
      document.getElementById(window.location.hash.slice(1))?.scrollIntoView()
```

reinink: Are you sure this works on method calls? I'm getting errors when I try.

innocenzi: Hmm, yes, definitely works. I've used it in the past. You can try on this page too: 

```js
// Will scroll you to the comment box
document.getElementById('issue-comment-box')?.scrollIntoView()


// Won't scroll and won't throw
document.getElementById('dummy-id-blah')?.scrollIntoView()
```

What error are you getting? If it's trouble to debug, this is not important to change anyway. Just a nice syntactic sugar tip.

reinink: So I think it's just eslint complaining. It does compile. 👍

reinink: This is going to require an eslint upgrade, which I'll do later. Going to leave it like this for now.

packages/inertia/src/inertia.js
```diff
@@ -107,6 +116,20 @@ export default {
...
+    try {
+      return new URL(url)
+    } catch (error) {
+      return new URL(`${window.location.origin}${url}`)
+    }
+  },
+
+  urlWithoutHash(url) {
+    const parts = url.href.split('#')
+    return parts[0]
```
claudiodekker, line 130:
```suggestion
    return url.href.split('#')[0]
```

claudiodekker, line 130:
```suggestion
    return url.href.split('#')[0]
```
```

File: packages/inertia/src/inertia.js (hunk 1 above)
Comment (sebastiandedeyne) [reply packages/inertia/src/inertia.js:73]: Or refactor to TypeScript 😏

**Review by sebastiandedeyne — APPROVED** [review]

File: packages/inertia/src/inertia.js [hunk 3]
```diff
@@ -170,8 +226,12 @@ export default {
...
         } else if (Axios.isCancel(error)) {
           onCancel()
-        } else if (this.isHardVisit(error.response)) {
-          this.hardVisit(error.response.headers['x-inertia-location'])
+        } else if (this.isLocationVisitResponse(error.response)) {
+          let locationUrl = this.normalizeUrl(error.response.headers['x-inertia-location'])
+          if (url.hash && this.urlWithoutHash(url) === this.urlWithoutHash(locationUrl)) {
+            locationUrl.hash = url.hash
+          }
+          this.locationVisit(locationUrl.href, preserveScroll)
```
claudiodekker, line 234: Is this even necessary? If we're not passing in the hash, I don't think we need to check it either?

claudiodekker, line 234:
```suggestion
          const locationUrl = this.normalizeUrl(error.response.headers['x-inertia-location'])
          this.locationVisit(locationUrl.href, preserveScroll)
```

packages/inertia/src/inertia.js
```diff
@@ -221,31 +278,33 @@ export default {
...
+    this.page = page
+    window.history.replaceState(page, '', page.url)
   },
 
-  restoreState(event) {
-    if (event.state) {
-      const page = this.transformProps(event.state)
+  handlePopstateEvent(event) {
+    if (event.state !== null) {
+      let page = event.state
```
claudiodekker, line 292:
```suggestion
      const page = event.state
```

packages/inertia/src/inertia.js
```diff
@@ -161,6 +214,9 @@ export default {
         if (only.length && response.data.component === this.page.component) {
           response.data.props = { ...this.page.props, ...response.data.props }
         }
+        if (url.hash && this.urlWithoutHash(url) === this.normalizeUrl(response.data.url).href) {
```
claudiodekker, line 217: Maybe we can extract this to a small helper method so that it's not as verbose? Maybe something like `this.isSamePageHashVisit`?

packages/inertia/src/inertia.js
```diff
@@ -221,31 +278,33 @@ export default {
...
+          this.page = page
           this.swapComponent({ component, page, preserveState: false }).then(() => {
-            this.restoreScrollPositions(page)
+            this.restoreScrollPositions()
             this.fireEvent('navigate', { detail: { page: page } })
           })
         }
       })
+    } else {
+      let url = this.normalizeUrl(this.page.url)
```
mathieutu, line 304: Same as previous comment, you have no re-attribution, so const is better here 
```suggestion
      const url = this.normalizeUrl(this.page.url)
```

packages/inertia/src/inertia.js
```diff
@@ -278,30 +337,32 @@ export default {
   },
 
   remember(data, key = 'default') {
-    let newState = { ...window.history.state }
-    newState.cache = newState.cache || {}
-    newState.cache[key] = data
-    this.replaceState(newState)
+    let page = { ...this.page }
```
mathieutu, line 340:
```suggestion
    const page = { ...this.page }
```

packages/inertia/src/inertia.js
```diff
@@ -278,30 +337,32 @@ export default {
...
+    page.rememberedState[key] = data
+    this.replaceState(page)
   },
 
   restore(key = 'default') {
-    if (window.history.state.cache && window.history.state.cache[key]) {
-      return window.history.state.cache[key]
+    if (window.history.state.rememberedState && window.history.state.rememberedState[key]) {
+      return window.history.state.rememberedState[key]
     }
```
mathieutu, line 348:
```suggestion
    return window.history.state?.rememberedState?.[key]
```

packages/inertia/src/inertia.js
```diff
@@ -278,30 +337,32 @@ export default {
   },
 
   remember(data, key = 'default') {
-    let newState = { ...window.history.state }
-    newState.cache = newState.cache || {}
-    newState.cache[key] = data
-    this.replaceState(newState)
+    let page = { ...this.page }
+    page.rememberedState[key] = data
+    this.replaceState(page)
```
mathieutu, line 342: Actually here if `rememberedState` is an object, you're mutating the original `this.page.rememberedState`, because your clone action is only on the first level.

You may need to do something like this:
```suggestion
    this.replaceState({
      ...this.page,
      rememberedState: {
    	...this.page.rememberedState,
    	[key]: data,
   	  }
    })
```

packages/inertia/src/inertia.js
```diff
@@ -221,31 +278,33 @@ export default {
...
-    if (event.state) {
-      const page = this.transformProps(event.state)
+  handlePopstateEvent(event) {
+    if (event.state !== null) {
+      let page = event.state
       let visitId = this.createVisitId()
       return Promise.resolve(this.resolveComponent(page.component)).then(component => {
         if (visitId === this.visitId) {
+          this.page = page
           this.swapComponent({ component, page, preserveState: false }).then(() => {
```
mathieutu, line 297: It could be relevant to use some async/await instead of nesting promises like that?

packages/inertia/src/inertia.js
```diff
@@ -63,42 +64,80 @@ export default {
...
+  isLocationVisitResponse(response) {
     return response && response.status === 409 && response.headers['x-inertia-location']
   },
 
-  fireEvent(name, options) {
-    return document.dispatchEvent(
-      new CustomEvent(`inertia:${name}`, options),
-    )
+  isInertiaResponse(response) {
+    return response && response.headers['x-inertia']
```
mathieutu, line 134:
```suggestion
    return response.?headers['x-inertia']
```

packages/inertia/src/inertia.js
```diff
@@ -63,42 +64,80 @@ export default {
...
+    }
   },
 
-  isHardVisit(response) {
+  handleLocationVisit(page) {
+    const locationVisit = JSON.parse(window.sessionStorage.getItem('inertiaLocationVisit'))
+    window.sessionStorage.removeItem('inertiaLocationVisit')
+    page.url += window.location.hash
+    page.rememberedState = window.history.state ? window.history.state.rememberedState : {}
+    page.scrollRegions = window.history.state ? window.history.state.scrollRegions : []
```
mathieutu, line 121: Made you some syntax suggestions which simplify code at several part. You take what you want. 🙂

```suggestion
    page.rememberedState = window.history.state?.rememberedState ?? {}
    page.scrollRegions = window.history.state?.scrollRegions ?? []
```

packages/inertia/src/inertia.js
reinink: The `hash` is included in the `href`. This is needed, as imagine you're making a request to `/users#create`, and a location visit (409) comes back. If we do a `window.location.href` visit there using the `X-Inertia-Location` header, need to add the hash back on, otherwise we'd only visit `/users`. Unless, of course, I'm missing something here.

packages/inertia/src/inertia.js
reinink: Hmm, I'm realizing that we actually need to check here if the response URL has a hash, otherwise we'll end up with multiple hashes. This is actually what we need:

```js
const responseUrl = this.normalizeUrl(response.data.url)
if (url.hash && !responseUrl.hash && this.urlWithoutHash(url) === responseUrl.href) {
  responseUrl.hash = url.hash
  response.data.url = responseUrl.href
}
```

packages/inertia/src/inertia.js
claudiodekker: Nevermind, you answered my question already: 

image

packages/inertia/src/inertia.js
reinink: This is awesome. 😍

packages/inertia/src/inertia.js
reinink: Pretty sure this should be `?.` 🤣

packages/inertia/src/inertia.js
reinink: Good catch! 👍

packages/inertia/src/inertia.js
reinink: If you wanted to submit a subsequent PR for this, I'd certainly consider it. I don't know what's all involved, so I'm going to keep it this way for this PR. 👍

packages/inertia/src/inertia.js
mathieutu: `?->` soon in php (8)! 😉

reinink: Honestly the feature I'm looking forward to the most in PHP 8! 🙌

dasundev:
Is there any option to `back()->withFragment('#users');` with Laravel?

PR #259, opened 2020-10-14
Update index.js

erokar:
The lowercase l in the Link import causes an import error

### DIFF
--- a/packages/inertia-svelte/src/index.js
+++ b/packages/inertia-svelte/src/index.js
@@ -1,5 +1,5 @@
 export { default as App, default as app, default as InertiaApp } from './App.svelte'
 export { default as Link, default as link, default as InertiaLink } from './Link.svelte'
-export { default as inertia } from './link'
+export { default as inertia } from './Link'
 export { default as page } from './page'
 export { default as remember } from './remember'


claudiodekker:
Hm, that's odd. From what I can tell _this change_ would be breaking, as the file itself _is_ lowercased, or am I missing something?

erokar:
You're right, my bad. I was mislead by this error from running `bin/webpack-dev-server` in a Rails project:

ERROR in ./node_modules/@inertiajs/inertia-svelte/src/index.js
Module not found: Error: [CaseSensitivePathsPlugin] `/Users/ero/code/rails/blog/node_modules/@inertiajs/inertia-svelte/src/link.svelte` does not match the corresponding path on disk `Link.svelte`.

It might be a bug in webpack-dev-server, but the proposed change will fix that error. Please disregard this pull request.

erokar:
I think the issue is with Rails' webpack setup, but I cannot get Inertia.js with Svelte to work with Rails without changing that line somehow. Changing the import to ".link.js" will fix it. The issue seems to be with Rails' webpack not being able to distinguish between the two Link/link imports.

Rails 6.0.3.4
MacOS Catalina 10.15.6

claudiodekker:
Yeah, that's annoying.. :/ I wish there was a way I knew how to help, but I'm entirely unfamiliar with the Rails world. 
Is there no way you can disable that `CaseSensitivePathsPlugin` in a webpack config somewhere?

erokar:
Thanks for your suggestions! I finally solved it by disabling the CaseSensitivePathsPlugin.

Leaving it here in case anyone has the same issue: In `config/webpack/environment.js` add this line before the last line: `environment.plugins.delete("CaseSensitivePaths")`

PR #262, opened 2020-10-16
Fix promise deprecation warning

### LINKED ISSUE #260: Promise deprecation warning triggers error in some situations
A Proxy is used to show a deprecation warning and return the original value:

https://github.com/inertiajs/inertia/blob/7f68b3e113571a9e1de150eab16db668a480173b/packages/inertia/src/inertia.js#L192

When you do a function call on the returned object, this works fine. But if you access a property, for example:

```js
this.$inertia.visit('...').foo
```

then you get the following error:

```
TypeError: Cannot read property 'bind' of undefined
```

I constantly have this error in my new app, where I use the Element framework and have things like this:

```html
<el-button @click="$inertia.visit('...')">Click Here</el-button>
```

I don't know what that framework does on the `click` event. I guess it's checking the returned result for stuff and tries to access properties. And these accesses trigger the error above...

Searched around on the topic, and it should be resolved when you change that line to:

```js
return Reflect.get(...arguments)
```

reinink:
Fixes #260

This PR updates the promise proxy used to show the deprecation warning for `then`, `catch` and `finally` calls. It also now only tries to bind the `target` if the property is a function.

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -256,8 +256,12 @@ export default {
         onFinish()
       }), {
         get: function(target, prop) {
-          console.warn('Inertia.js visit promises have been deprecated and will be removed in a future release. Please use the new visit event callbacks instead.\n\nLearn more at https://inertiajs.com/manual-visits#promise-deprecation')
-          return target[prop].bind(target)
+          if (['then', 'catch', 'finally'].includes(prop)) {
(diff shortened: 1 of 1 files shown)


klaasgeldof:
Looks good to me 👍

reinink:
Awesome, thanks @klaasgeldof. These has been merged and released: ` @inertiajs/inertia@v0.4.3`. 👍

klaasgeldof:
Perfect, thanks for the fix!

PR #263, opened 2020-10-16
Set link preserveState default based on method

### LINKED ISSUE #164: Set link preserveState default based on method
By default the post/put/patch methods preserve state by default ([see here](https://github.com/inertiajs/inertia/blob/master/src/inertia.js#L156-L174)). However, an Inertia link that uses post/put/patch will not do this by default, since the attribute default is false ([see here](https://github.com/inertiajs/inertia-vue/blob/master/src/link.js#L28)).

It makes me think that we should maybe check the link method, and if it's a post/put/patch request, automatically set it preserve state, like Inertia core does.

reinink:
Fixes #164

This PR changes the default `preserveState` value in the link components from `false` to `false` for `get` visits, and `false` for `post|put|patch|delete` visits.

- [x] Update the Vue 2 adapter
- [x] Update the Vue 3 adapter
- [x] Update the React adapter
- [x] Update the Svelte adapter

### DIFF
--- a/packages/inertia-react/src/Link.js
+++ b/packages/inertia-react/src/Link.js
@@ -8,12 +8,12 @@ export default forwardRef(function InertiaLink({
   data = {},
   href,
   method = 'get',
-  onClick = noop,
   preserveScroll = false,
-  preserveState = false,
+  preserveState = null,
   replace = false,
   only = [],
   headers = {},
+  onClick = noop,
   onCancelToken = noop,
   onStart = noop,
   onProgress = noop,
@@ -33,7 +33,7 @@ export default forwardRef(function InertiaLink({
           data,
           method,
           preserveScroll,
-          preserveState,
+          preserveState: preserveState ?? (method !== 'get'),
           replace,
           only,
           headers,
@@ -50,7 +50,6 @@ export default forwardRef(function InertiaLink({
       data,
       href,
       method,
-      onClick,
       preserveScroll,
       preserveState,
       replace,
@@ -63,7 +62,7 @@ export default forwardRef(function InertiaLink({
       onFinish,
       onCancel,
       onSuccess,
-    ]
+    ],
   )
 
   return createElement('a', { ...props, href, ref, onClick: visit }, children)
--- a/packages/inertia-react/src/Link.js
+++ b/packages/inertia-react/src/Link.js
@@ -33,7 +33,7 @@ export default forwardRef(function InertiaLink({
           data,
           method,
           preserveScroll,
-          preserveState: preserveState ?? (method !== 'get'),
+          preserveState: preserveState ?? (method.toLowerCase() !== 'get'),
           replace,
           only,
           headers,
--- a/packages/inertia-svelte/src/Link.svelte
+++ b/packages/inertia-svelte/src/Link.svelte
@@ -24,7 +24,7 @@
         data,
         method,
         preserveScroll,
-        preserveState: preserveState !== null ? preserveState : (method !== 'get'),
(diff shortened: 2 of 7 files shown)


reinink:
Pinging @sebastiandedeyne and @pedroborges. If you fine gents could help me make the same change in the React and Svelte adapters, I'd be super grateful.

reinink:
Update:

I've attempted to make the React and Svelte changes myself. I *think* I've done this correctly, but would very much appreciate a code review.

@pedroborges For some reason I was not able to use the nullish coalescing operator (`??`) in the Svelte adapter (like I did in the other adapters). I got a compile error. Any idea why that is?

packages/inertia-vue/src/link.js
```diff
@@ -67,7 +67,7 @@ export default {
               method: props.method,
               replace: props.replace,
               preserveScroll: props.preserveScroll,
-              preserveState: props.preserveState,
+              preserveState: props.preserveState ?? (props.method !== 'get'),
```
claudiodekker, line 70: Since this depends on user input, people could pass `GET` and experience unexpected behaviour

```suggestion
              preserveState: props.preserveState ?? (props.method.toLowerCase() !== 'get'),
```

packages/inertia-vue3/src/link.js
```diff
@@ -49,7 +49,7 @@ export default {
             method: props.method,
             replace: props.replace,
             preserveScroll: props.preserveScroll,
-            preserveState: props.preserveState,
+            preserveState: props.preserveState ?? (props.method !== 'get'),
```
claudiodekker, line 52:
```suggestion
            preserveState: props.preserveState ?? (props.method.toLowerCase() !== 'get'),
```

packages/inertia-svelte/src/Link.svelte
```diff
@@ -24,7 +24,7 @@
         data,
         method,
         preserveScroll,
-        preserveState,
+        preserveState: preserveState !== null ? preserveState : (method !== 'get'),
```
claudiodekker, line 27:
```suggestion
        preserveState: preserveState !== null ? preserveState : (method.toLowerCase() !== 'get'),
```

packages/inertia-react/src/Link.js
```diff
@@ -33,7 +33,7 @@ export default forwardRef(function InertiaLink({
           data,
           method,
           preserveScroll,
-          preserveState,
+          preserveState: preserveState ?? (method !== 'get'),
```
claudiodekker, line 36:
```suggestion
          preserveState: preserveState ?? (method.toLowerCase() !== 'get'),
```

sebastiandedeyne (changes requested):
No comments on the changes!

Typings need to be updated for this PR too, as `null` is now an allowed value for `preserveState`.

- https://github.com/inertiajs/inertia/blob/b308f0cdf5e2fa56cf28c7d412bb6f18210035e2/packages/inertia/index.d.ts#L27
- https://github.com/inertiajs/inertia/blob/b308f0cdf5e2fa56cf28c7d412bb6f18210035e2/packages/inertia-vue/index.d.ts#L35
- https://github.com/inertiajs/inertia/blob/b308f0cdf5e2fa56cf28c7d412bb6f18210035e2/packages/inertia-react/index.d.ts#L30

reinink:
@sebastiandedeyne Done, hopefully I did it right (see 6227d04169996ada1908d7472971a471163283a5). Going to merge this in, but if I'm wrong, please let me know. 👍

PR #264, opened 2020-10-17
Merge query strings with data

### LINKED ISSUE #65: Reloading with data duplicate query parameters

### LINKED ISSUE #182: Inertia link href value when using `data` property
First, just wanted to thank everyone involved for the work on Inertia. It's a really fantastic project!

---

I was working on adding pagination to an app and to support that I made use of the `data` property on the `IntertiaLink` component, roughly like:

``` js
<InertiaLink href="/projects" data={{ page: currentPage + 1, tags: ["1", "5"] }}>
  Next
</InertiaLink>
```

- **Clicking** - On click, this link behaved as expected, taking me to `/projects?page=2&tags[]=1&tags[]=5`
- **Cmd-clicking / "Open in New Tab"** - Unfortunately, when cmd-clicking or right click + "Open in new tab", I was instead taken to `/projects`.

Digging into the code a bit, I can see that the `href` prop of the `InertiaLink` components are passed directly to the underlying `<a>` tag, but the `onClick` handler will use `Inertia.visit` function, which in turn will pass the `data` prop to axios and thus generate and use the fully fleshed out URL.

----

One possible solution to this would be to use the `axios.getUri` function to generate the fully-specified URL in the `InertiaLink` component, such that click & cmd-click behavior would match.

```js
const fullHref = axios.getUri({
  url: '/projects',
  params: {
    page: 2,
    tags: ["1", "5"],
  }
})

// fullHref === "/projects?page=2&tags[]=1&tags[]5"
```

And then that `fullHref` value could be used in place of the default `href` in

``` js
// InteriaLink component
return createElement('a', { ...props, fullHref, onClick: visit }, children)
```

reinink:
Fixes #65
Fixes #182

This PR updates Inertia to automatically deep merge query strings with data for `GET` links and visits.

Previously if you had a link like `/users?page=1`, and set the data for that link to `{ page: 1 }`, you'd end up with a visit made to `/users?page=1&page=1`.

Further, previously if you had a link that set the query strings via the `data` property, "Open in New Tab" behaviour was broken, since the rendered `href` didn't include the data in the query string. For example:

```vue
<inertia-link href="/users" :data="{ page: 2 }">Next page</inertia-link>
```

Currently this will render a link with an `href` of `/users`, when what we actually want is `/users?page=2`. This PR updates the render functions of each link component to automatically merge the `data` into the `href` as query strings.

- [x] Update core
- [x] Update the Vue 2 adapter
- [x] Update the Vue 3 adapter
- [x] Update the React adapter
- [ ] Update the Svelte adapter


claudiodekker:
LGTM 👍 

The only thing I'm wondering is whether it would make sense to make some of these utilities part of an `utils` folder as to keep the root-level cleaner? At the same time, perhaps some of these URL-related things could be one file (e.g. `utils/url`) with multiple exports?

reinink:
@sebastiandedeyne If you could review my work on the React adapter, I'd appreciate it.

@pedroborges I honestly have no idea how to implement this in Svelte. I think I'm going to need your help. 😕

packages/inertia/src/mergeQueryStringsWithData.js
```diff
@@ -0,0 +1,10 @@
+import qs from 'qs'
+import deepmerge from 'deepmerge'
+
+export default function mergeQueryStringsWithData(method, url, data) {
```
sebastiandedeyne, line 4: `?foo=bar&baz=qux` is a (singular) query string.

```suggestion
export default function mergeQueryStringWithData(method, url, data) {
```

I believe singular is a better method name, otherwise it sounds like it accepts multiple query strings.

packages/inertia-react/src/Link.js
```diff
@@ -65,5 +65,9 @@ export default forwardRef(function InertiaLink({
     ],
   )
 
+  const [url, _data] = mergeQueryStringsWithData(method, hrefToUrl(href), data)
+  href = url.href
+  data = _data
```
sebastiandedeyne, line 70: The `_` is a bit icky looking, but I can't immediately come up with a better solution.

reinink: Yup, totally agree. This even took me a while to come up with.

packages/inertia/src/mergeQueryStringsWithData.js
reinink: Great point, will update. 👍

reinink:
@pedroborges Update: I've managed to kind of get this working in the Svelte adapter, with one limitation—it's not reactive. Meaning, if you update the `data` reactively for whatever reason, the `href` is not changing.

Here's the test I'm using:

```html
<script>
  let pagination = { page: 2 }
</script>

<!-- Using inertia directive -->
<a href="/organizations" use:inertia={{ data: pagination }}>Next page</a>

<!-- Using InertiaLink component -->
<InertiaLink href="/organizations" data={pagination}>Next page</InertiaLink>

<!-- Change pagination data -->
<button type="button" on:click={() => pagination = { page: 3 }}>Change next page</button>
```

What's interesting is that if you actually click either link, the data *is* correct (meaning that part updates), however the `href` doesn't update...meaning it's not re-rendering when the `data` changes.

Thoughts?

jordankaerim:
From [Svelte's API docs](https://svelte.dev/docs#use_action):

> An action can have a parameter. If the returned value has an update method, it will be called whenever that parameter changes, immediately after Svelte has applied updates to the markup.

So I think you should be able to do something like this to make it reactive.

```html
<script>
	let query = "?page=1"
	
	function directive(node, data) {
		node.href += data.query
			
		return {
			update(data) {
				node.href = data.query
			}
	}
}
</script>

<a href="https://inertiajs.com" use:directive={{ query }}>Inertia Website</a>

<button on:click={() => query = "?page=2"}>
	Page 2
</button>
```

[Link to REPL](https://svelte.dev/repl/2f2e59de197b42fdb0aee9e7fea745e2?version=3.29.0)

jordankaerim:
From [Svelte's API docs](https://svelte.dev/docs#use_action):

> An action can have a parameter. If the returned value has an update method, it will be called whenever that parameter changes, immediately after Svelte has applied updates to the markup.

So I think you should be able to do something like this to make it reactive.

```html
<script>
  let query = "?page=1"
	
  function directive(node, data) {
    node.href += data.query
			
    return {
      update(data) {
        node.href = data.query
      }
  }
}
</script>

<a href="https://inertiajs.com" use:directive={{ query }}>Inertia Website</a>

<button on:click={() => query = "?page=2"}>
	Page 2
</button>
```

[Link to REPL](https://svelte.dev/repl/2f2e59de197b42fdb0aee9e7fea745e2?version=3.29.0)

jordankaerim:
From [Svelte's API docs](https://svelte.dev/docs#use_action):

> An action can have a parameter. If the returned value has an update method, it will be called whenever that parameter changes, immediately after Svelte has applied updates to the markup.

So I think you should be able to do something like this to make it reactive.

```html
<script>
  let query = "?page=1"
	
  function directive(node, data) {
    const url = node.href;

    node.href += data.query;
			
    return {
      update(data) {
        node.href = url + data.query;
      }
    }
  }
</script>

<a href="https://inertiajs.com" use:directive={{ query }}>Inertia Website</a>

<button on:click={() => query = "?page=2"}>Page 2</button>
```

[Link to REPL](https://svelte.dev/repl/2f2e59de197b42fdb0aee9e7fea745e2?version=3.29.0)

reinink:
@jordankaerim Awesome, thanks for that! So obvious now. 😂

I was also able to update the `Link` component to reactively update by using a `beforeUpdate` hook. If you could review my implementation, I'd be grateful. Find this change here: 99dbd02fdb236020453ab4f46ca446d94de97535

packages/inertia/package.json
```diff
@@ -19,7 +19,9 @@
     "watch": "microbundle watch --format cjs"
   },
   "dependencies": {
-    "axios": "^0.19.0 || ^0.20.0"
+    "axios": "^0.19.0 || ^0.20.0",
+    "deepmerge": "^4.2.2",
+    "qs": "^6.9.4"
```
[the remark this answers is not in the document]
reinink: Basically my goal here being to improve the chances that multiple packages can share the same version of these two packages, since they are both very popular and used in other packages a lot. I'm just not sure how low to go with the versions.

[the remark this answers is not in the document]
claudiodekker: Yeah, I think that's fine.

PR #279, opened 2020-10-23
Add link slot/render function support (vue2)

### LINKED ISSUE #273: Add link slot/render function support
As suggested by @sebastiandedeyne [here](https://github.com/inertiajs/inertia/issues/268#issuecomment-712277597), it would be great to add slot/render function support to the Inertia link. This provides a more advanced API for constructing Inertia links.

Vue example:

```vue
<inertia-link href="/logout" method="POST" v-slot="{ visit }">
  <button @click="visit">Log out<button>
</inertia-link>
```

React example:

```js
<InertiaLink href="/logout" method="POST">
  {({ visit }) => <button onClick="visit">Log out</button>}
</InertiaLink>
```

This is similar to the `v-slot` API added to Vue Router in 3.1.0 (see [here](https://router.vuejs.org/api/#v-slot-api-3-1-0)).

nickknissen:
Should resolve #273 for vue2

I use Vue2 as my daily driver, but can look into vue3 if needed.

### DIFF
--- a/packages/inertia-vue/src/link.js
+++ b/packages/inertia-vue/src/link.js
@@ -40,7 +40,7 @@ export default {
       default: () => ({}),
     },
   },
-  render(h, { props, data, children }) {
+  render(h, { props, data, children, scopedSlots }) {
     data.on = {
       click: () => ({}),
       cancelToken: () => ({}),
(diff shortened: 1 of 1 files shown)


reinink:
So while this PR is cool, the Inertia Link `as` attribute really solves this problem, and if that doesn't, it's not hard to manually wire up a custom button with a `click` handler that just uses `Inertia.get()` (or whatever). Given that, and the fact that we haven't had anyone really running into limitations, I think I'm going to close this one. Thanks either way for your efforts @nickknissen!

PR #280, opened 2020-10-23
Update type definitions

RobertBoes:
This PR adds InertiaLink's missing type definitions for React and Vue.

For React I also allowed all HTML attributes, since you can now do the following to create a button: `<InertiaLink as="button" type="button" href="/test">Some test!</InertiaLink>`. I'm not sure if this is the correct way to do this in React/TS, but I've tested this and it works and the builds finish without errors.

The Vue TS definitions probably also require a change to allow HTML attributes, but I couldn't figure out how to do this with Vue.

To test this I simply created a React TS app (`npx create-react-app my-app --template typescript`), installed inertia/inertia-react, made the changes within the node_modules folder and simply added an InertiaLink with new Inertia attributes and some default HTML attributes.

### DIFF
--- a/packages/inertia-react/index.d.ts
+++ b/packages/inertia-react/index.d.ts
@@ -41,7 +41,7 @@ interface InertiaLinkProps {
 }
 
 type InertiaLink = React.FunctionComponent<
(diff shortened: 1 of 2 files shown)


reinink:
Thanks @RobertBoes!

PR #286, opened 2020-10-26
Add Typescript definitions for 'App' and 'plugin' named imports

wazum:
Following the recent changes on client-side framework setup:

```js
import { App, plugin } from '@inertiajs/inertia-vue'
```

### DIFF
--- a/packages/inertia-vue/index.d.ts
+++ b/packages/inertia-vue/index.d.ts
@@ -50,3 +50,7 @@ type InertiaLink = FunctionalComponentOptions<InertiaLinkProps>
 export const InertiaLink: InertiaLink
 
 export const InertiaApp: App
+
+export const App: App
+
+export const plugin: PluginObject


packages/inertia-vue/index.d.ts
```diff
@@ -50,3 +50,7 @@ type InertiaLink = FunctionalComponentOptions<InertiaLinkProps>
 export const InertiaLink: InertiaLink
 
 export const InertiaApp: App
+
+export const App: App
+
+export const plugin: PluginObject
```
reinink, line 56: This is this `PluginObject`, and not just `plugin`?

wazum: It's this one
https://github.com/vuejs/vue/blob/bb253db0b3e17124b6d1fe93fbf2db35470a1347/types/plugin.d.ts#L5

reinink: Ahh, gotcha. 👍

PR #290, opened 2020-10-26
Fix finish event from firing in the wrong sequence

### LINKED ISSUE #288: Make custom loading indicators easier to implement
So, with addition of the new event system, it's now possible to use your custom loading indicators with Inertia. In fact, `@inertiajs/progress` itself uses the event system to do this ([see here](https://github.com/inertiajs/progress/blob/master/src/progress.js#L24-L26)).

My concern, however, is that implementing these custom loading indicators never turned out as simple as I had hoped it would. Here was my original vision for this:

```js
import NProgress from 'nprogress'

Inertia.on('start', () => NProgress.start())
Inertia.on('progress', event => NProgress.set(event.detail.progress.percentage / 100 * .9))
Inertia.on('finish', () => NProgress.done())
```

(Note, this limits the progress to 90%, since 100% uploaded doesn't represent a finished visit, since you still need to wait for the server to respond).

And while this certainly works on the happy path, there are some issues and limitations to this approach:

## Preventing visits

If another `start` event listener calls `event.preventDefault()` (cancelling the visit), this implementation still starts the loading indicator, which is then never stopped, since the `finish` event never gets fired.

This can be solved by "waiting for the next tick", and then checking `event.defaultPrevented` on the event, like this:

```js
queueMicrotask(() => {
  if (event.defaultPrevented) {
    return
  }

  Inertia.on('start', () => NProgress.start())
})
```

And while this certainly works, the fact that you need this in order to properly implement your custom loading indicators is not easily discoverable.

## Cancelling visits

Even more difficult than handling prevented visits, is handling cancelled visits. More specifically, visits that are cancelled as a result of a visit to a new page (ex: clicking on another link before the current visit finishes).

The problem here is the order that the events get fired:

1. `start` is fired on the first page visit, which causes the progress indicator to start.
2. The user clicks another link, which causes a second `start` event to get fired (no change to the progress indicator).
3. The first visit asynchronously cancels, and fires the `finish` event, which causes the progress indicator to finish (clear).
4. The second visit continues to load, but now there is no progress indicator for it.
5. The second visit finishes, and fires the `finish` event, but there is no progress indicator to clear.

The heart of the issue here is the fact that the visit cancellation happens asynchronously, which means we can't force the `finish` event to fire before the new `start` event. (Ideally, I'd love to fix this, but I don't think if it's even technically possible, since I assume it's the browser that doing the `xhr` cancellation asynchronously.)

The solution here is to keep track of how many visits are currently in progress, and increment and decrement them as they start/finish, and then only clear the loading indicator when the count is back to zero.

You can see how I implement this in the `@inertiajs/progress` library using the `inProgress` variable ([see here](https://github.com/inertiajs/progress/blob/master/src/progress.js#L6)).

## Delay behaviour

The other thing the custom loading indicator implementation above doesn't handle is delayed showing of the loading indicator. When requests are fast (less than roughly 250ms), it's actually better to *not* show the loading indicator, since it really just causes a visual distraction to the page. Instead, it's better to only display the loading indicator in the event that the request is going slow. I have found 250ms to be a good threshold here. Turbolinks does the same thing, and actually delays for 500ms by default.

 To implement this properly requires some clever use of `setTimeout` and `clearTimeout`. You can see how I implement this in the `@inertiajs/progress` library using the `timeout ` variable ([see here](https://github.com/inertiajs/progress/blob/master/src/progress.js#L5)).

## Making things better

So, it's now clear to me that while the new event system can be used to implement your own custom loading indicator, in all likelihood, people are probably going to do it wrong, simply because it requires a pretty deep knowledge of how the event system works to get right. I think it would be better for Inertia to provide an abstraction layer over this that handles all this implementation detail for you.

This can be done in four ways:

## 1. A new progress package

The first option is to create a new generic progress package, similar to `@inertiajs/progress`, except without the NProgress integration. Under the hood this would use the event system, except it would provide three callbacks that can be used to implement your own custom loading indicator.

```js
import NProgress from 'nprogress'
import { progress } from '@inertiajs/generic-progress'

progress.init({
  delay: 500,
  onStart: () => NProgress.start(),
  onProgress: percentage => NProgress.set(percentage / 100 * .9),
  onFinish: () => NProgress.done(),
})
```

## 2. Add progress configuration to Inertia

The second option is to provide new configuration options into the Inertia adapters (and core). For example, we could add a new `loadingIndicator` prop that defines three callbacks:

```js
createApp({
  render: () => h(App, {
    loadingIndicator: {
    
      onStart: () => NProgress.start(),
      onProgress: percentage => NProgress.set(percentage / 100 * .9),
      onFinished: () => NProgress.done(),
    },
  }),
}).use(plugin).mount(el)
```

## 3. Add new progress events

The second option is a new set of events that we fire specifically for the purpose of the loading indicator:

```js
Inertia.on('progress-start', () => NProgress.start())
Inertia.on('progress-increment', event => NProgress.set(event.detail.percentage / 100 * .9))
Inertia.on('progress-finish', () => NProgress.done())
```

The nice thing about this approach is that progress packages can simply listen for these events, just like the `@inertiajs/progress`, without any dependencies on Inertia core.

What I really don't like is that we now have a `start` and `progress-start` event, a `progress` and `progress-increment` event, and a `finish` and `progress-finish` event. From for developer experience perspective, this seems really confusing. "What exactly is the difference between `start` and `progress-start`?"

I worry that this design just indicates a flaw in the event system design in general (see option four for more details).

## 4. Tweak the event system

I'd actually be interested in exploring what the implications of only firing the current `finish` event when <ins>all visits are done</ins> would be. Meaning, `finish` could technically get fired less times than `start`. I suspect this would actually be totally fine, since I suspect the primary use-case for the `finish` event is for hiding loading indicators. Further, the per-visit `onFinish` callback would still always be called, should you need to do clean up there.

Here is realistic NProgress implementation with this change...which honestly isn't bad.

```js
import NProgress from 'nprogress'

let delay = 500
let timeout = null

Inertia.on('start', event => {
  queueMicrotask(() => {
    if (!event.defaultPrevented) {
      clearTimeout(timeout)
      timeout = setTimeout(() => {
        NProgress.set(0)
        NProgress.start()
      }, delay)
    }
  })
})

Inertia.on('progress', event => {
  if (NProgress.isStarted() && event.detail.progress.percentage) {
    NProgress.set(event.detail.progress.percentage / 100 * .9)
  }
})

Inertia.on('finish', () => {
  NProgress.done()
  clearTimeout(timeout)
})
```

This could be further simplified by adding a new new `before-start` event. This event would basically be exactly the same as `start`, except you'd only be able to cancel visits using `before-start`, and not via `start`. Meaning, `start` will only fire if `event.preventDefault()` wasn't called on the `before-start` event.

Here's an updated NProgress implementation for this change, which no longer requires the `queueMicrotask()` call and `event.defaultPrevented` check:

```js
import NProgress from 'nprogress'

let delay = 500
let timeout = null

Inertia.on('start', () => {
  clearTimeout(timeout)
  timeout = setTimeout(() => {
    NProgress.set(0)
    NProgress.start()
  }, delay)
})

Inertia.on('progress', event => {
  if (NProgress.isStarted() && event.detail.progress.percentage) {
    NProgress.set(event.detail.progress.percentage / 100 * .9)
  }
})

Inertia.on('finish', () => {
  NProgress.done()
  clearTimeout(timeout)
})
```

As I write this, I'm really starting to think that the solution here is to change the `finish` event to only fire when <ins>all visits</ins> are done. This approach would allow us to avoid creating a new progress package, or adding additional configuration options to Inertia, or adding new progress specific events. Instead, we can simply document how to add custom loading indicators, sharing the above code snippet.

The only question left for me is whether it makes sense to add the new `before-start` event (which can be cancelled), and remove cancellation behaviour from the `start` event? My gut here is no, simply because there is a reasonable workaround using `queueMicrotask` and `event.preventDefault()`, and the amount of times you'll actually need to use this is going to be very minimal.

Basically this solution solves the most annoying/confusing part of displaying custom loading indicators, which is handling visit cancellations, where you get two `start` events fired and then two `finish` events fired.

reinink:
Closes #288

## The issue

As explained in issue #288, Inertia currently has an issue where cancelled visits are not firing the `finish` event in the right sequence. For example, if you click a link to visit a page, and then click another link before that page loads, you end up with the following events being fired: `start`, `start`, `finish`, `finish`.

This happens because the xhr cancellation happens asynchronously. This creates weird issues when using the event system, when you expect the events to fire in a certain order, but they don't. For example, when using the event system to setup custom progress indicators.

## Firing `finish` event synchronously

This PR corrects this issue by firing the `finish` event for cancelled visits right at the point of cancellation for both manually cancelled visits, as well as visits interrupted as a result of visiting another page.

Previously we fired this event in the `finally()` callback of the axios visit promise, and while this was fine for visits that finish properly, it was too late for interrupted visits. This meant that the new visit `start` event would be fired before the `finish` event for the cancelled visit.

## New `before-start` event

We also had to add a new `before-start` event, and corresponding `onBeforeStart()` callback. This event is identical to the existing `start` event, in that it receives the `visit` and can also be cancelled. This is now the event that you use to cancel a visit, and the `start` event is no longer cancellable.

Previously the `start` event listener was fired *before* the logic that cancelled any active visits (and subsequent `finish` events). This order was necessary, since the `start` event could be cancelled, in which case we didn't want to cancel any active visits. Unfortunately this meant that the `start` event was always called before the `finish` event caused by a cancelled event.

With this new event, events always fire in the correct order, while still having the ability to cancel visits.

1. `before-start`
2. `start`
3. `finish`

## Simplified `start` event usage

This change had an added benefit, in that it simplified the `start` event usage. Previously, if you had multiple `start` event listeners, and one of those listeners cancels the visit, the other listeners would still run as if it wasn't cancelled. Meaning, a loading indicator might appear, even though the visit never happened.

The workaround was to use `queueMicrotask` and `event.defaultPrevented` to detect in each event listener if the visit was cancelled. This was both complicated and not obvious to end users using this event.

## New `finish` event types

The `finish` event now also receives an instance of the `visit` that was made (or cancelled), which includes three "status" properties on it:

- `visit.completed`: When a visit finishes normally.
- `visit.cancelled`: When a visit is manually cancelled.
- `visit.interrupted`: When a visit is cancelled as a result of a new page visit.

What's really great about these visit statuses is they can be really useful when managing custom loading indicators. For example, as you can see in this [corresponding progress PR](https://github.com/inertiajs/progress/pull/5), we are now able to be smarter about how we handle things in the "finish" event listener:

- `visit.completed`: The progress bar completes by going all the way to the right.
- `visit.cancelled`: The progress bar is immediately removed from the page.
- `visit.interrupted`: The progress bar resets back to the start position, and continues for the new visit.

## Organized events

This PR also organizes all the events into a new `events.js` file, with exported functions for reach event type.


packages/inertia/src/inertia.js
```diff
@@ -136,11 +136,13 @@ export default {
     return response?.headers['x-inertia']
   },
 
-  cancelActiveVisits() {
-    if (this.cancelToken) {
-      this.cancelToken.cancel()
+  cancelActiveVisit({ cancelled = false, interrupted = false }) {
+    if (this.activeVisit) {
+      this.activeVisit.cancelToken.cancel()
+      this.fireEvent('finish', { detail: { visit: { ...this.activeVisit, completed: false, cancelled, interrupted } } } )
```
claudiodekker, line 142: Maybe we can place this in a generic onFinish method so that the event signature is only managed from a single place?

reinink: So yeah, I am thinking about actually creating a series of event helpers:

```js
this.fireBeforeStartEvent()
this.fireErrorEvent()
this.fireFinishEvent()
this.fireInvalidEvent()
this.fireNavigateEvent()
this.fireProgressEvent()
this.fireStartEvent()
this.fireSuccessEvent()
```

Because it's becoming hard to keep track off, and is making the source code kind of messy.

claudiodekker: Could even place them in an `events.js` and then include that into the instance, to keep them separate from the rest 👍

packages/inertia/src/inertia.js
```diff
@@ -165,23 +168,29 @@ export default {
...
-
-    let visit = { url, ...arguments[1] }
-    if (onStart(visit) === false || !this.fireEvent('start', { cancelable: true, detail: { visit } } )) {
+    const visit = { url, method, data, replace, preserveScroll, preserveState, only, headers, onCancelToken, onBeforeStart, onStart, onProgress, onFinish, onCancel, onSuccess }
+    if (onBeforeStart(visit) === false || !this.fireEvent('before-start', { cancelable: true, detail: { visit } } )) {
       return
     }
-    this.cancelActiveVisits()
+
+    this.cancelActiveVisit({ interrupted: true })
```
claudiodekker, line 176: Wouldn't this always mark the first first visit made as interrupted?

packages/inertia/src/inertia.js
reinink: Yeah cool, let me try.

reinink: And done 👍

claudiodekker (approved):

anubra266:
Thanks for this, Could you add the prediction so it can show in our IDE Intellisense. 
I kept reupdating Inertia, as i thought i didn't do it right, because the suggestion did not come up. It works thought, eneded up trying it in faith.
Thanks again.

PR #293, opened 2020-10-27
Update reload() method to preserve scroll position

reinink:
The goal with `Inertia.reload()` is to refresh the page data (props) using a nice terse API. This method is similar to `location.reload()`, except `location.reload()` also preserves the scroll position, where `Inertia.reload()` doesn't currently. This PR updates it to do that.

This also forces the `preserveState` to `true`, instead of allowing you to override this via the `options`. If you don't want `preserveScroll` and `preserveState`, use a different method. 👍 

I also took the opportunity to break the dependency on `Inertia.replace()`, which is being deprecated.

We also don't need to explicitly set `replace: true`, since that happens automatically by making a visit to the same URL.

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -310,7 +310,7 @@ export default {
   },
 
   reload(options = {}) {
-    return this.visit(window.location.href, { preserveScroll: true, preserveState: true, ...options })
+    return this.visit(window.location.href, { ...options, preserveScroll: true, preserveState: true })
   },
 
   replace(url, options = {}) {
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -310,7 +310,7 @@ export default {
   },
 
   reload(options = {}) {
-    return this.replace(window.location.href, options)
+    return this.visit(window.location.href, { preserveScroll: true, preserveState: true, ...options })
   },
 
   replace(url, options = {}) {


claudiodekker (approved):

claudiodekker:
LGTM 👍  

The only thought I had is: Do we perhaps want to _enforce_ those parameters when using this method, instead of allowing the user to override it? It might seem unfriendly, but it's also kind of a way to say "if you need different behaviour, use something else".

reinink:
@claudiodekker I had the exact same thought.

If you're going to do:

```js
Inertia.reload({ preserveState: false })
```

You should probably do:

```js
Inertia.get(location.href, { preserveScroll: true })
```

Because it's no longer what a "reload", it's a visit with scroll preserved.

Will update. 👍

claudiodekker (approved):

fourstacks:
Just in case anyone stumbles on this PR, the example of what you should do if you want `preserveState: false` is a little wrong.  Should be:

```
Inertia.get(location.href, {}, { preserveScroll: true })
```
The second param is for data, rather than options - just tripped me up for a few mins so figured it was worth mentioning!

barryp:
It seems inconsistent to me that .post(), .put(), .patch(), and .delete() are coded like 

    { preserveState: true, ...options,  

so that the caller can override the default behavior while .reload() is coded as 

    { ...options, preserveScroll: true, preserveState: true } 

so they can't be overridden

PR #297, opened 2020-10-28
Clone page before transforming props

### LINKED ISSUE #278: $inertia.delete sometimes throws weird error
I have the following code:
```
this.$inertia.delete(this.$route("users.destroy", user.id));
```
It works fine.
But when I put it in a arrow function like so:
```
.then(() => {
   this.$inertia.delete(this.$route("users.destroy", user.id));
});
```

...it throws the following error:
```
DOMException: Failed to execute 'replaceState' on 'History': function () { [native code] } could not be cloned.
```
Once this happens, all subsequent inertia visits throw the same error.

Any idea what's causing it?
I'm using the Vue adapter.

### LINKED ISSUE #295: Passing function to transformProps causes error
app.js:
```
// ...
  return (
    <StyletronProvider value={engine}>
      <BaseProvider theme={state.theme}>
        <Block {...blockProps}>
          <InertiaApp
            initialPage={JSON.parse(app.dataset.page)}
            resolveComponent={name =>
              import(`./Pages/${name}`).then(module => module.default)
            }
            transformProps={props => ({
              ...props,
              toggleTheme: () => {
                // ...
              }
            })}
          />
        </Block>
      </BaseProvider>
    </StyletronProvider>
  );
// ...
```
Passing function to transformProps causes same error as here #278:
```
Uncaught (in promise) DOMException: Failed to execute 'replaceState' on 'History': function toggleTheme() {// ...
        } could not be cloned.
    at Object.replaceState (webpack-internal:///./node_modules/@inertiajs/inertia/dist/index.js:1:9008)
    at eval (webpack-internal:///./node_modules/@inertiajs/inertia/dist/index.js:1:8704)
```
https://github.com/inertiajs/inertia/blob/c7a39e57d26c86a88513fc063f7d4fc74a9eb232/packages/inertia/src/inertia.js#L296

An empty function also throws this error.

I am using IIS with `yarn hot`.

The strange thing is that  passing function to `transformProps` worked for me before I updated Inertia packages. I don't remember what version I had before

reinink:
Fixes #278
Fixes #295 

This fixes an issue where, when transforming props, if you put non-primitives (functions/classes) into your page props, it generates the following error:

```
Uncaught (in promise) DOMException: Failed to execute 'replaceState' on 'History': function () { ... } could not be cloned.
```

This is because the object that's put in history state must be something that can be serialized, and non-primitives cannot be.

To fix this, we need to deep clone the `page` object, and then do the props transforming on that, and then pass that to the `swapComponent()` method to be used in our apps.

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -269,14 +269,15 @@ export default {
   setPage(page, { visitId = this.createVisitId(), replace = false, preserveScroll = false, preserveState = false } = {}) {
     return Promise.resolve(this.resolveComponent(page.component)).then(component => {
       if (visitId === this.visitId) {
-        page.props = this.transformProps(page.props)
         page.scrollRegions = page.scrollRegions || []
         page.rememberedState = page.rememberedState || {}
         preserveState = typeof preserveState === 'function' ? preserveState(page) : preserveState
         preserveScroll = typeof preserveScroll === 'function' ? preserveScroll(page) : preserveScroll
         replace = replace || hrefToUrl(page.url).href === window.location.href
         replace ? this.replaceState(page) : this.pushState(page)
-        this.swapComponent({ component, page, preserveState }).then(() => {
+        const clone = JSON.parse(JSON.stringify(page))
+        clone.props = this.transformProps(clone.props)
+        this.swapComponent({ component, page: clone, preserveState }).then(() => {
           if (!preserveScroll) {
             this.resetScrollPositions()
           }


sebastiandedeyne (approved):

m1guelpf:
I'm still getting `Uncaught DOMException: Failed to execute 'replaceState' on 'History': [object Object] could not be cloned.` from `replaceState` when some of the data passed is wrapped in a `Proxy`.

anubra266:
Mine was displaying this error in the use React useRememberedState hook.

reinink:
@m1guelpf Are you on `@inertiajs/inertia@v0.6.1`?

m1guelpf:
@reinink Yea

reinink:
@m1guelpf To be clear, are you using the `transformProps` feature to put a `Proxy` object into props? Or are you doing something different?

m1guelpf:
@reinink I'm trying to use Jetstream's inertia form plugin with Vue 3, which means that the props are computed properties now and that seems to mess up stuff quite a bit.

m1guelpf:
See https://github.com/laravel/jetstream-js/pull/12

reinink:
@m1guelpf Yeah, okay, so I think that's a different issue. I wouldn't even think that Jetstream would work with Vue 3 yet.

m1guelpf:
@reinink I've now been able to reproduce it with the non-composition API too using the currently-released code (aka without doing weird stuff:tm:).

reinink:
Can you submit this as a new issue (with steps to reproduce), as this particular issue was specifically for the `transformProps` behaviour. Thanks! 👍

PR #303, opened 2020-10-28
Update peer dependency

herpaderpaldent:
according to the upgrade instructions one must use `@inertiajs/inertia` to `^0.6.0` yet the peer dependency refers still on `^0.5.0`


reinink:
Technically it still works with `@inertiajs/inertia@0.5.0`, you just don't get access to the new `before` event. This is why I left the peer dependency at `0.5.0`. 👍

PR #306, opened 2020-10-29
Fix inertia-link crashing when self-closing

m1guelpf:
This PR fixes an error triggered when you create self-closing `inertia-link` component, which sets `slots` to null:

```vue
<inertia-link class="absolute inset-0 z-10" :href="route('dashboard.edit', dashboard).url()" />
```

### DIFF
--- a/packages/inertia-vue3/src/link.js
+++ b/packages/inertia-vue3/src/link.js
@@ -75,7 +75,7 @@ export default {
             })
           }
         },
-      }, slots.default())
+      }, slots && slots.default() || '')
     }
   },
 }
--- a/packages/inertia-vue3/src/link.js
+++ b/packages/inertia-vue3/src/link.js
@@ -75,7 +75,7 @@ export default {
             })
           }
         },
-      }, slots && slots.default() || '')
+      }, slots)
     }
   },
 }


packages/inertia-vue3/src/link.js
```diff
@@ -75,7 +75,7 @@ export default {
             })
           }
         },
-      }, slots.default())
+      }, slots && slots.default() || '')
```
lukaskleinschmidt, line 78: I just ran into the same issue.
From what I have tested you can also just pass `slots` here too.

claudiodekker:
Use-case: You have a list of blog posts, with a title, summary & "read more" link, but want the entire blog post to be clickable and behave like the "read more" link while still keeping links that are part of the summary functional.

Because HTML can't have nested links, there's two solutions to this problem:

1. Make the parent div clickable instead of wrapping it in a link (vue: `<div @click="readMoreAction" ...`)
2. [Create an empty link that you overlay on top of the whole blog post summary (z-index), and make the links inside the summary an even higher z-index so they end up on top of the overlay.](https://www.sarasoueidan.com/blog/nested-links/)

This PR aims to solve the 2nd approach.

dillingham:
@claudiodekker ran into this when copying the pingcrm paginator into a jetstream project

```html
<inertia-link v-else :key="key" class="mr-1 mb-1 px-4 py-3 text-sm leading-4 border rounded hover:bg-white focus:border-indigo-500 focus:text-indigo-500" :class="{ 'bg-white': link.active }" :href="link.url" v-html="link.label" />
```

Notably, the `v-html` at the end / absence of default slot.

packages/inertia-vue3/src/link.js
claudiodekker: I think that might actually be the better solution
```suggestion
      }, slots)
```

reinink:
Awesome, thanks for fixing this! 🙌

PR #312, opened 2020-11-03
Add links to be read out by npm

AdrianMrn:
This allows npm to display links like this:

image

This is useful because on Google, results for some queries including `inertia` still show the npm page higher than the GitHub repo and homepage.


reinink:
Nice, thanks @AdrianMrn! 💪

PR #315, opened 2020-11-05
Add new "inline visits" feature

reinink:
This PR adds the ability to make "inline visits", which lets you add a second page component to the current page with its own props and URL. This second page component is rendered as a child of the existing page component, and can be positioned elsewhere on the page using portals.

- [x] Core
- [ ] React adapter
- [ ] Svelte adapter
- [ ] Vue 2 adapter
- [x] Vue 3 adapter

## How to use it

To make an inline visit, set the `inline` attribute on an Inertia link:

```vue
<inertia-link href="/users/create" inline>
```

You can also do this programmatically:

```js
Inertia.get('/users/create', {}, { inline: true })
```

You can access the inline page details via `$page.property (Vue) or `usePage().inline`.

The `inline` property accepts three possible values:

- `true`: Inline the response
- `maintain`: Inline if the current page is inlined
- `same`: Inline if the current page is inlined, and the component is the same

## Name

This feature is not properly named yet. Working name ideas (bold ones I like):

- **inline**
- partial
- insert
- append
- **stack**
- **nest**
- fragment
- sibling

## Brainstorming notes

1. If Inertia *partial visit*
    - Add the partial to the current page
2. Otherwise:
    - Prepare partial object (`component`, `props`, `url`), and flash it in the session
    - Redirect to `fallack` url
    - When sending response back to browser
        - use fallback page as the main page (`page`)
        - and use the partial as the partial (`page.partial`)
        - with the partial URL as the url (`page.partial.url)

```php
class UserController
{
    public function index()
    {
        return Inertia::render('Users/Index', [
            'users' => User::with('company')->get()->map(fn ($user) => [
                'id' => $user->id,
                'name' => $user->name,
                'company' => $user->company->name,
            ])
        ]);
    }

    public function create()
    {
        return Inertia::partial('Users/Create', [
            'companies' => Company::get(['id', 'name']),
        ])->fallback('/users');
    }
}
```

### DIFF
--- a/packages/inertia-vue3/src/app.js
+++ b/packages/inertia-vue3/src/app.js
@@ -4,8 +4,8 @@ import { computed, h, markRaw, ref } from 'vue'
 import { Inertia } from '@inertiajs/inertia'
 
 const component = ref(null)
+const inlineComponent = ref(null)
 const page = ref({})
-const inline = ref({})
 const key = ref(null)
 
 export default {
@@ -31,9 +31,9 @@ export default {
       transformProps,
       swapComponent: async (args) => {
         component.value = markRaw(args.component)
+        inlineComponent.value = args.inlineComponent ? markRaw(args.inlineComponent) : null
         page.value = args.page
-        key.value = (args.preserveState || args.inline) ? key.value : Date.now()
-        inline.value = args.inline ? { component: markRaw(args.inline.component), props: args.inline.props, url: args.inline.url } : null
+        key.value = args.preserveState ? key.value : Date.now()
       },
     })
 
@@ -56,13 +56,13 @@ export default {
 
           return [
             h(component.value.layout, () => child),
-            inline.value ? h(inline.value.component, inline.value.props) : null,
+            inlineComponent.value ? h(inlineComponent.value, page.value.inline.props) : null,
           ]
         }
 
         return [
           child,
-          inline.value ? h(inline.value.component, inline.value.props) : null,
+          inlineComponent.value ? h(inlineComponent.value, page.value.inline.props) : null,
         ]
       }
     }
@@ -73,7 +73,6 @@ export const plugin = {
   install(app) {
     Object.defineProperty(app.config.globalProperties, '$inertia', { get: () => Inertia })
     Object.defineProperty(app.config.globalProperties, '$page', { get: () => page.value })
-    Object.defineProperty(app.config.globalProperties, '$inline', { get: () => inline.value })
     app.mixin(remember)
     app.component('InertiaLink', link)
   },
@@ -85,6 +84,5 @@ export function usePage() {
     url: computed(() => page.value.url),
     component: computed(() => page.value.component),
     version: computed(() => page.value.version),
-    inline: computed(() => inline.value),
   }
 }
--- a/packages/inertia-vue3/src/app.js
+++ b/packages/inertia-vue3/src/app.js
@@ -5,7 +5,7 @@ import { Inertia } from '@inertiajs/inertia'
 
 const component = ref(null)
 const page = ref({})
-const partials = ref([])
+const partial = ref(null)
 const key = ref(null)
 
 export default {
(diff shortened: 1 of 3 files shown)


wilburpowery:
This can't come soon enough folks!! Keep up the amazing work! ♥️

jeffreyvanhees:
> This can't come soon enough folks!! Keep up the amazing work! ♥️

Haha, is your project stagnating because you want to use this feature? Just like with me!

claudiodekker:
Sorry for the delay, we'll be looking into this soon-ish! 😅

A-Lawrence:
> Sorry for the delay, we'll be looking into this soon-ish! 😅 

There's plenty of excitement around this one 🙏

reinink:
Hey folks! Believe me, I want this feature done as well, but there is lots of testing and work to do here still. I'm pretty slammed with work right now, and probably will be until the end of the year. I'm hoping to get back into the development of this feature in January. 👍

snaptopixel:
I'm curious what the use case for this is, and why it should be a concern for Inertia? You can already return the same component at different routes and use props for a sub page (and it's props)

At first blush this feels like putting the cart before the horse.

I do think there is a case for a global "layout" with its own props. We rolled our own version of this in our app (which uses web-components) and it's quite useful for things like flash messages, modals, etc. But we were able to do it ourselves with what Inertia provides.

I just don't see how/why adding partials is in line with the vision of the project. If you want to do partials, do them on the server-side and pass as props to the frontend. This is admittedly different than typical SPA stuff, but so is Inertia 😉

aleksblendwerk:
https://twitter.com/reinink/status/1324148296198348800

If I am getting this right, to me this looks like a great way to e.g. open a modal when you are already on a "Create contact" page and notice that you haven't created the new organization to add this contact to yet. The "Create organization" page could be opened in a modal and after saving, you could immediately select that new organization from within the "Create contact" page's "organization" select box.

snaptopixel:
@aleksblendwerk that would work only if inline links had a "reload parent when closed" setting, otherwise the parent page will remain stale (not knowing what happened in the child page)

snaptopixel:
IMHO the problems/solutions shown in the twitter thread by @reinink are not justification for what's being advocated here (a comprehensive partial approach/system) hence my GIF above :)

I do however think there is a case to be made for supporting distinct "layers" in Inertia apps, namely:

- Persistent Layer (global navigation/menus, etc)
- Page Layer (what we know and love currently)
- Transient Layer (Pop-overs, Modals, Toasts, etc)

timkley:
Will it be possible to trigger an `inline` visit from an `inline` page?  
Example: triggering another modal from the `CompanyModal` to create another entity which `Company` can belong to?

girardinsamuel:
Hello there  👋 
What is the expected release date for this ? Would this be like in Q1, Q2 ? 
As a lot of users, I am really looking forward to be able to use this 👍 
Thanks !

spencerrlongg:
@girardinsamuel I just found this response from 4 days ago here: https://github.com/inertiajs/inertia/discussions/249#discussioncomment-362694

adrianb93:
For the time being, could this perhaps be merged for us Vue 3 users?

reinink:
@adrianb93 The thing holding this back isn't the fact that only the Vue 3 adapter is "done", it's that as a core team we're still not totally convinced that this is the right solution. Maybe it is...but we have some other ideas we want to explore first.

claudiodekker:
@reinink haha, you beat me to it!

One thing I'd like to add to @reinink's comment, is that for us it's important that _every client-side adapter_ has feature parity, meaning that all of them (react, vue, vue3 & svelte) should support the same things and work (mostly) the same way. 

This also makes it so that Inertia is and remains one project with support for multiple JS frameworks, instead of multiple projects under the same name that started with the same vision/origin while heading into their own directions over time (_remember Java & C#, anyone?_)

reinink:
Closing this PR in favour of #642, the evolution of these ideas. 👍

PR #323, opened 2020-11-12
Fix type definition for usePage and add missing properties to Page

### LINKED ISSUE #321: [TypeScript React] Incorrect type for usePage hook?
Printing the object returned from `usePage` shows the structure:
```json
{
  "component": "...",
  "props": {},
  "url": "/",
  "version": "...",
  "scrollRegions": [],
  "rememberedState": {}
}
```
However, when trying to access any property on this object, TypeScript fails to compile:
```ts
const page = usePage()
const props = page.props
// ==> Property 'props' does not exist on type 'PageProps'.
```
The type definition for `usePage` in `inertia-react/index.d.ts` is:
```ts
export function usePage<PageProps extends Inertia.PageProps = Inertia.PageProps>(): PageProps
```
Changing the type definition to...
```ts
export function usePage<Page extends Inertia.Page = Inertia.Page>(): Page
```
... seems to fix the error and VS Code autocompletion shows the correct types:

image

I could submit a PR with this change if you think it should be this way? I'm no TypeScript expert which is why I'm asking/checking if this is how it's meant to be.

Also, the `Page` type seems to be missing `scrollRegions` and `rememberedState` so I could include those in the PR, too.

Thanks for reading.

stephensamra:
- Set the return type for `usePage` to `Inertia.Page` instead of `Inertia.PageProps`
- Add `scrollRegions` and `rememberedState` to `Page`

Resolves #321 

/cc @sebastiandedeyne

### DIFF
--- a/packages/inertia-react/index.d.ts
+++ b/packages/inertia-react/index.d.ts
@@ -46,8 +46,8 @@ type InertiaLink = React.FunctionComponent<
 >
 
 export function usePage<
-  PageProps extends Inertia.PageProps = Inertia.PageProps
->(): PageProps
+  Page extends Inertia.Page = Inertia.Page
+>(): Page
 
 export function useRememberedState<RememberedState>(
   initialState: RememberedState,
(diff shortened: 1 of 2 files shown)


sebastiandedeyne (approved):

sebastiandedeyne:
Thanks!

Issue #324, opened 2020-11-13
before event not included in InertiaEvent definition?


franjid:
I was taking a look at the [documentation](https://inertiajs.com/events#before) and found out a `before` event that I could use to develop some feature I'm working on.

But when I try to use it like this:
```
removeBeforeEventListener = Inertia.on('before', (event: any) => {
  event.preventDefault();
})
```

I got this error:
```
TS2345: Argument of type '"before"' is not assignable to parameter of type 'InertiaEvent'.
```

I then was taking a look at the Inertia code and saw this in `index.d.ts`
```
type InertiaEvent = 'start' | 'progress' | 'success' | 'invalid' | 'error' | 'finish' | 'navigate'
```

Is it possible that you guys forgot to add a `before` there? I was also wondering if that event was deprecated, but in `events.ts` I can see that you handle it:
```
export function fireBeforeEvent(visit) {
  return fireEvent('before', { cancelable: true, detail: { visit } } )
}
```

Any idea? If `before` is still a valid event I can make a pull request to add it to `InertiaEvent` ;)

Thanks!

reinink:
> Any idea? If before is still a valid event I can make a pull request to add it to InertiaEvent ;)

Yes, please do this. Right now the types are managed by the community, and we're only supporting PRs to fix them, since I don't use TypeScript myself (yet!).

Thanks @franjid! 👍

PR #328, opened 2020-11-14
Added missing 'before' event to InertiaEvent

franjid:
Fix for [issue: before event not included in InertiaEvent definition](https://github.com/inertiajs/inertia/issues/324)


reinink:
Thanks!

PR #334, opened 2020-11-24
Allow external link clicks with inertia-link

liamjoc:
This PR enables `inertia-link` to redirect to a non inertia page.

```html
<inertia-link href="/" external>External page</inertia-link>
```

### DIFF
--- a/packages/inertia-vue/src/link.js
+++ b/packages/inertia-vue/src/link.js
@@ -39,6 +39,10 @@ export default {
       type: Object,
       default: () => ({}),
     },
+    external: {
+      type: Boolean,
+      default: false,
+    }
   },
   render(h, { props, data, children }) {
     data.on = {
@@ -70,6 +74,10 @@ export default {
         ...data.on,
         click: event => {
           data.on.click(event)
+
+          if (external) {
+              window.location.href = url.href
+          }
 
           if (shouldIntercept(event)) {
             event.preventDefault()


reinink:
Hey @liamjoc! Thanks so much for your interest in Inertia.js, and for your sponsorship support as well!

So, I don't think this really makes sense to add. If you know that a link is going to be external, it makes more sense to just use a regular `<a>` anchor link, instead of an Inertia link. There is absolutely nothing wrong with mixing the two link types in the same app.

Hope that makes sense. 👍

liamjoc:
I understand it's very simple to use an `<a>` tag when you _know_ it's going to be external. But what if you don't? And in my case the latter is true.

I have created a component to switch between `<a>` and `<inertia-link>` but I still believe a simple `external` prop would be a nice addition to the link api instead of the following...

```html
<template>
    <div>
        <inertia-link
            v-if="!external"
            v-bind="$attrs"
            :href="href"
            role="menuitem"
        >
            <slot />
        </inertia-link>

        <a
            v-if="external"
            v-bind="$attrs"
            :href="href"
            role="menuitem"
        >
            <slot />
        </a>
    </div>
</template>
```

reinink:
Yeah, you make a decent argument. However, I *still* think it's better to do what you're doing in these situations. I think it's important that we keep it clear that Inertia.js requests are different than regular links. If your app needs this kind of a behaviour, I think a simple wrapper makes sense...or even a conditional, assuming you don't need to do this often.

claudiodekker:
Besides, how would you know when to apply the `external` tag? Wouldn't that imply knowing it's going to be external? ;)

Anyway, I just wanted to mention that another thing you could do in your custom component, is to consider 'absolute' links external and 'relative' urls local, by doing something like this: 
```js
isExternal() {
  return /^(http(s)?:\/\/)/.test(this.http);
}
```

PR #347, opened 2020-12-10
Transfrom props when handling popstate events

### LINKED ISSUE #343: router/navigate/history: props don't transform when go back or forward
For example App with this initialPage's props
```javascript
/*
    props = {
          isTransformed: false
    }
*/
   <App
    resolveComponent={name => require(`./Pages/${name}`).default}
    transformProps={props => {
      return {
        ...props,
        isTransformed: true,
      }
    }}
  />
```
Now page props
```javascript
/*
    props = {
          isTransformed: true
    }
*/
```

after navigate when I click go back or or forward browser's history
```javascript
/*
    props = {
          isTransformed: false
    }
*/
```

jorams:
The page is stored in the history state before transforming props. History state is serialized, so transforming it beforehand is not possible without losing information. Instead it needs to be transformed again after retrieval.

Fixes #343

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -337,6 +337,7 @@ export default {
   handlePopstateEvent(event) {
     if (event.state !== null) {
       const page = event.state
+      page.props = this.transformProps(page.props)
       let visitId = this.createVisitId()
       return Promise.resolve(this.resolveComponent(page.component)).then(component => {
         if (visitId === this.visitId) {


claudiodekker:
Hmm... While this issue definitely needs a fix, I don't think `popState` is the right place for this, as it is the result of a page that's already rendered. 

@reinink Would this be fixed by rearranging the following, or is there some reason for not doing that?
https://github.com/inertiajs/inertia/blob/a5d04b55d1f3656ca298989f6b031847bb69f73f/packages/inertia/src/inertia.js#L288-L291

jorams:
I don't think rearranging those lines fixes it, because the history state is serialized and thus doesn't preserve all aspects of the values you put into it. As a simple example:

```js
class A { constructor () {}}
const a = new A()
a.constructor.name === "A"

history.pushState(a, "")
history.state.constructor.name === "Object"
```

Of course transforming the props again is also not an ideal solution, if `transformProps` is not a pure function, but if it is the result will be identical.

weavdale:
I just ran into this problem today when displaying toast notifications. If I press browser back button then forward button, the toast notification reappears. Example:
https://youtu.be/r2hfTpz1oqA?t=776

amowu:
rollback to `"@inertiajs/inertia": "0.6.0"` is work for me, `0.6.1` have this problem.

saibotk:
> I don't think rearranging those lines fixes it, because the history state is serialized and thus doesn't preserve all aspects of the values you put into it. As a simple example:
> 
> ```js
> class A { constructor () {}}
> const a = new A()
> a.constructor.name === "A"
> 
> history.pushState(a, "")
> history.state.constructor.name === "Object"
> ```
> 
> Of course transforming the props again is also not an ideal solution, if `transformProps` is not a pure function, but if it is the result will be identical.

Is there any other serialization method that provides this functionality? Otherwise, we may use this patch or another solution if @claudiodekker has a better approach and document the drawbacks on the documentation. Because the current situation breaks the back button entirely.

saibotk:
Ok, this solution does also break on some cases and will cause inertia links to generate `[Vue warn]: Error in v-on handler: "DataCloneError: The object could not be cloned."` errors after pressing back and clicking on a link.

reinink:
Thanks for submitting this. This has been "fixed" by us removing `transformProps()` from Inertia. See #693 for more information. 👍

PR #351, opened 2020-12-13
Add a displayName to the PageContext

hailwood:
Presently in the react devtools we have this:
image

Context.Provider there being the pageContext we create as that is the default name,
but it's possible to name the contexts which makes navigating the react tree much nicer - After this pull request it will display as:
image

### DIFF
--- a/packages/inertia-react/src/PageContext.js
+++ b/packages/inertia-react/src/PageContext.js
@@ -1,3 +1,6 @@
 import { createContext } from 'react'
 
-export default createContext()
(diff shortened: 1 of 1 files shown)


reinink:
Interesting, we do [something similar](https://github.com/inertiajs/inertia/blob/master/packages/inertia-vue/src/app.js#L8) in Vue to make the devtools nicer to use.

Ping @sebastiandedeyne for review. 👍

sebastiandedeyne:
Perfect!

PR #357, opened 2020-12-19
Add first class support for validation errors

### LINKED ISSUE #356: Stop calling "onSuccess" when validation errors are present
Something that new users to Inertia keep bumping into is the fact that the `onSuccess` callback is called even when validation errors are present (#252, #325, #314, https://github.com/inertiajs/inertia-laravel/issues/172, https://github.com/inertiajs/inertia-laravel/issues/184). This is because, in an Inertia app, you never receive a `422` response. Instead, when you submit a form, if there are validation errors, you get redirected back, server-side, to the form page you're currently on, and the validation errors are then available in the `errors` prop. This redirect results in a `200` response on the form page, not a `422` response. **This is exactly like standard full page form submissions with server-side rendering.** Meaning, even if there are validation errors, Inertia still calls the `onSuccess` callback...because it was a successful response. There just happens to also be validation errors present in the props.

This is generally fine, since our forms reactively update based on the `errors` prop and display the validation errors. However, there are edge cases where you actually want to do something after a response has finished, in the event that no validation errors have occurred. For example, consider an update profile form, which has a password input. It's very common to want to clear the password form on a successful form submission. The currently recommended way to do this is via the `onSuccess` callback, checking to see if any errors are present.

```js
this.$inertia.post('/profile', this.form, {
  onSuccess: page => {
    if (Object.keys(page.props.errors).length === 0) {
      this.form.password = null
    }
  },
})
```

## A better API

And while this works (and technically makes sense), it's not the most ergonomic API. It would be nicer if Inertia could somehow detect if the response included validation errors, and only call the `onSuccess` callback if no errors are present. For example:

```js
this.$inertia.post('/profile', this.form, {
  onSuccess: () => this.form.password = null,
})
```

We could even add a new `onError` callback, should you want to do some type of form cleanup.

```js
this.$inertia.post('/profile', this.form, {
  onError: errors => { /* do something */ },
})
```

## Resolving errors

The hard part here is knowing when a response includes validation errors. Remember, there is no 422 response. One solution is to look for the presence of an `errors` prop. In Laravel, it's totally common to have a global `errors` prop (in fact, the Laravel adapter automatically sets this), however, other server-side frameworks may not work this way. To solve this, we would add a new `resolveErrors` callback, where you could set this for your own application.

```js
createApp({
  render: () => h(App, {
    resolveErrors: page => page.props.errors,
  })
}).use(plugin).mount(el)
```

In the event that the `errors` object that is returned from this function contains any errors, Inertia would consider it a "validation error response", and would call `onError` instead of `onSuccess`. This errors object would also be passed to the new `onError` callback.

The `resolveErrors` callback would be optional, and the default implementation would simply look for the root level `errors` prop.

## Global event

I think it would also make sense to add a global `error` event in addition to the `onError` callback. Unfortunately, we already have a global `error` event. I propose that we rename the current `error` event to `exception`, and then use the `error` event for validation errors.

This is probably a good change either way, since the current `error` only event fires on unexpected XHR errors, such as network interruptions, and for errors generated in the `resolveComponent()` callback. It does not fire for XHR requests that receive 400 and 500 level responses, or for non-Inertia responses, as these situations are handled in other ways by Inertia. Meaning, these are "exceptional" events.

## Link component

We will need to update the `<inertia-link>` components (all four adapters) add add the new `onError` callback.

reinink:
Closes #356

### DIFF
--- a/packages/inertia-react/src/App.js
+++ b/packages/inertia-react/src/App.js
@@ -2,14 +2,12 @@ import { Inertia } from '@inertiajs/inertia'
 import { createElement, useEffect, useState } from 'react'
 import PageContext from './PageContext'
 
-const identity = i => i
-
 export default function App({
   children,
   initialPage,
   resolveComponent,
   resolveErrors,
-  transformProps = identity,
+  transformProps,
 }) {
   const [current, setCurrent] = useState({
     component: null,
--- a/packages/inertia-react/src/App.js
+++ b/packages/inertia-react/src/App.js
@@ -8,6 +8,7 @@ export default function App({
   children,
   initialPage,
   resolveComponent,
+  resolveErrors,
   transformProps = identity,
 }) {
   const [current, setCurrent] = useState({
@@ -20,6 +21,7 @@ export default function App({
     Inertia.init({
       initialPage,
       resolveComponent,
+      resolveErrors,
       transformProps,
       swapComponent: async ({ component, page, preserveState }) => {
         setCurrent(current => ({
@@ -29,7 +31,7 @@ export default function App({
         }))
       },
     })
-  }, [initialPage, resolveComponent, transformProps])
+  }, [initialPage, resolveComponent, resolveErrors, transformProps])
 
   if (!current.component) {
     return createElement(PageContext.Provider, { value: current.page }, null)
--- a/packages/inertia-svelte/src/App.svelte
+++ b/packages/inertia-svelte/src/App.svelte
@@ -6,11 +6,13 @@
   export let
     initialPage,
     resolveComponent,
+    resolveErrors,
     transformProps = props => props
 
   Inertia.init({
     initialPage,
     resolveComponent,
+    resolveErrors,
     transformProps,
     swapComponent: async ({ component, page, preserveState }) => {
       store.update(current => ({
--- a/packages/inertia-svelte/src/App.svelte
+++ b/packages/inertia-svelte/src/App.svelte
@@ -7,7 +7,7 @@
     initialPage,
     resolveComponent,
     resolveErrors,
-    transformProps = props => props
+    transformProps
 
   Inertia.init({
     initialPage,
--- a/packages/inertia-vue/src/app.js
+++ b/packages/inertia-vue/src/app.js
@@ -15,6 +15,10 @@ export default {
       type: Function,
       required: true,
     },
+    resolveErrors: {
+      type: Function,
+      required: false,
+    },
     transformProps: {
       type: Function,
       default: props => props,
@@ -32,6 +36,7 @@ export default {
     Inertia.init({
       initialPage: this.initialPage,
       resolveComponent: this.resolveComponent,
+      resolveErrors: this.resolveErrors,
       transformProps: this.transformProps,
       swapComponent: async ({ component, page, preserveState }) => {
         this.component = component
--- a/packages/inertia-vue/src/app.js
+++ b/packages/inertia-vue/src/app.js
@@ -21,7 +21,7 @@ export default {
     },
     transformProps: {
       type: Function,
-      default: props => props,
+      required: false,
     },
   },
   data() {
--- a/packages/inertia-vue3/src/app.js
+++ b/packages/inertia-vue3/src/app.js
@@ -24,7 +24,7 @@ export default {
     },
     transformProps: {
       type: Function,
-      default: props => props,
+      required: false,
     },
   },
   setup({ initialPage, resolveComponent, transformProps, resolveErrors }) {
--- a/packages/inertia-vue3/src/app.js
+++ b/packages/inertia-vue3/src/app.js
@@ -18,15 +18,20 @@ export default {
       type: Function,
       required: true,
     },
+    resolveErrors: {
+      type: Function,
+      required: false,
+    },
     transformProps: {
       type: Function,
       default: props => props,
     },
   },
-  setup({ initialPage, resolveComponent, transformProps }) {
+  setup({ initialPage, resolveComponent, transformProps, resolveErrors }) {
     Inertia.init({
       initialPage,
       resolveComponent,
+      resolveErrors,
       transformProps,
       swapComponent: async (args) => {
         component.value = markRaw(args.component)
@@ -58,7 +63,7 @@ export default {
         return child
       }
     }
-  }
+  },
 }
 
 export const plugin = {
--- a/packages/inertia/src/events.js
+++ b/packages/inertia/src/events.js
@@ -8,8 +8,8 @@ export function fireBeforeEvent(visit) {
   return fireEvent('before', { cancelable: true, detail: { visit } } )
 }
 
-export function fireErrorEvent(error) {
-  return fireEvent('error', { cancelable: true, detail: { error } })
+export function fireExceptionEvent(exception) {
+  return fireEvent('exception', { cancelable: true, detail: { exception } })
 }
 
 export function fireFinishEvent(visit) {
--- a/packages/inertia/src/events.js
+++ b/packages/inertia/src/events.js
@@ -8,6 +8,10 @@ export function fireBeforeEvent(visit) {
   return fireEvent('before', { cancelable: true, detail: { visit } } )
 }
 
+export function fireErrorEvent(error) {
+  return fireEvent('error', { detail: { error } })
+}
+
 export function fireExceptionEvent(exception) {
   return fireEvent('exception', { cancelable: true, detail: { exception } })
 }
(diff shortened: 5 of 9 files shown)


packages/inertia-vue/src/app.js
```diff
@@ -15,6 +15,10 @@ export default {
       type: Function,
       required: true,
     },
+    resolveErrors: {
+      type: Function,
+      required: false,
```
reinink, line 20: So, I wonder if we should set the default here (instead of in Inertia core), simply because that's what we did with `transformProps`.

```suggestion
      default: page => (page.props.errors || {}),
```

packages/inertia-vue3/src/app.js
```diff
@@ -18,15 +18,20 @@ export default {
       type: Function,
       required: true,
     },
+    resolveErrors: {
+      type: Function,
+      required: false,
```
reinink, line 23:
```suggestion
      default: page => (page.props.errors || {}),
```

packages/inertia/src/events.js
```diff
@@ -9,7 +9,11 @@ export function fireBeforeEvent(visit) {
 }
 
 export function fireErrorEvent(error) {
```
reinink, line 11: I think plural here makes more sense.
```suggestion
export function fireErrorEvent(errors) {
```

packages/inertia/src/events.js
```diff
@@ -9,7 +9,11 @@ export function fireBeforeEvent(visit) {
...
-  return fireEvent('error', { cancelable: true, detail: { error } })
+  return fireEvent('error', { detail: { error } })
```
reinink, line 12:
```suggestion
  return fireEvent('error', { detail: { errors } })
```

packages/inertia-react/src/App.js
```diff
@@ -8,6 +8,7 @@ export default function App({
   children,
   initialPage,
   resolveComponent,
+  resolveErrors,
```
reinink, line 11:
```suggestion
  resolveErrors = resolveErrors,
```

Will also need to add this method definition about, like `transformProps`.

```js
const resolveErrors = page => (page.props.errors || {})
```

I'd also love to rename the `transformProps` callback from `identity` to `transformProps`. I really don't understand the name `identity` at all, and now we have two callbacks, so we really need better naming. Maybe @sebastiandedeyne can provide some insight here.

packages/inertia-svelte/src/App.svelte
```diff
@@ -6,11 +6,13 @@
   export let
     initialPage,
     resolveComponent,
+    resolveErrors,
```
reinink, line 9:
```suggestion
    resolveErrors = page => (page.props.errors || {}),
```

packages/inertia/src/inertia.js
```diff
@@ -1,19 +1,31 @@
...
+  fireBeforeEvent,
+  fireErrorEvent,
+  fireExceptionEvent,
+  fireFinishEvent,
+  fireInvalidEvent,
+  fireNavigateEvent,
+  fireProgressEvent,
+  fireStartEvent,
+  fireSuccessEvent,
+} from './events'
```
reinink, line 14: Can we keep this as a one-line import? Even if we have to do `// prettier-ignore` at the end of the line. :feelsgood:

packages/inertia/src/inertia.js
```diff
@@ -1,19 +1,31 @@
...
 import { hrefToUrl, mergeDataIntoQueryString, urlWithoutHash } from './url'
 
 export default {
   resolveComponent: null,
+  resolveErrors: page => (page.props.errors || {}),
```
reinink, line 19: Since the default will be set in the adapters.
```suggestion
  resolveErrors: null,
```

packages/inertia/src/inertia.js
```diff
@@ -1,19 +1,31 @@
...
   swapComponent: null,
   transformProps: null,
   activeVisit: null,
   visitId: null,
   page: null,
 
-  init({ initialPage, resolveComponent, swapComponent, transformProps }) {
+  init({ initialPage, resolveComponent, resolveErrors, swapComponent, transformProps }) {
     this.resolveComponent = resolveComponent
+    this.resolveErrors = resolveErrors || this.resolveErrors
```
reinink, line 28:
```suggestion
    this.resolveErrors = resolveErrors
```

packages/inertia/src/inertia.js
```diff
@@ -236,6 +249,11 @@ export default {
         }
         return this.setPage(response.data, { visitId, replace, preserveScroll, preserveState })
       }).then(() => {
+        const errors = this.resolveErrors(this.page)
+        if (Object.keys(errors).length > 0) {
+          fireErrorEvent(errors)
+          return onError(errors)
+        }
```
ajnsn, line 256: What if `resolveErrors` does not return an object?

claudiodekker: Then that should lead to a fix in the user's `resolveErrors` implementation 😅

reinink: Yeah, exactly. It's basically on the developer to ensure the `resolveErrors()` method always returns an object. And I fully expect most Laravel/Rails developers to never have to set this, and can simply reply on the default behaviour (which an empty object if no errors are present).

ajnsn:
Can we talk one more time about the naming/placement of this new event?  According to the [Documentation](https://inertiajs.com/events) `error` fires on unexpected XHR errors. `invalid` fires for every non-Inertia response.
Now we have this great new validation error approach is inside the `error` event.

Every now and then I am already finding it difficult to find the right event to listen to or at least listen to mulitple events to achieve a proper server error handling, e.g. on `4xx` and `5xx` responses. Not only that `200` responses are also in the `invalid` callback, currently I need to also listen to `error` event in case some network issues occur. 

When the validation errors are now passed to the `error` event, maybe we can also have all other servers (`4xx`, `5xx`) errors also there?

claudiodekker:
Hi @ajnsn,

While the event system opened up a lot of new opportunities, we too noticed that not only did the naming of certain events make things confusing to some people, it also became confusing to us after a while. This is in part why we're now making this change (#356)

As you're correctly pointing out, the `error` event indeed fires on unexpected XHR errors such as a network disconnect, however, we consider these to be more rare situations, things that are really _exceptional_ and don't have anything to do with you having an error in your code or such. For this reason, this PR changes the `error` event to `exception`.

If we consider this change alone, there's now three 'main' response states:
- `success`: When a valid Inertia response comes back.
- `invalid`: When a response comes back, that isn't generated by Inertia's server-side counterpart.
- `exception`: When something exceptional happens, such as a network disconnect.

While this already improves things, it still confuses people, because intuitively the `success` event tells you the request succeeded. While this is also true from Inertia's perspective, it still means a different thing entirely:
- To developers, `success` means the request succeeded. For example, a form was submitted, no validation errors occurred, and whatever change on the back-end was processed.
- To Inertia, `success` means that whatever response came back was a _valid_ Inertia response. A `4xx` or `5xx` error generated by Inertia's back-end counterpart, is still a `success` response.

However, because validation errors are so common and are almost always described by an error in the `errors` prop (unless you modified it), we've decided to now re-introduce the `error` event, but this time to trigger when there's validation errors, making for four possible response states:

- `success`: When a valid Inertia response comes back, without any (validation) errors.
- `error`: When a valid Inertia response comes back, but with (validation) errors.
- `invalid`: When a response comes back, that isn't generated by Inertia's server-side counterpart.
- `exception`: When something exceptional happens, such as a network disconnect.

We believe that these changes should both make the events easier to understand, as well as improve the developer experience. I hope this helps!

ajnsn:
Hey @claudiodekker 

Thank you so much for your detailed and informative response. 
I did not recognize that you also changed the `error` event. The `exception` totally makes sense and yes, it should be very rare.

However the part about the perspectives is interesting. I do believe that you will keep answering a lot of confused developer questions if you keep this "how Inertia sees it" approach. While this is technical correct and also internally recommended, I think that a developer faced "framework" should be intuitive to the developer.

For example if you do not want the [recommend dedicated error component ](https://inertiajs.com/error-handling) and use a globally registered notification component for errors, you might end up with the following code:

```js
Inertia.on('invalid', event => {
    if (status !== 200) {
        event.preventDefault(); // Do not display Inertia's modal
        dispatchNotification(event.detail);
    }
});
```

For me this is totally counterintuitive. I am checking an `invalid` event and exclude a successful response.
A developer orientated approach might be:

- `success`: When a valid Inertia response comes back.
- `error`: When a response comes back, that isn't generated by Inertia's server-side counterpart or does not have a 2xx status.
- `invalid`: When a valid Inertia response comes back, but with validation errors.
- `exception`: When something exceptional happens, such as a network disconnect.

This would mean that `error` is always an error, means something went wrong. 
The `invalid` turns into the validation failed event. For me this is easily understandable `invalid` is typically something not in the correct format etc.

Finally, you can proceed with the above explained approach. It's totally fine. You just need to get your head around it and see it how Inertia sees it.

PR #363, opened 2020-12-21
Add Inertia Form to Vue 2 adapter

claudiodekker:
This PR brings a first-class form helper to InertiaJS, as part of the Inertia Plugin (`this.$inertia.form`). The goal here is to integrate Jetstream's form helper and to keep the signature as similar as possible to make migration easy, while also improving the developer experience at the same time.

As an example, let's imagine we're building an Inertia page where you can edit the user's details using an Inertia form. 
If you've used Inertia with Laravel Jetstream before, this will probably look very familiar:
```html
<template>
  ...
  <input v-model="form.email" type="text" placeholder="foo@example.com" />
  <input v-model="form.password" type="password" placeholder="Password" />
  ...
</template>

<script>
props: {
  email: String, // The user's existing email, passed as an Inertia prop
},
data: () => {
  return {
    form: this.$inertia.form({
      email: this.email, // Assigning the form's default values
      password: ''
    })
  }
),
methods: {
  submit() {
    this.form.post(this.$inertia.page.url)
  }
}
</script>
```

This by itself will create a POST request to the current page, submitting the form itself. 

## Resetting fields (to their initial values)

Now, let's imagine a requirement to enter the user's current `password` as an extra layer of validation before changing the user details. In normal situations, a successful update would send you back to the same page, clearing out the `password` field, but this wouldn't happen when an invalid password is given and validation errors are displayed, meaning we have to manually reset the password field somehow. 

With the power of events and the Inertia form, this task becomes very simple:

```js
submit() {
  this.form.post(this.$inertia.page.url, {
    onError() {
      this.form.reset('password')
    }
  })
}
```

The above ensures that only the `password` field gets reset back to it's initial value, as define in when the form was initialized (in the `data` section of the page component). Perhaps the `password` field isn't the best of examples here, because it's default value is an empty string instead of an actual starting value, but you get the idea.

#### Usage
- `form.reset()`: Resets the entire form back to it's initial values
- `form.reset('foo')` Only resets the 'foo' field back to it's initial value
- `form.reset(['foo', 'bar'])`: Resets only the 'foo' and 'bar' fields back to their initial values
- `form.reset('foo', 'bar')`: Same as array-based entry, but using a fluent list of arguments.

## Errors

Form errors work very similar to the page-wide `errors` prop, but are scoped to the form itself, meaning they'll only show up or get updated when you've submitted that specific form. Usage is as you'd expect:

```html
<template>
  ...
  <input v-model="form.email" type="text" placeholder="foo@example.com" id="email" />
  <label v-if=form.errors.email for="email">{{ form.errors.email }}</label>
  ...
</template>
```

## Processing state
If you want to display a custom loading indicator while the form is being submitted, this is also very easy to do, because each Inertia form instance also contains a `processing` prop that will be automatically updated as the request gets processed:

```html
<span v-if="form.processing">Please wait.. the form is being submitted.</span>
```

## Methods

So far, we've only used the `post` method in our example, but other methods are also available:

- `form.post()`
- `form.put()`
- `form.patch()`
- `form.delete()`

Lastly, something that is nice to know, is that if you don't provide any arguments (like in the signatures listed here), the form will submit the form using the requested method toward the current page (`$inertia.page.url`) with the default Inertia options for that request method..

### DIFF
--- a/packages/inertia-vue/src/app.js
+++ b/packages/inertia-vue/src/app.js
@@ -1,5 +1,5 @@
-import link from './link'
 import form from './form'
+import link from './link'
 import remember from './remember'
 import { Inertia } from '@inertiajs/inertia'
 
--- a/packages/inertia-vue/src/app.js
+++ b/packages/inertia-vue/src/app.js
@@ -1,4 +1,5 @@
 import link from './link'
+import form from './form'
 import remember from './remember'
 import { Inertia } from '@inertiajs/inertia'
 
@@ -78,6 +79,7 @@ export default {
 export const plugin = {
   install(Vue) {
     Object.defineProperty(Vue.prototype, '$inertia', { get: () => Inertia })
+    Object.defineProperty(Vue.prototype.$inertia, 'form', { get: () => form })
     Object.defineProperty(Vue.prototype, '$page', { get: () => app.page })
     Vue.mixin(remember)
     Vue.component('InertiaLink', link)
--- a/packages/inertia-vue/src/form.js
+++ b/packages/inertia-vue/src/form.js
@@ -77,9 +77,6 @@ export default data => {
 
       return this
     },
-    hasErrors() {
-      return Object.keys(form.errors).length > 0
-    },
   }
 
   return new Proxy(form, {
--- a/packages/inertia-vue/src/form.js
+++ b/packages/inertia-vue/src/form.js
@@ -66,7 +66,13 @@ export default function(data) {
             return options.onStart(visit)
           }
         },
-        onProgress: progress => (this.progress = progress),
+        onProgress: event => {
+          this.progress = event
+
+          if (options.onProgress) {
+            return options.onProgress(event)
+          }
+        },
         onSuccess: page => {
           this.clearErrors()
 
--- a/packages/inertia-vue/src/form.js
+++ b/packages/inertia-vue/src/form.js
@@ -1,10 +1,10 @@
 import Vue from 'vue'
 import { Inertia } from '@inertiajs/inertia'
 
-export default initialData => {
+export default data => {
   const form = {
-    data: { ... initialData },
-    defaults: { ... initialData },
+    data: { ... data },
+    defaults: { ... data },
     errors: {},
     processing: false,
     submit(method, url, options) {
--- a/packages/inertia-vue/src/form.js
+++ b/packages/inertia-vue/src/form.js
@@ -11,7 +11,7 @@ export default initialData => {
       url = url || Inertia.page.url
       options = options || {}
 
-      const requestOptions = {
+      Inertia[method](url, form.data, {
         ... options,
         onStart: visit => {
           form.processing = true
@@ -41,9 +41,7 @@ export default initialData => {
             return options.onSuccess(page)
           }
         },
-      }
-
-      Inertia[method](url, form.data, requestOptions)
+      })
     },
     post(url, options) {
       form.submit('post', url, options)
--- a/packages/inertia-vue/src/form.js
+++ b/packages/inertia-vue/src/form.js
@@ -19,18 +19,16 @@ export default function(data) {
         }, {})
     },
     reset(fields) {
-      const args = Array.isArray(fields)
-        ? fields
-        : Object.values(arguments)
+      fields = Array.isArray(fields) ? fields : Object.values(arguments)
 
-      if (args.length === 0) {
+      if (fields.length === 0) {
         Object.assign(this, defaults)
       } else {
         Object.assign(
           this,
           Object
             .keys(defaults)
-            .filter(key => args.includes(key))
+            .filter(key => fields.includes(key))
             .reduce((carry, key) => {
               carry[key] = defaults[key]
               return carry
--- a/packages/inertia-vue/src/form.js
+++ b/packages/inertia-vue/src/form.js
@@ -40,7 +40,8 @@ export default function(data) {
       return this.clearErrors()
     },
     clearErrors() {
-      this.errors = {}
+      Vue.set(this, 'errors', {})
+      this.hasErrors = false
 
       return this
     },
@@ -80,8 +81,7 @@ export default function(data) {
           }
         },
         onSuccess: page => {
-          Vue.set(this, 'errors', {})
-          this.hasErrors = false
+          this.clearErrors()
 
           if (options.onSuccess) {
             return options.onSuccess(page)
--- a/packages/inertia-vue/src/form.js
+++ b/packages/inertia-vue/src/form.js
@@ -63,12 +63,15 @@ export default data => {
       if (args.length === 0) {
         form.data = { ... form.defaults }
       } else {
-        Object.assign(form.data, Object.keys(form.defaults)
-          .filter(key => args.includes(key))
-          .reduce((carry, key) => {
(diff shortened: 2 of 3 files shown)


irakan:
Would be awesome if the form can also be used for a none inertia request ( ordinary ajax requests )

claudiodekker:
@irakan I'm sure something like that already exists somewhere ([like this one from Spatie](https://github.com/spatie/form-backend-validation/blob/master/src/Form.js)), but you could of course always create your own NPM package for that! 👍

packages/inertia-vue/src/form.js
```diff
@@ -0,0 +1,99 @@
...
+      if (args.length === 0) {
+        form.data = { ... form.defaults }
+      } else {
+        Object.assign(form.data, Object.keys(form.defaults)
+          .filter(key => args.includes(key))
+          .reduce((carry, key) => {
+            carry[key] = form.defaults[key]
+            return carry
+          }, {})
+        )
```
pxlrbt, line 72: I think readability could be improved a bit if both arguments are on a separate line. Otherwise it's a bit confusing what `.filter` and `.reduce` are working on.

```js
Object.assign(
    form.data,
    Object.keys(form.defaults)
        .filter(key => args.includes(key))
        .reduce((carry, key) => {
            carry[key] = form.defaults[key]
            return carry
        }, {})
)
```

Or even:
```js
Object.assign(
    form.data,
    Object
        .keys(form.defaults)
        .filter(key => args.includes(key))
        .reduce((carry, key) => {
            carry[key] = form.defaults[key]
            return carry
        }, {})
)
```

packages/inertia-vue/src/form.js
```diff
@@ -77,9 +77,6 @@ export default data => {
 
       return this
     },
-    hasErrors() {
-      return Object.keys(form.errors).length > 0
-    },
```
reinink: Okay, so, I get that we're not using this, but wouldn't this be a super handy method for userland code? I hate having to do this:

```js
Object.keys(form.errors).length > 0
```

Much better to do:

```js
form.hasErrors()
```

claudiodekker: Yeah, I already re-added it after talking it over with @pxlrbt 😅 Just didn't push it up yet because lunch & the "remember" problem I'm working on ;)

packages/inertia-vue/src/form.js
```diff
@@ -0,0 +1,99 @@
...
+    put(url, options) {
+      form.submit('put', url, options)
+    },
+    patch(url, options) {
+      form.submit('patch', url, options)
+    },
+    delete(url, options) {
+      form.submit('delete', url, options)
+    },
+    reset(fields) {
```
reinink, line 58: When resetting the form, should we not also reset the errors to an empty object? It seems to me like if I'm resetting the form, I also want to clear any previous errors as well.

In fact, I'd even suggest having a `form.clearErrors()` method, just in case you want to manually clear the errors.

packages/inertia-vue/src/form.js
reinink: Haha, understood. 👍

packages/inertia-vue/src/form.js
pxlrbt: > In fact, I'd even suggest having a form.clearErrors() method, just in case you want to manually clear the errors.
+1 for the `form.clearErrors()`. No concrete use case right now, but seems nice to have.

> It seems to me like if I'm resetting the form, I also want to clear any previous errors as well.
Use cases I can think of:
1. Creating stuff and resetting the form and errors afterwards to add another entry. But in this case the errors are already reset by the `onSuccess` handler.
2. User wants to reset the form. So it would make sense to reset the errors aswell, as they are no longer valid.

Anything else?

claudiodekker: Yeah, so, while I'm not opposed to this, I'd really like to know of some _real_ use cases first.

1. With Inertia this would already be reset because of the Inertia visit that happens in the background, so it's really not something you would do
2. Would you really do this one? So far with pretty much every webpage I've seen ever, this doesn't happen, and I'm really having a hard time coming up with a real scenario for this?

reinink: So the use-case I'm thinking is the modal form in my app. When you close the modal, and then open it again, it would be nice to call `form.reset()`, and have all the original form values reset and the errors cleared. 👍

claudiodekker: Yep, okay, adding it in 👍

packages/inertia-vue/src/form.js
```diff
@@ -0,0 +1,108 @@
...
+              carry[key] = defaults[key]
+              return carry
+            }, {}),
+        )
+      }
+
+      return this.clearErrors()
+    },
+    clearErrors() {
+      this.errors = {}
```
pxlrbt, line 92: I think this should reset `hasErrors`.

```suggestion
      this.errors = {}
      this.hasErrors = false
```

Not sure whether this needs to use `Vue.set` (You used it in `onSuccess`)
```suggestion
      Vue.set(this, 'errors', {})
      this.hasErrors = false
```

packages/inertia-vue/src/form.js
```diff
@@ -0,0 +1,108 @@
...
+          Vue.set(this, 'errors', errors)
+          this.hasErrors = true
+
+          if (options.onError) {
+            return options.onError(errors)
+          }
+        },
+        onSuccess: page => {
+          Vue.set(this, 'errors', {})
+          this.hasErrors = false
```
pxlrbt, line 49: This can use the `clearErrors` function.

```suggestion
          this.clearErrors()
```

claudiodekker: Good catch! 👍

packages/inertia-vue/src/app.js
```diff
@@ -1,4 +1,5 @@
 import link from './link'
+import form from './form'
```
reinink, line 2:
```suggestion
import form from './form'
import link from './link'
```

packages/inertia-vue/src/form.js
```diff
@@ -0,0 +1,114 @@
...
+        .keys(data)
+        .reduce((carry, key) => {
+          carry[key] = this[key]
+          return carry
+        }, {})
+    },
+    reset(fields) {
+      const args = Array.isArray(fields)
+        ? fields
+        : Object.values(arguments)
```
reinink, line 24: Instead of using the yucky `args` variable here, couldn't we just reassign `fields`?

```suggestion
      fields = Array.isArray(fields) ? fields : Object.values(arguments)
```

packages/inertia-vue/src/form.js
```diff
@@ -0,0 +1,114 @@
...
+
+      if (args.length === 0) {
```
reinink, line 26:
```suggestion
      if (fields.length === 0) {
```

packages/inertia-vue/src/form.js
```diff
@@ -0,0 +1,114 @@
...
+        Object.assign(this, defaults)
+      } else {
+        Object.assign(
+          this,
+          Object
+            .keys(defaults)
+            .filter(key => args.includes(key))
```
reinink, line 33:
```suggestion
            .filter(key => fields.includes(key))
```

PR #365, opened 2020-12-23
Add Vue 3 form helper

reinink:
This PR adds the same form helper that's available in the Vue 2 adapter to the Vue 3 adapter. It is literally identical. However, since we don't want this logic in core, we'll leave it duplicated for now.

Here's how to use it:

```vue
<template>
  <form @submit.prevent="form.post('/login')">
    <input type="text" v-model="form.email" :error="form.errors.email" />
    <input type="password" v-model="form.password" :error="form.errors.password"/>
    <button type="submit" :disabled="form.processing">Login</button>
  </form>
</template>

<script>
export default {
  data() {
    return {
      form: this.$inertia.form({
        email: null,
        password: null,
      }),
    }
  },
}
</script>
```

It also comes with a `form.reset()` helper for resetting the form back to its original state. Or, if you want to only reset specific fields, you can call `form.reset('fieldname', 'anotherfieldname')`.

Also, in the event that you're uploading files, the current progress event is available at `form.progress`, which makes it really easy to show upload progress in your views:

```vue
<progress :value="form.progress.percentage" max="100">{{ form.progress.percentage }}%</progress>
```

And, of course, you can also use the form helper with the new composition API:

```js
import { useForm } from '@inertiajs/inertia-vue3'

export default {
  setup () {
    const form = useForm({
      email: 'johndoe@example.com',
      password: 'secret',
      remember: null,
    })

    return { form }
  },
}
````

### DIFF
--- a/packages/inertia-vue3/src/app.js
+++ b/packages/inertia-vue3/src/app.js
@@ -1,3 +1,4 @@
+import form from './form'
 import link from './link'
 import remember from './remember'
 import { computed, h, markRaw, ref } from 'vue'
@@ -68,6 +69,7 @@ export default {
 
 export const plugin = {
   install(app) {
+    Inertia.form = form
     Object.defineProperty(app.config.globalProperties, '$inertia', { get: () => Inertia })
     Object.defineProperty(app.config.globalProperties, '$page', { get: () => page.value })
     app.mixin(remember)
--- a/packages/inertia-vue3/src/app.js
+++ b/packages/inertia-vue3/src/app.js
@@ -85,3 +85,7 @@ export function usePage() {
     version: computed(() => page.value.version),
   }
 }
+
+export function useForm(data) {
+  return ref(form(data))
+}
--- a/packages/inertia-vue3/src/form.js
+++ b/packages/inertia-vue3/src/form.js
@@ -1,0 +1,109 @@
+import { Inertia } from '@inertiajs/inertia'
+
+export default function(data) {
+  const defaults = JSON.parse(JSON.stringify(data))
+
+  return {
+    ...defaults,
+    errors: {},
+    hasErrors: false,
+    processing: false,
+    progress: null,
+    data() {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by innocenzi** [comment]
Would you be against a hook instead of the example you're showing in the second code block? 

```js
import { useForm } from '@inertiajs/inertia'

export default {
  setup () {
    const form = useForm({
      email: 'johndoe@example.com',
      password: 'secret',
      remember: null,
    })

    return { form }
  },
}
```

Simple syntactical sugar, but way more readable. Would look like that in `app.js`:

```js
import { ref } from 'vue'

export function useForm(data) {
  return ref(form(data))
}
```

That being said, I think the `form` object should work with reactive variables instead, but that would require a little refactor, and I don't think it would change that much.

reinink:
@innocenzi Ahh yes! I was totally planning on adding that. You want to dig in that?

innocenzi:
> You want to dig in that?

Sure, I'll fork and play with it a bit 👍

reinink:
@innocenzi I'm hoping to release ASAP, so I'm going to try and tackle it myself. 👍

innocenzi:
Sounds good, I'll PR later if I find stuff I can improve 👍

reinink:
@innocenzi Done: 02c138a25946034c82ab9905652efb726b70eeb9

innocenzi:
Played with it and I don't think this needs a refactor, wrapping with `ref` is good enough and I don't really see a drawback except for the readability of the code of the adapter, which will be improved with the upcoming TypeScript refactor anyway.

This new form helper is super nice, really good work on that API. 😊

reinink:
@innocenzi Hey, amazing, thanks for reviewing buddy, and thanks for the suggestion! Can't believe I almost forgot it.

As for the API, I agree, it turned out really nice. Mostly the work of Taylor and Spatie over the years. ❤️

PR #395, opened 2021-01-04
Add Cypress-based test suite

claudiodekker:
This PR adds Inertia tests to cover all documented Inertia features in combination with the Vue adapter. 

### Goals
- Catch breaking issues before we release them
- Test changes across multiple browsers
- Make the upcoming TypeScript refactor more straight forward
- Enable developers to open issues with a failing test to simplify reproduction and fixing.

### TODO
- Changes in https://github.com/inertiajs/inertia/releases/tag/v0.8.3 & https://github.com/inertiajs/inertia/releases/tag/inertia-vue%40v0.5.5

### Things that require some help or feedback:
- My testing style. Both Jonathan and I aren't all that familiar with Javascript-based testing, so please let me know if I'm doing something wrong here.
- **Some of the tests are currently marked as _skipped_, simply because I couldn't get them working correctly or reliably**, and while this kind sucks, it's still better to at least have them defined than to not have them at all (as otherwise it would feel as if the tests were incomplete/not covering everything). I would REALLY appreciate it if someone more experienced with JS-based testing could help out with fixing these.


reinink:
🕺 🎉💥

THANKS @claudiodekker!!!!!!!!!!!!!!!!!!!!!

PR #398, opened 2021-01-04
Use remember history state when visiting the same component and preserving state

reinink:
This PR fixes an edge case where the remember history state is lost when making a visit to the same page component and while preserving state.

For example, if you're submitting a form, and you end up back on the same page (ie. to show errors), we want to make sure that all the remember data is preserved. Previously, since a visit happened, it was lost.

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -244,6 +244,9 @@ export default {
         if (only.length && response.data.component === this.page.component) {
           response.data.props = { ...this.page.props, ...response.data.props }
         }
+        if (preserveState && window.history.state?.rememberedState && response.data.component === this.page.component) {
+          response.data.rememberedState = window.history.state.rememberedState
+        }
         const responseUrl = hrefToUrl(response.data.url)
         if (url.hash && !responseUrl.hash && urlWithoutHash(url).href === responseUrl.href) {
           responseUrl.hash = url.hash


sebastiandedeyne:
Does this solve the same problem as https://github.com/inertiajs/inertia/pull/331? (ping @AdrianMrn)

AdrianMrn:
I think so!

reinink:
@AdrianMrn I'm going to merge this in and publish it today. If you could test to see if this solves your issue, that would be much appreciated!

AdrianMrn:
Will do!

AdrianMrn:
@reinink Sorry for taking so long! It seems like this PR doesn't fix my issue. I believe my fix might also handle the edge case you fixed with this PR.

reinink:
So, I literally *just* released this a few minutes ago. Are you on the latest versions?

AdrianMrn:
image
image

Yeah :/

AdrianMrn:
If this makes it any easier for you to reproduce, this is the code I'm using:
```tsx
const newUrl = new URL(url);
newUrl.searchParams.delete('page');

if (newUrl.toString() !== window.location.href) {
    window.history.replaceState(
        { ...window.history.state, url: newUrl.toString() },
        '',
        newUrl.toString()
    );
}
```

AdrianMrn:
While typing out the above comment, I realized I could call the function that updates my querystring **after** calling the remembered state setter. This is a workaround for my issue, but I still feel like my PR might be a suitable solution in case other people run into this (unlikely).

PR #411, opened 2021-01-05
Debounce remember method

sebastiandedeyne:
Safari has this bonkers limitation of limiting `pushState` and `replaceState` calls to 100 times per 30 seconds. That's 3 calls per second. Clicking around an app won't cause this, but binding `Inertia.remember` to a textfield can. The problem here is that once Safari throws the error, all navigation breaks.

Screenshot 2021-01-05 at 18 35 35

To work around this issue, we could add a debounce to the `remember` method at about 400ms. 333ms would work, but better to have some leeway.

Unable to test this atm, but this is my idea on how we can debounce the `remember` method without interfering with other `pushState`/`replaceState` calls.

I'm not sure about the code, the `debounce` placement is ugly, and I don't like the `cancelRemember` switch, but we can at least discuss and play around with this PR.

### DIFF
--- a/packages/inertia/src/debounce.js
+++ b/packages/inertia/src/debounce.js
@@ -1,7 +1,11 @@
 export default function debounce(fn, delay) {
   let timeoutID = null
-  return function () {
+  let debouncedFn = function () {
     clearTimeout(timeoutID)
     timeoutID = setTimeout(() => fn.apply(this, arguments), delay)
   }
+  debouncedFn.cancel = function () {
+    clearTimeout(timeoutID)
+  }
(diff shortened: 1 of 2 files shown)


claudiodekker:
Ridiculous, @WebKit..

Perhaps there's some Safari check we can do, to prevent handicapping Chrome, Edge, Firefox etc.? Clicking a link/button/whatever [can easily be done in under 150ms](https://instant.page/), not even to mention situations in which you really put your mind to it (eg. a web-game like cookie clicker)

reinink:

PR #435, opened 2021-01-15
Provide $inertia and $page so they can be injected

grantholle:
To take advantage of the new [provide/inject](https://v3.vuejs.org/guide/component-provide-inject.html) dependency system, this provides the `$inertia` and `$page` properties so they can be injected in composition functions.

```js
import { defineComponent, inject } from 'vue'

export default defineComponent({
  setup () {
    const $page = inject('$page')
    const user = $page.props.user

    return {
      user,
    }
  },
})
```


reinink:
Hey @grantholle, thanks for this PR. Have you tested this to make sure the injected `$page` variable in your component is actually reactive? I feel like I tried this in Vue 2 (which also had provide/inject), and while it got the initial page props, they were not reactive.

grantholle:
Ok, great point.

After some testing, the way this would work in a component's `setup()` function would be to provide the page reactive variable directly and then in the component to wrap it in a `computed` call.

```js
import { defineComponent, inject, computed } from 'vue'

export default defineComponent({
  setup () {
    const $page = inject('$page')
    const user = computed(() => $page.value.props.user)

    return {
      user,
    }
  }
})
```

Which, once I thought about, makes sense.

Is it worth doing it this way over using the simple `usePage`? Not sure. It does seem more "vue-y" to use `provide` but also is quite verbose.

Up to you on adding it. Thanks for taking a look.

grantholle:
Also the above example is if you want to have the page prop be reactive. This example is reactive without any of the computed stuff.

```vue
<template>
  <div>
    {{ $page.props.user.name }}
  </div>
</template>

<script>
import { defineComponent, inject } from 'vue'

export default defineComponent({
  setup () {
    const $page = inject('$page')

    return {
      $page,
    }
  }
})
</script>
```

reinink:
@grantholle Oh interesting, thanks for digging into this. I guess the question is, does using provide/inject add anything over just use our `usePage()` hook?

grantholle:
I think from the technical standpoint of getting the page props to any component, _no_.

My two cents is that I personally like the provide/inject system. Adding a couple lines seems worth it to me for the flexibility. I also see the risk of increasing the maintenance burden, however small it is...

Up to you Jonathan! 😀

freakzlike:
I like to use `provide/inject` because this makes it easy to mock for unit testing.

Currently to mock the page object during unit test when a component uses `usePage`, you have to do something like this, which is very limited in handling:
```js
jest.mock('@inertiajs/inertia-vue3', () => ({
  usePage: () => ({
    props: { value: { auth: null } }
  })
}))
```

With `provide/inject` this would be possible by using `vue-test-utils` `mount`:
```js
const wrapper = mount(Component, {
  global: {
    provide: {
      $page: { value: { auth: null } }
    }
  }
})
```

But I think `provide/inject` should be used by `usePage` directly. I have often seen this by other libraries. For example in `vue-router`, `vuex` and `quasar`:
See [https://github.com/vuejs/vue-router-next/blob/master/src/useApi.ts](https://github.com/vuejs/vue-router-next/blob/master/src/useApi.ts)
```typescript
export function useRouter(): Router {
  return inject(routerKey)!
}
```

During plugin installation of `vue-router`:
```js
// [...]
app.config.globalProperties.$router = router
app.provide(routerKey, router)
// [...]
```

Tofandel:
I think this can be closed as we now have `usePage` and `useInertia` which solves this issue

grantholle:
Yes, I agree. Still like `inject` a bit better in this instance 🤭

Issue #394, opened 2021-01-04
svelte: Class constructor Link cannot be invoked without 'new'


christoomey:
### Versions:

- `@inertiajs/inertia` version: 0.8.0
- `@inertiajs/inertia-svelte` version: 0.6.0
- `@babel/core` version: 7.11.1
- `webpack` version: 4.44.2
- `@rails/webpacker` version: 5.2.1

### Describe the problem:

While trying to use the `use:inertia` action on an `<a>` tag I run into the following error:

```
Uncaught (in promise) TypeError: Class constructor Link cannot be invoked without 'new'
```

This error seems to be related to the code in the library using non-es5 features. I did a quick check with:

```
❯ npx are-you-es5 check -r .
npx: installed 10 in 2.303s
❌ @inertiajs/inertia-svelte is not ES5
❌ svelte-preprocess is not ES5
❌ tailwindcss is not ES5
```

**Note** - If I copy the current `inertia-svelte/src/link.js` into my app then things work fine, which leads me to believe that the issue is related to the use of non-es5 syntax, and node_modules not being transpiled.

That said, it's not clear to me what exactly in the inertia-svelte Link code that would be causing this, nor why this isn't a problem for other packages which certainly seem to be using non-es5 syntax.

I've tried a number of babel & webpack tweaks to try to get them to transpile the inertia-svelte code but thus far no luck. That's likely operator error on my part, but more generally I think it would be great to not have to do any of that.

### Steps to reproduce:

In my case I'm using webpack via rails webpacker gem.

1. Add `import { inertia } from '@inertiajs/inertia-svelte'`
2. Apply the inertia action to a link `<a use:inertia ...`
3. See the `Uncaught (in promise) TypeError: Class constructor Link cannot be invoked without 'new'` error

Currently I'm able to get by either using the `InertiaLink` component or by copying the action code into my repo.

reinink:
I wonder if this is related to #414 at all. 🤔

christoomey:
Oh, interesting! Looking at the inertia-svelte package I can see it has no build process currently, so projects are running directly against the `src/*` files and thus hitting the error described above. Previously I couldn't see why the `@inertiajs/inertia` package would be different, but looking again I see that you've got a build step configured there, and projects are running against a bundled and transpiled `dist/*` file.

[...]

PR #439, opened 2021-01-15
WIP Add build step for inertia-svelte

christoomey:
This change represents a first step at adding a building step for the
inertia-svelte package, specifically hoping to address #394.

---

I've introduced rollup as that seems to be the common practice in the
svelte world and I believe I have a working build, but frankly I'm solidly
out of my depth here. One concerning aspect is that the built dist/index.js
is 128kB which makes me think I have some things misconfigured. Hoping to
poke a bit more and perhaps solicit some help from folks who know more than
I do, but wanted to share what I have now.

### DIFF
--- a/packages/inertia-svelte/package.json
+++ b/packages/inertia-svelte/package.json
@@ -28,6 +28,7 @@
     "@rollup/plugin-node-resolve": "^6.0.0",
     "eslint": "^7.0.0",
     "rollup": "^2.30.0",
+    "rollup-plugin-peer-deps-external": "^2.2.4",
     "rollup-plugin-svelte": "^7.0.0",
     "svelte": "^3.31.2"
   },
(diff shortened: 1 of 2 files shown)


christoomey:
Small update based on some pointers from folks on twitter to add a rollup plugin to keep peer deps out of built dist/index.js. Down to 20k (8k gzipped) now, but still thinking that's unreasonably high.

mariojankovic:
Hey @christoomey, this looks good but needs additional tuning. Adding stuff as peerDeps requires both the `external` and `globals` flags. I've fixed these and introduced `terser` to uglify the end results and have been able to bring this down to 7kb and 8kb respectfully. Also note that rollup supports `ROLLUP_WATCH` to watch for dev/production envs.

christoomey:
Closing in favor of https://github.com/inertiajs/inertia/pull/531.

claudiodekker:
Thanks for the contribution @christoomey!

PR #445, opened 2021-01-17
Add array support for errors to form helpers

### LINKED ISSUE #437: Access Laravel array validation messages from Inertia
### Versions:

### Describe the problem:
I'm trying to access to all the messages passed from an array validation in Laravel.

### Steps to reproduce:
In the controller I defined this validation method:
```php
request()->validate([
    'parameters.*.value' => 'required'
]);
```

This is what Laravel pass to Inertia props:
```js
props: {
  errors: {
    "parameters.0.value": "The field is required",
    "parameters.1.value": "The field is required"
}
```

I'm using the new Inertia form validation helper so I have something like this in the Vue component:
`<jet-input-error :message="form.errors.*" class="mt-2" />`

But I can't access to the specific error in the array, because the <jet-input-error> it's inside a foreach in a form

ajnsn:
Resolves #437

This PR adds support for array based (dot.noted) error messages to the form helpers. It also adds support for multiple error messages per field and clearing all nested errors.

When using Laravel, you validate an array like this

```php
$request->validate([
  'field' => 'required|array|min:3',
  'field.*.value' => 'required',
]);

```

The form.errors object is populated as follows

```json
{
  "field": "The field must have at least 3 items.",
  "field.0.value": "The field.0.value field is required.",
  "field.1.value": "The field.1.value field is required."
}
```

### getErrors('field')

The introduced `getErrors('field')` will return an array of error messages or `null` if there are no errors. 

```js
[
  "The field must have at least 3 items.",
  "The field.0.value field is required.",
  "The field.1.value field is required."
]
```

You can use `v-if` (or `v-show`) and `v-for`.

```html
<div v-if="form.getErrors('field')">
    <p v-for="(message, index) in form.getErrors('field')" :key="index" v-text="message" />
</div>
```

It's totally optional. You still can access `form.errors.field` directly. No breaking changes.

If you want to have an universal approach to always receive error messages as an array, you may want it use also for non-array values. 

Even multiple error message per field are supported, e.g. if you override or do not use the `resolveValidationErrors()` in the Laravel adapter to not only return the first error message via `$errors[0]`, you will receive receive a response like this:

```json
{
  "field": [
    "The field must be 6 digits.",
    "The selected field is invalid."
  ]
}
```

`getErrors('field')` will return:

```js
[
  "The field must be 6 digits.",
  "The selected field is invalid."
]
```

### clearErrors('field')

The `clearErrors()` now also clears related error messages, e.g. `clearErrors('field')` clears `field`, `field.0.value`, `field.1.value` etc.

It still works for non-array fields, also no breaking change here.

### DIFF
--- a/packages/inertia-vue/src/form.js
+++ b/packages/inertia-vue/src/form.js
@@ -44,12 +44,30 @@ export default function(data = {}) {
 
       return this
     },
+    getErrors(field) {
+      const errors = Object
+        .keys(this.errors)
+        .filter(key => key === field || key.startsWith(`${field}.`))
+        .reduce((carry, key) => {
+          if (Array.isArray(this.errors[key])) {
+            this.errors[key].forEach(message => {
+              carry.push(message)
+            })
+          } else {
+            carry.push(this.errors[key])
+          }
+
+          return carry
+        }, [])
+
+      return errors.length > 0 ? errors : null
+    },
     clearErrors(...fields) {
       this.errors = Object
         .keys(this.errors)
-        .reduce((carry, field) => ({
+        .reduce((carry, key) => ({
           ...carry,
-          ...(fields.length > 0 && !fields.includes(field) ? { [field] : this.errors[field] } : {}),
+          ...(fields.length > 0 && !fields.some(field => key === field || key.startsWith(`${field}.`)) ? { [key]: this.errors[key] } : {}),
         }), {})
 
       this.hasErrors = Object.keys(this.errors).length > 0
--- a/packages/inertia-vue3/src/form.js
+++ b/packages/inertia-vue3/src/form.js
@@ -45,12 +45,30 @@ export default function form(data = {}) {
 
       return this
     },
+    getErrors(field) {
+      const errors = Object
+        .keys(this.errors)
+        .filter(key => key === field || key.startsWith(`${field}.`))
+        .reduce((carry, key) => {
+          if (Array.isArray(this.errors[key])) {
+            this.errors[key].forEach(message => {
+              carry.push(message)
+            })
+          } else {
+            carry.push(this.errors[key])
+          }
+
+          return carry
+        }, [])
+
+      return errors.length > 0 ? errors : null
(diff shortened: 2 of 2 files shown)


claudiodekker:
Personally, I'm not a huge fan of this change for two reasons/concerns:

1. It would fragment the 'form errors' API that we now have: IMO there should be a single way to do it, otherwise you end up with one method that can get/show your error, and another that can get/show your error _even if it's deeply nested_, which is weird.
2. Given that we already have a [server-side error sharing mechanism that you can modify](https://github.com/inertiajs/inertia-laravel/blob/master/src/Middleware.php#L44-L51), as well the [modifiable client-side `resolveErrors` callback](https://github.com/inertiajs/inertia/blob/master/packages/inertia/src/inertia.js#L11), I'm having a hard time seeing the use-cases/situations in which this feature would be useful?

ajnsn:
I often have the situation that there a group of fields (e.g. checkboxes) or a bunch of select-boxes to be validated. Then it's very helpful and often an UX requirement to display the error messages at the botton of this input group.

This was also inspired by [Spatie's Errors class](https://github.com/spatie/form-backend-validation/blob/master/src/Errors.js) which supports those array errors. 👍 

Yes, you can rewrite `resolveValidationErrors` to combine all array errors into one but debugging will be hard and you if-clause by if-clause e.g. for a separte approach for one page. 

I agree about the fragmentation though. I was also thinking about Rails handles this and how this could be compatible with the Rails adapter response. So I expected comments and feedback. Thanks!

rovansteen:
@claudiodekker given that `resolveErrors` has been removed. Are you willing to reconsider implementing this or another solution for this problem?

claudiodekker:
@rovansteen Probably not, as you can still modify this server-side, and conditionals (e.g. something that happens on a single page) can also be done by accessing the Request URL before 'transforming' them. A large part of the reason we ended up removing `resolveErrors` is because it ended up causing more problems than it solved.

PR #451, opened 2021-01-20
Change Vue3 adapter to reactive

ajnsn:
A little "addon" to #440. I noticed that the Vue3 adapter is also using `ref()` for non-primitive values.

This PR changes to a reactive `current` object. This should be also more comparable what the React adapter does with `useState()`.

### DIFF
--- a/packages/inertia-vue3/src/app.js
+++ b/packages/inertia-vue3/src/app.js
@@ -1,12 +1,14 @@
 import form from './form'
 import link from './link'
 import remember from './remember'
-import { computed, h, markRaw, ref } from 'vue'
+import { computed, h, markRaw, reactive } from 'vue'
 import { Inertia } from '@inertiajs/inertia'
 
-const component = ref(null)
-const page = ref({})
-const key = ref(null)
+const current = reactive({
+  component: null,
+  page: {},
+  key: null,
+})
 
 export default {
   name: 'Inertia',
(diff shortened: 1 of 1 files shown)


reinink:
Hey thanks @ajnsn! Cool, thanks for this PR. I'll do some testing and see how it works. 👍 

One question, is this a breaking change in any way?

ajnsn:
I've tested it in a current project and did not encounter any changes. Afaik Vue uses `reactive()` under the hood if you init a non-primitive value with `ref()`, so I would not expect any changes.

But I did not test it thoroughly.

haikyuu:
I think it would be a good idea to have integration tests using pingCRM and [playwright](https://playwright.dev/). I'm currently working on a node adapter and I will probably give it a shot after porting ping crm to node & vue.

Integration tests are the way to go since Inertia's definition of works is: that an inertia backend adapter works well with the front-end library and one of its framework adapters.

claudiodekker:
@haikyuu We actually already have a test suite that's ready to go (https://github.com/inertiajs/inertia/pull/395, @reinink just hadn't had the time to review it yet), so I'd recommend using that instead 👍

reinink:
@ajnsn Hey thanks again for this PR, but I think I'm just going to leave things as they are right now, as the current setup works great, and I'd rather not touch it unless there's an obvious benefit. 👍

PR #456, opened 2021-01-24
Deep clone form defaults

### LINKED ISSUE #455: Inertia form helper cast attributes to string
### Versions:

### Describe the problem:

I'm using Buefy (Bulma for Vue.js)
I have a edit page that receives data from Laravel backend, and in this form, I have date fields and the buefy datepicker component accepts Date, or Array.
In the data page attribute i'm initializing all the data that comes from Laravel, including creating a new date object for that date field.
The problem is that, when the page is rendered, the date attribute is casted to string causing the datepicker to throw an error.
The expected result is that the form helper does not cast to string the attributes.

### Steps to reproduce:

```vue
<template>
    <section class="section">
        <div class="container">
            <p class="title">Edit page</p>
            <form>
                <b-field label="This date field don't work">
                    <b-datepicker
                            v-model="form.thisDateDontWork"
                            editable
                            required
                    />
                </b-field>
                <b-field label="This date works">
                    <b-datepicker
                            v-model="thisDateWorks"
                            editable
                            required
                    />
                </b-field>
            </form>
        </div>
    </section>
</template>

<script>
  export default {
    data() {
      const backendData = {
        date: "2021-03-05T03:00:00.000000Z"
      }
      
      return {
        form: this.$inertia.form({
          thisDateDontWork: new Date(backendData.date)
        }),
        thisDateWorks: new Date(backendData.date)
      }
    }
  }
</script>
```

ajnsn:
Resolves #455 

This PR deep clone the defaults object and adds support for complex types (e.g. `Date()`) within the form (defaults) object. This replaces the current deep clone with `JSON.parse/stringify`.

`cloneDeep` is using a helper function `getType` to determine if we need to go deeper, e.g. on `objects` and `arrays` or if we need to return the complex type. As nearly everything in JS is an object (`typeof new Date() === 'object'` will also be `true`), this helps the `cloneDeep` to have universal approach. So we do not need to check for every single type which typically results in multiple `if` and `else if` clauses.

### DIFF
--- a/packages/inertia-vue/src/form.js
+++ b/packages/inertia-vue/src/form.js
@@ -1,7 +1,13 @@
 import { Inertia } from '@inertiajs/inertia'
+import { cloneDeep } from './objects'
 
 export default function(data = {}) {
-  const defaults = JSON.parse(JSON.stringify(data))
+  const defaults = Object.keys(data).reduce((carry, key) => {
+    carry[key] = cloneDeep(data[key])
+
+    return carry
+  }, {})
+
   let recentlySuccessfulTimeoutId = null
   let transform = data => data
 
--- a/packages/inertia-vue/src/objects.js
+++ b/packages/inertia-vue/src/objects.js
@@ -1,0 +1,21 @@
+export function cloneDeep(object) {
+  const type = getType(object)
+
+  if (type === 'array') {
+    return object.map(function (item) {
+      return cloneDeep(item)
+    })
+  } else if (type === 'object') {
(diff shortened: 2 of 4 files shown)


reinink:
Hey @ajnsn, thanks so much for all your effort reporting bugs and coming up with fixes. I've decided to pull in a library to do the deep cloning, which you can see here: c0c2d97739a6712179a9117648e35972acf21d87

Meaning, this will be fixed as part of #575! 👍 

Even though your PR won't be merged in, it's still helpful. 💜

PR #486, opened 2021-02-08
Store page props in session storage

RobertBoes:
Currently Inertia "enforces" a character limit to the response size, especially for Firefox. This is because all page props are stored within the browsers History object. This artificial limit can be quite cumbersome to workaround, data is usually user-generated and having a few varchar(255) fields with a text field can potentially cause issues.

With this PR I wanted get some feedback about the idea of using sessionStorage for props and if there's any issues with this that I'm missing.

### DIFF
--- a/packages/inertia/package.json
+++ b/packages/inertia/package.json
@@ -30,7 +30,8 @@
   "dependencies": {
     "axios": "^0.21.1",
     "deepmerge": "^4.0.0",
-    "qs": "^6.9.0"
+    "qs": "^6.9.0",
+    "uuid": "^8.3.2"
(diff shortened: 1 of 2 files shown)


reinink:
Hey Robert! Thanks for your efforts here, and for your interest in Inertia.js.

So, at this time I don't think I want to resort to using other storage mechanisms beyond history state. I sort of feel like pages should never show more than `640kB` of data (which is a TON). Typically there are ways to reduce the amount of data sent to the page to prevent this issue.

That said, we have had reports of this, and will certainly reconsider this in the future if needed.

PR #515, opened 2021-02-22
Add blob support

### LINKED ISSUE #512: Assigning a blob to a form does not convert it to FormData
### Versions:

- `@inertiajs/inertia` version: 0.8.4
- `@inertiajs/inertia-vue3` version: 0.3.5

### Describe the problem:

When assigning a `Blob` to a form created with `useForm`, the form's payload is not converted to a `FormData` object prior to submission. Hence, the uploaded file is never sent.

### Steps to reproduce:

```js
form : useForm({
    avatar : Blob,
}),
```

reinink:
Fixes #512

This PR adds `Blob` support to Inertia, when determining if a request should be sent using `FormData`.

### DIFF
--- a/packages/inertia/src/files.js
+++ b/packages/inertia/src/files.js
@@ -1,5 +1,6 @@
 export function hasFiles(data) {
   return data instanceof File
+    || data instanceof Blob
     || data instanceof FileList
     || (typeof data === 'object' && data !== null && Object.values(data).find(value => hasFiles(value)) !== undefined)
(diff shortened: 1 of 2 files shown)


claudiodekker (approved):

reinink:
Published: https://github.com/inertiajs/inertia/releases/tag/v0.8.5

ninjaparade (commented):
I'll allow this.

PR #516, opened 2021-02-24
Add form helper to React adapter

reinink:
This PR adds the form helper to the React adapter, similar to the existing form helpers in the Vue 2 and Vue 3 adapters. The main difference between this form helper and the Vue form helper is that this one automatically remembers the form data and errors in history state. I actually think it makes sense to update the Vue form helpers to do the same thing.

## Example

```jsx
import React from 'react'
import { useForm } from '@inertiajs/inertia-react'

import TextInput from './TextInput'
import LoadingButton from './LoadingButton'

const { data, setData, post, processing, errors } = useForm({
  name: '',
  email: '',
})

function submit(event) {
  event.preventDefault()
  post('/users')
}

return (
  <form onSubmit={submit}>
    <TextInput label="Name" value={data.name} errors={errors.name} onChange={e => setData('name', e.target.value)} />
    <TextInput label="Email" value={data.email} errors={errors.email} onChange={e => setData('email', e.target.value)} />
    <LoadingButton loading={processing} type="submit">Submit</LoadingButton>
  </form>
)
```

## API

Here are the available properties:

- `data`
- `errors`
- `hasErrors`
- `processing`
- `progress`
- `wasSuccessful`
- `recentlySuccessful`

And the available methods:

- `setData(key, value)`, `setData(values)`, `setData(callback)`
- `transform(callback)`
- `reset()`, `reset(fields)`
- `clearErrors()`, `clearErrors(fields)`
- `submit(method, url, options)`
- `get(url, options)`
- `post(url, options)`
- `put(url, options)`
- `patch(url, options)`
- `delete(url, options)`

## Naming considerations

Keep in mind that `delete` is a reserved keyword, meaning if you'd like to destructure, you'll need to alias it. For example:

```js
const { delete: destroy } = useForm({ ... });
```

I'm also not sure if we need both `set()` and `setData()`. I like `set()` as a shortcut to set a specific field...but maybe we should call it `setField()` instead, or just let folks rely on `setData()`. The other issue with `set()` right now is that it only works for top level fields. For example: `set('name', name)`. It would be great to also support dot notation for the keys, like this: `set('user.name', name)`. Some choices here:

```js
// Option 1: set() and SetData()
form.set('name', value)
form.setData({ name: value })

// Option 2: setField() and setData()
form.setField('name', value)
form.setData({ name: value })

// Option 3: set() with type check
form.set('name', value)
form.set({ name: value })

// Option 4: setData() with type check
form.setData('name', value)
form.setData({ name: value })
```

## Future improvements

- Add the ability to use dot/array notation with `setData(key, value)` to set a nested value.
- Add the ability to manually set errors for client-side validation.
- Update the Vue form helpers to automatically remember the form data and errors in history state (like this form helper).

### DIFF
--- a/packages/inertia-react/src/index.js
+++ b/packages/inertia-react/src/index.js
@@ -1,4 +1,5 @@
 export { default as App, default as app, default as InertiaApp } from './App'
 export { default as Link, default as link, default as InertiaLink } from './Link'
+export { default as useForm } from './useForm'
 export { default as usePage } from './usePage'
 export { default as useRemember, useRememberedState } from './useRemember'
--- a/packages/inertia-react/src/useForm.js
+++ b/packages/inertia-react/src/useForm.js
@@ -120,6 +120,7 @@ export default function useForm(defaults, key) {
       )
       setHasErrors(Object.keys(errors).length > 0)
     },
+    submit,
     get(url, options) {
       submit('get', url, options)
     },
--- a/packages/inertia-react/src/useForm.js
+++ b/packages/inertia-react/src/useForm.js
@@ -81,16 +81,21 @@ export default function useForm(defaults, key) {
 
   return {
     data,
-    setData,
+    setData(key, value) {
+      if (typeof key === 'string') {
+        setData({ ...data, [key]: value })
+      } else if (typeof key === 'function') {
+        setData(data => key(data))
+      } else {
+        setData(key)
+      }
+    },
     errors,
     hasErrors,
     processing,
     progress,
     wasSuccessful,
     recentlySuccessful,
-    set(key, value) {
-      setData({ ...data, [key]: value })
-    },
     transform(callback) {
       transform = callback
     },
--- a/packages/inertia-react/src/useForm.js
+++ b/packages/inertia-react/src/useForm.js
@@ -3,7 +3,7 @@ import { useRef, useCallback, useState } from 'react'
 import useRemember from './useRemember'
 
 export default function useForm(defaults, key) {
-  let transform = data => data
+  let transform = (data) => data
 
   const recentlySuccessfulTimeoutId = useRef(null)
   const [data, setData] = useRemember(defaults, `${key}-data`)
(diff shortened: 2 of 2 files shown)


Landish:
Hey @reinink , excellent PR. 

Couple of questions though. 

1. What is the use case for `recentlySuccessful`? And why it's hardcoded for 2 seconds? 
2. Maybe `data` can be renamed to `values` ? 

> I'm also not sure if we need both set() and setData(). I like set() as a shortcut to set a specific field...but maybe we should call it setField() instead, or just let folks rely on setData(). 

I would probably call this `setValue` and `setValues` and have both methods available.  

P.S. Can't wait to update my projects, including [pingcrm-react](https://github.com/Landish/pingcrm-react/).

claudiodekker:
@Landish https://inertiajs.com/forms#form-helper
> When a form has been successfully submitted, the `form.wasSuccessful` property will be `true`. In addition to this, there is also a `form.recentlySuccessful` property, which will be set to `true` for two seconds after a successful form submission. This is helpful for showing temporary success messages.

As for `data`, this primarily has to do with consistency that we maintain internally all across the board (in all the other parts of the app as well as in the other adapter's form helpers)

Landish:
@claudiodekker 

My bad. Haven't read the form adapter docs for vue.  In that case it makes sense. 👍

reinink:
@claudiodekker Thanks for jumping in. 👍 

@Landish I do really like `setValue` and `setValues`, but yeah, we don't use "value/values" ever...it's always data. 🤔

claudiodekker (approved):

hailwood:
Why not make set work with both?

```js
.set('name', value)
.set({name: value})
// and eventually 
.set('nested.name', value')
.set({'nested.name': value})
```

Can we also have a generic `submit` method that takes a method?

reinink:
@hailwood Yeah, I considered just having "set", that works for all the situations. Should it be called `setData()` maybe though...to match typical `setState()` naming?

Also, curious, why would you want the generic `submit` method? When would you use this over one of the other methods?

hailwood:
I personally would reach for `setField('', '')` and `setField(s)({'': ''})` if I was going to try and use it without reading the docs (and let's be real, in a team of multiple developers that happens more than we'd like - one implements, the others copy/paste & rely on autocompletion).

I actually think `set` and `setData` are too generic here, it makes me wonder if I can update errors since they're data too right?

On that note, could we have `setError`/`setErrors` too? I can see myself wanting to merge in client side validation with the server side validation

reinink:
Just posting my naming ideas here as well:

```js
// Option 1: set() and SetData()
form.set('name', value)
form.setData({ name: value })

// Option 2: setField() and setData()
form.setField('name', value)
form.setData({ name: value })

// Option 3: set() with type check
form.set('name', value)
form.set({ name: value })

// Option 4: setData() with type check
form.setData('name', value)
form.setData({ name: value })
```

jryd:
We tend to use uncontrolled inputs for a lot of our forms. We add an `onSubmit` handler that then does something like `const data = new FormData(e.target)`.

I suppose the idea, with `setData` would be something like this:

```
import React from 'react';
import TextInput from '@/Shared/TextInput';
import LoadingButton from '@/Shared/LoadingButton';
import { useForm } from '@inertiajs/inertia-react';

const { data, setData, post, processing, errors } = useForm({
  name: '',
  email: '',
});

function submit(event) {
  event.preventDefault();
  setData(Object.fromEntries(new FormData(e.target))
  post('/users')
}

return (
  <form onSubmit={submit}>
    <UncontrolledTextInput label="Name" name="name" errors={errors.name} defaultValue={data.name} />
    <UncontrolledTextInput label="Email" name="email" errors={errors.email} defaultValue={data.email} />
    <LoadingButton loading={processing} type="submit">Submit</LoadingButton>
  </form>
)
```

Is that broadly the approach you see being taken?

Given that `setData` is a hook, I don't believe it is safe to set its value (an async operation) and then immediately post.

reinink:
@jryd So the issue with uncontrolled inputs is that Inertia cannot remember their values to the history state using the [remember state feature](https://inertiajs.com/remembering-state), which this form helper user.

And yes, you're right, calling `setData()` right before `post()` will result in the data not being sent, because it is asynchronous. I think the correct way to do this is to use a `useEffect` and watch for changes to `data`. Just keep in mind that `useEffect` runs on the initial render...which is kind of annoying.

So, while I like your approach (`Object.fromEntries(new FormData(e.target)` is very clever!), I'm not sure this is the best approach with Inertia and this particular form helper.

reinink:
@hailwood Thanks for your input! Responses:

> I actually think `set` and `setData` are too generic here, it makes me wonder if I can update errors since they're data too right?

I get where you're coming from, but I also think the concept of "form data" is very common. In fact, there's the [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) interface that uses this exact language.

> On that note, could we have `setError`/`setErrors` too? I can see myself wanting to merge in client side validation with the server side validation

We don't currently support this with the Vue form helper, but I'm not necessary opposed. I'd suggest we wait until this PR is merged in, and then consider that as a subsequent addition.

> The submit method is more for a wrapper, I'd probably write something like

This has been added: a836b6c7147a0a0cf15cd12f3e5a52f54493eb0c. Will that work for your use cases?

anubra266:
@reinink I go with:
>// Option 1: set() and SetData()
>form.set('name', value)
>form.setData({ name: value })

Don't know how far fetched it is, but one beautiful feature would be the laravel way. i.e. make it dynamic.
If I have
```js
const { data, setData, post, processing, errors } = useForm({
  name: '',
  email: '',
});
```

Then I want to be able to  do: 
```js
form.setName(value)
form.setEmail(value)
```

hailwood:
Thanks @reinink that's perfect.

reinink:
Okay, so I decided to go with "Option 4: setData() with type check": eccbbdf7bafbeb0bfa36f64b8210d6aa9aafe8c3

This can be used three ways:

```js
// Key/value
form.setData('name', value)

// Value
form.setData({ name: value })

// Callback
form.setData(data => ({ ...data, name: value }))
```

I would like to eventually support either dot notation or array notation for the key/value approach, but I think we'll handle that as a subsequent PR.

PR #531, opened 2021-03-03
WIP Add build step for inertia-svelte

### LINKED ISSUE #394: svelte: Class constructor Link cannot be invoked without 'new'

mariojankovic:
Introducess additional changes to Chris' PR. See here: https://github.com/inertiajs/inertia/pull/439
Fixes #394.


packages/inertia-svelte/package.json
```diff
@@ -19,13 +19,27 @@
...
+  },
+  "scripts": {
+    "build": "rollup -c",
+    "watch": "rollup -c -w",
+    "prepublishOnly": "npm run build"
   },
   "peerDependencies": {
-    "@inertiajs/inertia": "^0.8.0",
-    "svelte": "^3.20.0"
+    "@inertiajs/inertia": "0.8.5",
```
reinink, line 42: Do we need `@inertiajs/inertia`  as both `devDependencies` and a `peerDependencies`? Shouldn’t it be one or the other?

mariojankovic: Nope. Good catch.

christoomey:
@mariojankovic this is great! Thanks for picking this up. I'm happy to consolidate efforts here and have closed my PR.

I've just tested this change locally and while it does resolve the `Class constructor Link cannot be invoked without 'new'` issue, I'm now seeing a new issue.

Using the `inertia` action is throwing an `Error: Function called outside component initialization` error, seemingly related to the [usage of createEventDispatcher](https://github.com/inertiajs/inertia/blob/da13285e33bde891ab968515d31658882a2932ce/packages/inertia-svelte/src/link.js#L9-L12) in an action ([this issue seems to cover similar things](https://github.com/sveltejs/sapper/issues/592#issuecomment-469924549)). I'm poking at different versions locally to see if I can resolve this and I'll update if I get to anything.

reinink:
I've been able to test this locally as well using [Ping CRM Svelte](https://github.com/inertiajs/pingcrm-svelte), and it appears to be working. 🙌 

I am about to do the dev and prod builds, using either `import` (code splitting) or `require` (one big bundle).

mariojankovic:
@christoomey that's great. I've tried including svelte within `devDependencies` and it seems the file size went up from 7kb to ~12kb (production /w terser). If that's fine you can test if that works for you.

nathancahill:
This looks good to me!

mariojankovic:
The PR does seem to fix this issue. I've been able to replicate the setup with no errors when using the transpiled files from the `dist` folder instead of the ones from `package.json`.

reinink:
Closing this in favour of #549. 👍

PR #534, opened 2021-03-04
add property to set axios - http header

StefanNeuser:
If you tell me what is your code formatting standard i will use them.

Please search for `axiosConfig` . The change is all about to have the ability to apply global headers for all Axios calls that inertia will make.

This is a use case in one of your Shopify App. We some headers for the hmac check in a third party application world:

```javascript
import { InertiaApp } from '@inertiajs/inertia-react'
import React from 'react'
import { render } from 'react-dom'
import axios from 'axios'

window.axios = axios
window.axios.defaults.headers.common['X-CSRF-TOKEN'] = document.head.querySelector('meta[name="csrf-token"]').content
window.axios.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest'

document.head.querySelectorAll('meta[name^="X-Shop-"]').forEach(
  elem => window.axios.defaults.headers.common[elem.name] = elem.content
)

render(
  <InertiaApp
    axiosConfig={{ headers: window.axios.defaults.headers }}
    initialPage={JSON.parse(app.dataset.page)}
    resolveComponent={name =>
      import(`@/views/${name}`).then(module => module.default)
    }
  />,
  app
)
```

### DIFF
--- a/packages/inertia-react/src/App.js
+++ b/packages/inertia-react/src/App.js
@@ -5,23 +5,28 @@ import PageContext from './PageContext'
 export default function App({
   children,
   initialPage,
+  axiosConfig,
   resolveComponent,
   resolveErrors,
   transformProps,
-}) {
+})
+{
   const [current, setCurrent] = useState({
     component: null,
     page: {},
     key: null,
   })
 
-  useEffect(() => {
+  useEffect(() =>
+  {
     Inertia.init({
       initialPage,
+      axiosConfig,
       resolveComponent,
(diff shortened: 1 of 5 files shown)


reinink:
Hey there @StefanNeuser! Thanks for your interest in Inertia.js, and for this contribution.

And while I like this idea, we're actually planning on removing Axios from Inertia sooner than later (see #216). As part of that change, I want to find a way to support global headers.

Hope that makes sense! 👍

PR #549, opened 2021-03-13
Renames link component to avoid compilation conflicts.

### LINKED ISSUE #285: Error loading `inertia-svelte` under Rails (via webpacker)
```
8:34:16 PM webpack.1 |  Version: webpack 4.44.2
8:34:16 PM webpack.1 |  Time: 3899ms
8:34:16 PM webpack.1 |  Built at: 10/25/2020 8:34:15 PM
8:34:16 PM webpack.1 |                                                     Asset       Size       Chunks                         Chunk Names
8:34:16 PM webpack.1 |                       application-141aa79e56d42feb82ca.js    986 KiB  application  [emitted] [immutable]  application
8:34:16 PM webpack.1 |                   application-141aa79e56d42feb82ca.js.map   1.02 MiB  application  [emitted] [dev]        application
8:34:16 PM webpack.1 |                            client-141aa79e56d42feb82ca.js    654 KiB       client  [emitted] [immutable]  client
8:34:16 PM webpack.1 |                        client-141aa79e56d42feb82ca.js.map    674 KiB       client  [emitted] [dev]        client
8:34:16 PM webpack.1 |                                             manifest.json  832 bytes               [emitted]              
8:34:16 PM webpack.1 |  media/images/banner-5f539d39241aceb17c07bbaa6e611cac.svg   16.2 KiB               [emitted]              
8:34:16 PM webpack.1 |    media/images/icon-03960d2ed0f0c47a77a9fbf21ddf798f.svg   2.71 KiB               [emitted]              
8:34:16 PM webpack.1 |  ERROR in ./node_modules/@inertiajs/inertia-svelte/src/index.js
8:34:16 PM webpack.1 |  Module not found: Error: [CaseSensitivePathsPlugin] `/Users/jason/Projects/gateway/node_modules/@inertiajs/inertia-svelte/src/link.svelte` does not match the corresponding path on disk `Link.svelte`.
8:34:16 PM webpack.1 |  ℹ ｢wdm｣: Failed to compile.
```

I followed the instructions in the Inertia docs, and couldn't get past this issue. I believe it has to do with webpack being confused while loading both `Link.svelte` and `link.js`. My app has a fairly vanilla setup, but I will try to reproduce it in a empty rails app.

mariojankovic:
Fixes #285 and #394 without introducing additional build setups.


reinink:
Thanks so much for digging into this @mariojankovic!

I am curious though, why exactly can't we have a component named `Link.svelte`? Where exactly is the conflict coming? From Svelte core? Or from user-land code? If user-land, can't they simply rename to avoid the conflict?

reinink:
Never mind, I see what's wrong here. We have both a `Link.svelte` and `link.js` file, and these are conflicting. Nice find @mariojankovic. 🙌

reinink:
@christoomey Are you able to test this solution in Rails? I think we've finally got down to the root of the issue here.

christoomey:
@mariojankovic this is great! Just tested and this definitely resolved the `CaseSensitivePathsPlugin` error (so nice to not see that warning in the output!).

Unfortunately, I'm still running into the `Cannot read property 'c' of undefined` issue right now. @mariojankovic I know you tested against the reproduction repo I shared, but I'm wondering if you happened to click around? I'm seeing the mentioned issue only after clicking an inertia-powered link (apologies for not making this clear initially). I've just pushed a commit to that repo (https://github.com/christoomey/inertia-svelte-repro) that adds some notes to the README on how to reproduce the issue and simplifies getting started. Would you potentially be able to take another look?

As an aside, I ran into an issue where `inertia-svelte` couldn't find `svelte` when I first tried to use `yarn link` to test. I was able to get things working by running `yarn add svelte` in the `inertia-svelte` package, but was surprised that the local `node_modules` was empty and that it installed in the `../../node_modules` in the root. I'm guessing this is fine/normal and just reflects my novice-ness with yarn link and monorepos, but wanted to highlight in case it is a bigger issue.

---

Tangentially related to all of the above, thank you _so much_ for the continued effort here. The combo of intertia and svelte are increasingly my preferred way to build apps, but unfortunately, these fixes have been outside my knowledge base enough. Really exciting to see this specific package getting some attention. Just wanted to clarify that it is very appreciated!!!

reinink:
Thanks so much @mariojankovic! 🙌

PR #526, opened 2021-02-27
Add react form helper typings

anubra266:
Closes #523 

This [PR](#526) adds Typescript declarations for the `useForm` hook, used by the React adapter.

image

PR #555, opened 2021-03-17
Generic useForm

Bloemendaal:
Hi 👋!

While programming, I patched inertia/react with the typescript variant of useForm. However, the current implementation is not generic, which makes type suggestions from my IDE useless.

Here is my implementation of useForm, thanks to #526 😁!

### DIFF
--- a/packages/inertia-react/index.d.ts
+++ b/packages/inertia-react/index.d.ts
@@ -84,4 +84,4 @@ export interface InertiaFormProps<TForm = Record<string, any>> {
 	delete: (url: string, options?: Inertia.VisitOptions) => Promise<void>
 }
 
-export function useForm<TForm = Record<string, any>>(initialValues: TForm | Partial<TForm>): InertiaFormProps<TForm>;
+export function useForm<TForm = Record<string, any>>(initialValues: TForm): InertiaFormProps<TForm>;
--- a/packages/inertia-react/index.d.ts
+++ b/packages/inertia-react/index.d.ts
@@ -60,25 +60,28 @@ export const Link: InertiaLink
 
 export const InertiaApp: App
 
-export interface InertiaFormProps {
-	data: object
-	errors: any
+type setDataByObject<TForm> = (data: TForm) => void
+type setDataByMethod<TForm> = (data: (previousData: TForm) => TForm) => void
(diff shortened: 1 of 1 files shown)


packages/inertia-react/index.d.ts
```diff
@@ -60,25 +60,28 @@ export const Link: InertiaLink
...
+	patch: (url: string, options?: Inertia.VisitOptions) => Promise<void>
+	post: (url: string, options?: Inertia.VisitOptions) => Promise<void>
+	put: (url: string, options?: Inertia.VisitOptions) => Promise<void>
 	delete: (url: string, options?: Inertia.VisitOptions) => Promise<void>
 }
 
-type InertiaForm = (initialValues: { [key: string]: any }) => InertiaFormProps
-export const useForm:InertiaForm;
\ No newline at end of file
+export function useForm<TForm = Record<string, any>>(initialValues: TForm | Partial<TForm>): InertiaFormProps<TForm>;
```
innocenzi, line 87: What's the benefit of using `TForm | Partial<TForm>` instead of just `Partial<TForm>`?

Bloemendaal: That's because I wasn't sure if all values should have an initial value. If they do, `Partial<TForm>` should be left out, otherwise, just `Partial<TForm>` is fine.

innocenzi: I think we should remove `Partial<TForm>` from the union, because if we don't specify a type for `TForm` it is not useful, and if we do specify a type, we most likely want to be exhaustive when setting the initial values - and if we don't, we can use `Partial<OurType>` on `useForm` anyway as an escape gate. 

What do you think?

```suggestion
export function useForm<TForm = Record<string, any>>(initialValues: TForm): InertiaFormProps<TForm>;
```

Bloemendaal: Agreed! They can also set optional values in `TForm` and leave them undefined. Enough options 👍

innocenzi (approved):
I didn't actually test, but the typings look good to me. 👍

sebastiandedeyne (approved):

reinink:
Thanks!

PR #558, opened 2021-03-18
Adds form helper for Svelte

mariojankovic:
I'm happy to announce that the Svelte form helper is almost live. For anyone involved - please checkout this PR and test it out. We want to make sure this works as intended before it goes live.

You can try it out with the PingCRM demo or in one of your projects by doing the following:
```html
<script>
import { useForm } from '@inertiajs/inertia-svelte'

let form = useForm({
  name: null,
  email: null,
  phone: null,
  address: null,
  city: null,
  region: null,
  country: null,
  postal_code: null,
})

function submit() {
  $form.post(route('organizations.store'))
}
</script>

<form on:submit|preventDefault={submit}>
  <input type="text" bind:value={$form.name} />
  <input type="text" bind:value={$form.email} />
  <input type="text" bind:value={$form.phone} />
  <input type="text" bind:value={$form.address} />
  <input type="text" bind:value={$form.city} />
  <input type="text" bind:value={$form.region} />

  {#if $form.processing}
    <p class="processing">Please wait</p>
  {/if}

</form>
```


reinink:
This looks awesome @mariojankovic! Thanks so much for cracking this! 🙌

I've made a few changes, mostly around error handling and form resetting. Those were not working 100%.

I do wonder about one thing...the naming. I see you're calling the file `form.js`, but the export is `useForm`. However, our `remember` helper is just exported as `remember`. I wonder if we should either change `useForm` to just `form`, or maybe update `remember` to be `useRemember`.

I personally prefer `useForm` and `useRemember`, since that's consistent with what we're doing in the Vue 3 and React plugins. Plus, more importantly, you don't run into variable naming collisions this way. What do you think?

mariojankovic:
@reinink Happy to help. I agree, I'd consider renaming `remember` to `useRemember` since we're using the same in both the vue and react adapter. I guess the naming stems from react's naming convention like `useRef` or `useState` but I'm happy we keep it like that as `form` or `remember` seems too generic, and we've already had to deal with conflicts in our svelte adapter by now.

reinink:
Awesome, I've updated `form` to `useForm`, and I've updated `remember` to `useRemember`. For backwards compatiblitiy, I've keep `remember` as an export as well. 👍

reinink:
Okay, so I just finished adding automatic "remember" functionality to the form helper. This will automatically remember and restore form data and errors in history state. You can find these changes here: 77b04467bc8694ae39f0c2cb4c32ca3c31584eb8

I think everything is ready to go now. All look good @mariojankovic?

reinink:
🙌

PR #569, opened 2021-03-23
Fix bug `<InertiaLink>` without default slot

Neophen:
This seems to break the render function when you use the `<InertiaLink :href="" />` without the default slot provided.


Neophen:
Duplicate of #306

PR #575, opened 2021-03-25
Automatically remember form data and errors

### LINKED ISSUE #417: Issue using 'remember' since Inertia Form update
### Versions:

### Describe the problem:

Upgraded our app to Jetstream V2 this morning, and as a consequence to the latest versions of Inertia and the Vue 2 adapter, and started seeing the following error:

```
[Vue warn]: Error in created hook: "TypeError: Cannot read property 'serialize' of null"
```

### Steps to reproduce:

The below example should allow you to easily recreate the issue:

```
<script>
export default {
  data: () => ({
    foo: null,
  }),

  remember: ['foo']
}
</script>
```

---

I've been able to resolve the issue locally and will be submitting a PR shortly!

### LINKED ISSUE #432: Form helper inside Vue 3 Composition API cannot access methods
### Versions:

- `@inertiajs/inertia-vue3` version: 0.3.4

### Describe the problem:

When using the Vue 3 Composition API with the `@inertiajs/inertia-vue3` form helper, the methods of the form helper (`post()`, `submit()` etc.) only work when used inside the Options API or the template.

When using `setup` to expose your own method, an error is logged to the console, e.g.

```
Uncaught TypeError: form.post is not a function
``` 

### Steps to reproduce:

```js
<template>
    <form @submit.prevent="submit()">
        <button type="submit">Log in</button>
    </form>
</template>

<script>
import { useForm } from '@inertiajs/inertia-vue3'

export default {
  setup () {
    const form = useForm({
      email: null,
      password: null,
      remember: false,
    })

    function submit() {
        form.post('/');
    }
	
    return { form, submit }
  },
}
</script>

```

### Fix

This issue occurs because the [export](https://github.com/inertiajs/inertia/blob/896c53bb0e80bb6ba56e588a3afe665fa8aa0498/packages/inertia-vue3/src/form.js#L149) in uses `ref()` instead of `reactive()`.  I think `ref()` is more for primitive types.

```js
export function useForm(data) {
  return ref(form(data))
}
```

Changing it to `reactive()` should fix this issue:

```js
export function useForm(data) {
  return reactive(form(data))
}
```

### LINKED ISSUE #455: Inertia form helper cast attributes to string

### LINKED ISSUE #552: Vue3 form helper with 'remember' throws error if data contains array
### Versions:

- `@inertiajs/inertia` version: 0.8.5
- `@inertiajs/inertia-vue3` version: 0.3.5

### Describe the problem:

Using 'remember' with the form helper throws an error if the data contains an array.

```DOMException: Failed to execute 'replaceState' on 'History': [object Object] could not be cloned.```

### Steps to reproduce:

```js
export default {
  remember: 'form',
  data() {
    return {
      form: this.$inertia.form({
        name: '',
        items: [],
      }),
    }
  }
}
```
It works without the form helper. And it works in Vue2 with and without the form helper.

### LINKED ISSUE #553: The form does not reset when it contains an array of objects
### Versions:

- `@inertiajs/inertia` version: 0.8.5
- `@inertiajs/inertia-vue3` version: 0.3.5

### Describe the problem:

The form does not reset when it contains an array of objects.

### Steps to reproduce:

1. Create an instance of inertia form that has an array attribute

```
data() {
    return {
        form: this.$inertia.form({
              users: []
        });
    }
}
```

2. Then push some values into that array
```
this.form.users.push({id: 1, name: 'first'});
```

3. Then reset the form
```
this.form.reset();
```

The form will not reset and still contains the pushed items.

reinink:
This PR updates the form helper in all four adapters to standardize the "remember" behaviour. In particular, it updates the Vue 2 and Vue 3 adapters to automatically remember form data and errors, just like the React and Svelte adapters currently do. Previously in Vue 2 and Vue 3 you had to do this manually:

```js
// Object syntax
remember: 'form'

// Composition API
useRemember(useForm({ ... }))
```

Doing this automatically is not only easier, but it's also more in line with default browser behaviour.

## 🔧 Options

There is a new form options API, which allows you to set a form `key` (needed if there are multiple forms on the page), as well as a `remember` option to disable the automatic remember behaviour, if needed. Here's an example:

```js
// Vue 2
this.$inertia.form({
  email: null,
  password: null,
  remember: null,
}, {
  key: 'login',
  remember: false,
})

// Vue 3/React/Svelte
const form = useForm({
  email: null,
  password: null,
  remember: null,
}, {
  key: 'login',
  remember: false,
})
```

## ⚠️ Breaking changes

I made this change in a backwards compatible way, using a `__rememberable` marker in the form helper, which will prevent the form helper from being remembered twice.

Note, to make this work in Vue 2, I needed to use `Vue.observable()`, which was added to Vue 2 in version `2.6.0`. Meaning, if you're on an older version of Vue, you can no longer use the form helper.

If you're using the `useForm` hook in React, the (undocumented) second argument (`key`) has changed to the options object. Be sure to update the key to the new format.

```js
// Before
useForm(data, 'my-form')

// After
useForm(data, { key: 'my-form' })
```

## 🐞 Bug fixes

This also fixes numerous issues with the remember feature:

Closes #418
Closes #440
Closes #456
Fixes #417
Fixes #432
Fixes #455
Fixes #552
Fixes #553

This change allowed me to remove all the serializing and unserializing jazz we had to do previously to get things working properly in the Vue adapters.

## 📺 Demo

Here's a quick demo:

https://user-images.githubusercontent.com/882133/112546418-c7fcb800-8d8f-11eb-87fa-0d81082ce9dc.mov


iksaku (approved):
Loving that we've moved into a `reactive()` Form object :tada:

packages/inertia-vue3/src/useForm.js
```diff
@@ -146,9 +135,14 @@ export default function form(data = {}) {
     delete(url, options) {
       this.submit('delete', url, options)
     },
+  })
+
+  if (remember) {
+    watch(form, (value) => {
+      const raw = toRaw(unref(value))
+      Inertia.remember({ data: raw.data(), errors: raw.errors }, key)
```
RobertBoes, line 143:
```suggestion
      const keepRememberedKeys = obj => {
        Object
          .keys(obj)
          .reduce((carry, field) => ({
            ...carry,
            ...(fields.length > 0 && !remember.includes(field) ? {} : { [field] : obj[field] }),
          }), {})
      }

      Inertia.remember({
        data: keepRememberedKeys(raw.data()),
        errors: keepRememberedKeys(raw.errors)
      }, key)
```
Would something like this be beneficial? Don't know if it does work, but the idea is you can also pass `remember: ['email']` to the form, all other form fields won't be remembered

reinink: So I don't think I want users to have to specify which fields they want to remember. Generally speaking remembering the whole form is fine. Not to say we couldn't add more granularity like this in the future. I just think I want to start with a simple enable/disable option.

claudiodekker (approved):

PR #580, opened 2021-03-29
[Vue 3] Match Typescript definition with v0.3.6 update

iksaku:
The recent v0.3.6﻿ release of the Vue 3 adapter comes with an improved `useForm` and
`useRemember` functions, which were type-hinted before, but weren't updated for the corresponding release.

This PR tries to fix that, so that Typescript users stop getting warnings from their IDEs.

### DIFF
--- a/packages/inertia-vue3/index.d.ts
+++ b/packages/inertia-vue3/index.d.ts
@@ -63,8 +63,6 @@ interface Form<Data> {
   delete(url: string, options?: Inertia.VisitOptions): void
 }
 
-type FormWithData<Data> = Data & Form<Data>
-
 export const App: InertiaApp
 
 export const Link: InertiaLink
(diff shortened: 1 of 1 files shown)


reinink:
Thanks for this. Any chance you could update the [React](https://github.com/inertiajs/inertia/blob/master/packages/inertia-react/index.d.ts) and [Vue 2](https://github.com/inertiajs/inertia/blob/master/packages/inertia-vue/index.d.ts) versions at the same time?

iksaku:
@reinink Will take a look at it this afternoon :+1:

reinink:
Fair warning, we're working adding official TS support to Inertia very soon, so it may not be worth the effort. Might be better to just wait until we're done that. 🤷‍♂️

iksaku:
Alright, will wait then. If there's anything I can help with your TS development, I would gladly help with the Vue 3 adapter, and Vue 2 if needed :tada:

reinink:
Okay cool, I'm going to close this issue, as we're getting close on the TS update: #592

Feel free to join the discussion there! 👍

PR #588, opened 2021-03-30
Use function for layout child in Vue 3 adapter

### LINKED ISSUE #579: vue3 warn when using array for layout
### Versions:

- `@inertiajs/inertia` version: 0.8.6
- `@inertiajs/inertia-vue3` version: 0.3.6

```
[Vue warn]: Non-function value encountered for default slot. Prefer function slots for better performance. 
```

```vue
<template>
  <div>
    Home
  </div>
</template>

<script>
import Layout1 from "@/Layouts/Layout1";
import Layout2 from "@/Layouts/Layout2";

export default {
  layout: [Layout1, Layout2],
}
</script>
```

reinink:
Fixes #579

This update prevents Vue 3 from throwing a warning when setting an array of layouts.

### DIFF
--- a/packages/inertia-vue3/src/app.js
+++ b/packages/inertia-vue3/src/app.js
@@ -55,7 +55,7 @@ export default {
             return component.value.layout
               .concat(child)
               .reverse()
-              .reduce((child, layout) => h(layout, [child]))
+              .reduce((child, layout) => h(layout, () => child))
           }
 
           return h(component.value.layout, () => child)


jorgemurta:
@reinink just found this Merge Request after facing the same error following the documentation: https://inertiajs.com/pages#creating-layouts

If you look at the Persistent Layouts section under Vue 3 you'll see that there is documented the method that throws the warning in vue 3. I tried to update the documentation myself but after searching for a bit it feels that its not publicly available on a repo.

I hope I could be of help, and great job on Inertia!

reinink:
@joburgard Good catch, thanks for sharing this. I've just corrected this (it's deploying). 👍

joburgard:
> 
> 
> @joburgard Good catch, thanks for sharing this. I've just corrected this (it's deploying). 👍

I think you meant @jorgemurta 😄

reinink:
Haha, shoot, yup. I just picked the first username that came up assuming it was right. 😅

PR #591, opened 2021-03-30
Fix remember

iamgergo:
This is a follow-up for the PR https://github.com/inertiajs/inertia/pull/587.

Personally, I use a custom Form component, instead of using the built-in form helper.

My custom component has the `remember` property, which leads to the same issue that is solved for the built-in form helper.

With this change, it works well now. I guess this does not affect any other part of the code.


reinink:
Makes sense! Thanks for providing this improvement. 👍

PR #592, opened 2021-03-30
Convert @inertiajs/inertia to TypeScript

claudiodekker:
This PR sets up an initial attempt at converting Inertia's core to TypeScript, and serves as a placeholder as we work on this.

Right now, everything works as-is with the existing adapters, but I'm still not happy with the following things:
- The exact types (naming, composition)
- Certain parameters/types might need extracting into separate types for better clarity


innocenzi:
> The 'Proxy' class is horrible

I'm not 100% sure to understand its goal. If it is there in order to be able to import `Inertia` and keep doing `Inertia.init`, you could consider moving the constructor logic to an `init` method, and export a constant named `Inertia`.

```ts
// inertia.ts
export class Core {
	// ...
	constructor() {}

	init(options: Options) {
		// Initialization logic here
	}
	
	// ...
}

export const Inertia = new Core()
```

```ts
// In adapters
import { Inertia } from 'inertia'

Inertia.init({ /* ... */})
// ^ Inertia is a singleton of Core here
```

Alternatively, you could have an `init` named constructor, which would assign a new instance of Inertia to a file-scoped variable.

claudiodekker:
@innocenzi Right, but keep in mind things such as `$page` / `$inertia` should also keep on working, which I wasn't able to.
I'd love to see if you could solve it, though? Feel free to fork it & PR towards this branch so I can see what you're doing & even merge it in!

innocenzi:
> I'd love to see if you could solve it, though? Feel free to fork it & PR towards this branch so I can see what you're doing & even merge it in if it's better!

Done, it's in #600! I'd love to contribute some more if you need help somewhere.

claudiodekker:
@innocenzi Thanks a lot!

> I'd love to contribute some more if you need help somewhere.

By all means, if you think you can improve on this PR in any other way, such as by addressing the points I listed in the PR description, feel free to! 

Just keep in mind that if you decide on doing so, that we don't want to make _too big_ of a change (if at all) regarding the structure/logic flow for now, and we preferably also want to keep the adapters untouched/as-is. That way, the PR remains understandable/reviewable by both @reinink and I, while also preventing us from having to tag and release 4-5 breaking changes 👍

packages/inertia/src/inertia.ts
```diff
@@ -0,0 +1,529 @@
+import {
```
reinink, line 1: Rename this file to `router.ts`, and rename the class to `Router`, but still export it as `Inertia`.

PR #597, opened 2021-04-02
Disable automatic form helper remembering

reinink:
We recently updated the Vue 2 and Vue 3 form helpers to automatically remember form data and errors in history state. This was to match the behaviour of the React and Svelte adapters, which already did this.

While I really liked this behaviour, I've since come to believe that doing this automatically dramatically increases the chances of bugs in user land code. The issue is when you have more than one form on a page. In these situations you MUST provide a unique remember key to your form, otherwise Inertia doesn't know what saved form data to use from the history state. However, other than the docs, there is nothing pushing you to provide this key. Meaning, it's very easy to simply add a bunch of forms to a page, and suddenly things stop working, without any real explanation why.

I ran into this first hand when upgrading to the latest version of the Inertia Vue adapter in my own app. Suddenly a bunch of forms stopped working. I had thought that this update was done in a non-breaking way, but evidently I was wrong. And it makes sense. Automatically remembering the form data isn't enough. We also need a key. Otherwise, when forms save their data to the history state, the last form to make a change stomps over the data of any other forms. Then, when navigating history and the form data is restored, all the forms on the page all use the data from that last form, which creates a real mess with the form data, and often JavaScript errors.

This PR provides a patch, essentially reverting these changes. By default, form data and errors are *no longer* automatically saved to the history state.

This PR also includes some form helper API changes to improve things moving forward. I've removed the form helper options as the second argument, which were arguably kind of ugly. Instead, now you can provide a string as the first argument, that will be used as the remember key, and the form data will be remembered without conflicts (assuming you provide a unique key of course). Omitting it will disable the remember behaviour.

```js
// Vue 2/Vue 3 (options API)
this.inertia.$form(data)                    // Not remembered
this.inertia.$form('CreateContact', data)   // Remembered

// Vue 3/React/Svelte
let form = useForm(data)                    // Not remembered
let form = useForm('CreateContact', data)   // Remembered
```

Note, this also updates the Svelte and React adapters to work in the same way. Moving forward, they will NOT remember form data by default either, as they suffer from the exact same issues.

Overall, while I'm disappointed that there isn't a path forward to have form data automatically saved, I do think this is a good change, that's going to improve DX moving forward. And, in reality, it's not uncommon for libraries like Vue and React to require keys, so requiring this in Inertia isn't all that odd.


packages/inertia-vue/src/form.js
```diff
@@ -136,13 +137,23 @@ export default function(data = {}, { key = 'form', remember = true } = {}) {
...
     },
+    __rememberable: rememberKey === null,
+    __remember() {
+      return { data: this.data(), errors: this.errors }
+    },
+    __restore(restored) {
+      Object.assign(this, restored.data)
+      Object.assign(this.errors, restored.errors)
+      this.hasErrors = Object.keys(this.errors).length > 0
+      return this
```
claudiodekker, line 148:
```suggestion
```

claudiodekker, line 148:
```suggestion
      this.hasErrors = Object.keys(this.errors).length > 0
```

claudiodekker, line 148:
```suggestion
```

packages/inertia-vue3/src/useForm.js
```diff
@@ -136,12 +137,21 @@ export default function useForm(data = {}, { key = 'form', remember = true } = {
...
     },
+    __rememberable: rememberKey === null,
+    __remember() {
+      return { data: this.data(), errors: this.errors }
+    },
+    __restore(restored) {
+      Object.assign(this, restored.data)
+      Object.assign(this.errors, restored.errors)
+      this.hasErrors = Object.keys(this.errors).length > 0
+      return this
```
reinink, line 148:
```suggestion
```

claudiodekker:
Hey @sebastiaanluca,

So you're saying that an anonymous form (`this.$inertia.form({ data: value })`) doesn't populate the errors?
If that's the case, can you please open this as a separate issue (that ideally references this comment)?

Thanks!

sebastiaanluca:
@claudiodekker That was seemingly the case at the time, but I remember fixing it or that I was using it incorrectly. Don't ask me how though, can't remember 😄 So not an issue IIRC.

PR #599, opened 2021-04-02
React HMR fix

sebastiandedeyne:
While the `useEffect` dependencies are correct in theory, they're a pain for hot module replacement with React.

When HMR kicks in, the effect gets cleaned up and `Inertia.init` gets called again. This causes the initial page to be rendered instead of the current one.

There might be a cleaner way to solve this along the road, but I can't think of any better, and don't see any practical downsides of this change.

### DIFF
--- a/packages/inertia-react/src/App.js
+++ b/packages/inertia-react/src/App.js
@@ -29,7 +29,7 @@ export default function App({
         }))
       },
     })
-  }, [initialPage, resolveComponent, resolveErrors, transformProps])
+  }, [])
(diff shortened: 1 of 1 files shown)


reinink:
Yeah, this is 100% okay. I probably just included them because my editor was yelling at me to add them. In reality, none of these values are going to ever change, so we can safely omit them.

reinink:
Released: https://inertiajs.com/releases/inertia-react-0.5.8

PR #600, opened 2021-04-02
refactor: remove the need for inertia-proxy.ts

innocenzi:
This PR renames the `Inertia` class as `Core` in order to be able to export an `Inertia` constant (which acts as a singleton). 

Additionally, the initialization logic is moved to `init` to avoid any breaking change. As a drawback, the non-null operator has been added to a few properties of the class. This should be avoided: ideally, a check is performed in order to ensure that Inertia has been initialized, and these properties are marked as possibly null. But this is not a big deal here (no-one uses Inertia as-is anyway), just something to keep in mind.

This effectively allows for the removal of `inertia-proxy.ts`.

I tested that change with the PingCRM app, everything was working as expected.

cc @claudiodekker


claudiodekker:
Thanks for helping out here @innocenzi! 👍

PR #602, opened 2021-04-03
Make Page props available in Persistent Layouts

claudiodekker:
This PR adds the ability to use a Persistent Layout's props to directly reference the currently-viewing page's props, without having to rely on the `$page.props` Inertia plugin.

## Example
```vue
<template>
  <div>
    This is my persistent Layout
    Example prop: {{ foo }}
    <slot />
  </div>
</template>

<script>
export default {
  props: {
    foo: String
  }
}
</script>
```

Previously, you had to do something like this in order to achieve the same goal:
Example:
```vue
<template>
  <div>
    This is my persistent Layout
    Example prop: {{ foo }}
    Another way to access the same prop: {{ $inertia.page.props.foo }}
    <slot />
  </div>
</template>

<script>
export default {
  computed: {
    foo() {
      return this.$inertia.page.props.foo
    }
  }
}
</script>
```

## Want to contribute? Here's some TODO's:
If you want to pick any of these up, please let me know so I can assign you to the task!
- [x] Add React implementation
- [x] Add Svelte implementation


reinink:
I've updated the React adapter to do this. It was really only needed when using the array shorthand:

```js
Dashboard.layout = [Layout]
```

When using a render function, this was already possible by manually passing along the page props:

```diff
- Dashboard.layout = page => <Layout>{page}</Layout>
+ Dashboard.layout = page => <Layout {...page.props}>{page}</Layout>
```

@sebastiandedeyne I'd love your input here, if there is a better way to handle either of these situations. 👍

reinink:
And I've updated the Svelte adapter to support this as well.

@mariojankovic If love your eyes on this. 👍

mariojankovic:
Looks good to me!

reinink:
Thanks @mariojankovic!

Issue #447, opened 2021-01-19
Provide "modern" microbundle build


jalt007:
👋

To get inertia working with native es module build tools such as [Vite](https://github.com/vitejs/vite), I had to add a "modern" build for all the packages.

- https://github.com/developit/microbundle/#-modern-mode-

This involves adding `"exports": "dist/index.modern.js",` & e.g. `"build:modern": "microbundle --format modern",` for all packages `package.json`.

An easy PR if need be, but maybe there are other things I haven't considered?

claudiodekker:
Ooh, it would be really cool to support Vite, and since Inertia only supports modern browsers anyway I don't think this is an unreasonable thing to add (especially since it doesn't bring any maintenance downsides from what I can tell?).

What do you think, @reinink?

jalt007:
Yeah i couldn't immediately see a maintenance burden because microbundle handles everything nicely!

mindreframer:
Any love for this issue? It seems to be a rather non-intrusive change and getting it working with Vite would be a real boost, since it is growing in popularity in the JS community.

reinink:
Seems reasonable to me. 👍

@sebastiandedeyne, any thoughts on this, as our resident publishing and Vite expert?

mindreframer:
@reinink Actually, I got this to compile properly. I needed to add the `util` package, to polyfill the Node util package for vite. That was everything needed, it worked without the "modern" microbundle build. 🤷  

So, I am not sure if we need to do something, besides maybe documenting the extra step with the `util` package in combination with Vite.js

sebastiandedeyne:
You can use Inertia with Vite without any additional configuration. 

@mindreframer are you sure the `util` package is Inertia-specific? I can't find any references in this code base.

That said, I do believe we should add an ESM build to Inertia. I've opened #610 to continue this discussion.

mindreframer:
@sebastiandedeyne 

By using `yarn list|grep ...` this is what I have: 


```
├─ @inertiajs/inertia@0.8.6
│  ├─ axios@^0.21.1
│  ├─ deepmerge@^4.0.0
│  └─ qs@^6.9.0




├─ qs@6.10.1
│  └─ side-channel@^1.0.4


├─ side-channel@1.0.4
│  ├─ call-bind@^1.0.0
│  ├─ get-intrinsic@^1.0.2
│  └─ object-inspect@^1.9.0

````

- https://github.com/inspect-js/object-inspect/issues/38
  

```
[...]

PR #610, opened 2021-04-06
Update build configuration

sebastiandedeyne:
_This is a draft PR and currently only targets the main Inertia package for the sake of discussion. I'll update the other build configurations after we've drawn our conclusions._

Triggered by the discussion at #447 I decided to look into the current build situation and see if we can do better. It's been over a year since this was all set up, and with JavaScript tools moving forward it doesn't hurt to take a step back every now and then.

**Tldr;** we should support ESM in addition to CJS and UMD.

---

My dream situation would be a single `"build": "microbundle"` script without any additional configuration, but alas I don't think that's a fit for Inertia.

## Current status

We currently ship two builds: 

- **CJS** (CommonJS): supported in bundlers like Webpack, Vite, or Rollup
- **UMD**: supported in in browsers

UMD builds use the `--external none` setting. This ensures that the package dependencies (`axios`, `deepmerge` & `qs`) are bundled with the build. When someone loads the file straight in their browser, they don't need to add additional dependencies. The CommonJS build doesn't need this, because bundlers take care of this at compile time.

## Supporting ESM

Fast forward since this was first set up, and there's new relevant module format on the block: **ESM**. ESM is the one-standard-to-rule-them-all system that tools like Webpack and Vite are using by default now. The benefit of ESM is that it should work everywhere (or will at some point) and is tree-shakable. Build tools are able to detect unused code and slim down the module.

_Side note on Vite: Vite uses ESM as the default format. Anything non-ESM needs to be transformed before it can run. The more dependencies ship ESM, the less work Vite has._

The `--external none` thing is still an issue though: we don't want to ship dependencies to bundlers, but we want them readily available for browsers. Microbundle has a new-ish "modern" build mode which generates an ESM bundle specifically for browsers. Similar to UMD, we'd add the `--external none` flag for the browser build.

That brings us to this updated configuration:

```diff
  {
    "scripts": {
-     "build": "npm run build:cjs && npm run build:umd",    
+     "build": "npm run build:bundler && npm run build:browser",
-     "build:cjs": "microbundle --format cjs",
+     "build:bundler": "microbundle --format es,cjs",
-     "build:umd": "microbundle --format umd --name Inertia --external none",
+     "build:browser": "microbundle --format modern,umd --name Inertia --external none"
    }
  }
```

(I don't like the `build:bundler` script name but haven't thought of anything better yet)

## Do browser builds make sense?

Final question: do browser builds actually make sense for Inertia? While technically possible, I haven't come across anyone using Inertia without a bundler. Is it worth supporting browser builds?

## Resources

- [Difference between the formats](https://dev.to/iggredible/what-the-heck-are-cjs-amd-umd-and-esm-ikm)
- [Related Microbundle discussion](https://github.com/developit/microbundle/issues/633#issuecomment-730164690)
- [How Microbundle decides which dependencies to bundle](https://github.com/developit/microbundle/wiki/How-Microbundle-decides-which-dependencies-to-bundle)


claudiodekker:
> do browser builds actually make sense for Inertia? 

To me it does, at least for local development. In prod. I'd definitely use a bundled/non-browser build, but if they're hot-swappable that totally shouldn't make any difference.

sebastiandedeyne:
Then we'll definitely keep those!

garlandcrow:
@sebastiandedeyne I was playing around for the past few days trying to get Vite + InertiaJS + Svelte to work. Initial attempt was to do the same thing as a few people had done to get Vue working, the critical Vite config change needed for Vue is 
```js
optimizeDeps: {
    include: [
      "@inertiajs/inertia",
      "@inertiajs/inertia-vue3",
    ],
```
But when switching everything to the Svelte equivalents, this gets things compiling but using shared props (and probably other things) do not work:
```js
import { page } from "@inertiajs/inertia-svelte";
console.log($page)
```
This logs an empty object and doesn't observe any changes (@robert on the discord confirmed that there seems to be 2 stores created and thats why there is no update). So I tried out using the config file in this PR and removing the `optimizeDeps` key and found that solves the issue. So for InertiaJS + Svelte to work in Vite, this config change with the ESM build is required. 

So hopefully we can get this one merged in the next release 😄   Let me know if any questions.

claudiodekker:
@reinink thoughts? Personally, I'd love to see this move forward before eventually tagging 1.0.

claudiodekker (approved):

reinink:

thecrypticace:
FYI, v1.0 shipped with ESM and CJS builds when we swapped out microbundle for esbuild.

PR #611, opened 2021-04-06
fix InertiaLinkProps typing

AdrianMrn:
The way `InertiaLinkProps` is currently typed, causes `data` and `headers` to not accept any value, as it's typed as `string` in `React.AllHTMLAttributes`, and as `object` in `InertiaLinkProps`.

As is already done for `onProgress`, we should omit all the props Inertia is using, to avoid any other conflicts in the typing.

Credits to @sebastiandedeyne for the idea :)

### DIFF
--- a/packages/inertia-react/index.d.ts
+++ b/packages/inertia-react/index.d.ts
@@ -41,7 +41,7 @@ interface BaseInertiaLinkProps {
   onSuccess?: () => void
 }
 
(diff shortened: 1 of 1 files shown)


sebastiandedeyne (approved):

reinink:
Thanks!

AdrianMrn:
Thanks for the quick merge!

AdrianMrn:
Hey @reinink, is it possible that the latest release (0.5.10) does not include the changes? I'm still seeing the old code after upgrading 😬

sebastiandedeyne:
Republished as 0.5.11 :)

PR #614, opened 2021-04-08
Update vue3 adapter type definition file

ycs77:
Some change is refer to [Inertia React Adapter's type definition file](https://github.com/inertiajs/inertia/blob/master/packages/inertia-react/index.d.ts)

Update list:
* Rename Form type to `InertiaForm` and `InertiaFormProps`
* Rename `Data` type to `TForm`
* Change `errors` property type to `Record`
* Remove `serialize()` and `unserialize()` methods
* Add `useForm()` overload
* Code formatting


reinink:
Thanks!

PR #621, opened 2021-04-12
Fix 'remember' issue for null values

### LINKED ISSUE #620: Issue using 'remember' with null values (regression)
### Versions:

- `@inertiajs/inertia` version: `v0.8.6`
- `@inertiajs/inertia-vue` version: `v0.5.10`

### Describe the problem:

It seems that the [previously fixed issue](https://github.com/inertiajs/inertia/issues/417) when using a `null` value with `remember` is broken again. I just re-verified in my app that in `v0.5.6` it's fixed and after `v0.5.7` it's broken.

### Steps to reproduce:

The below example should allow you to easily recreate the issue:

```
<script>
export default {
  data: () => ({
    foo: null,
  }),

  remember: ['foo']
}
</script>
```

--- 

I'll be submitting a PR to fix it in a moment!

aryehraber:
This PR fixes #620 by checking whether a value is `null` before the `typeof this[key].__remember === 'function'` check. After making this change, the remember functionality started working as expected again.

Quick note: I had a look around the codebase to see if the optional chaining operator (`?.`) syntax was used anywhere but couldn't see it so didn't use it here. The syntax could however be cleaned up a bit by using:

```js
const hasCallacks = typeof this[key]?.__remember === 'function' && typeof this[key]?.__restore === 'function'
```

Let me know if there are any questions or if any changes are needed.

### DIFF
--- a/packages/inertia-vue/src/remember.js
+++ b/packages/inertia-vue/src/remember.js
@@ -29,7 +29,7 @@ export default {
     })
 
     rememberable.forEach(key => {
-      const hasCallacks = typeof this[key].__remember === 'function' && typeof this[key].__restore === 'function'
+      const hasCallacks = this[key] !== null && typeof this[key].__remember === 'function' && typeof this[key].__restore === 'function'
 
       if (this[key] !== undefined && restored !== undefined && restored[key] !== undefined) {
         hasCallacks ? this[key].__restore(restored[key]) : (this[key] = restored[key])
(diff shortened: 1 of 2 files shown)


reinink:
Hey thanks so much for this @aryehraber! The remember functionality has been a headache of recent...wow.

I've made some other changes to this PR, to further improve this fix, and also add the Vue 3 adapter fix.

While working on that I found another bug with the logic used to determine if there are remember/restore callbacks, and have fixed that as well.

Hopefully this is all taken care of now! 😅

PR #658, opened 2021-05-05
Added react hooks to shortcut Inertia.on

kioltk:
Basically made hooks that covers Inertia.on events. So instead of 
```js
import { useEffect } from 'react';
import { Inertia } from '@inertiajs/inertia';

useEffect(()=>{
   return Inertia.on('start', () => {
      // effect
   });
}, []);
```
You can use

```js
import { useStart } from '@inertiajs/inertia-react';

useStart(()=>{
   // effect
});
```

### DIFF
--- a/packages/inertia-react/src/hooks.js
+++ b/packages/inertia-react/src/hooks.js
@@ -1,0 +1,50 @@
+import { useEffect } from 'react';
+import { Inertia } from '@inertiajs/inertia';
+
+export function useStart(callback) {
+  useEffect(() => {
+    return Inertia.on('start', callback);
+  }, []);
+}
+
+export function useSuccess(callback) {
+  useEffect(() => {
+    return Inertia.on('success', callback);
+  }, []);
+}
+
+export function useBefore(callback) {
+  useEffect(() => {
+    return Inertia.on('before', callback);
+  }, []);
+}
+
+export function useError(callback) {
+  useEffect(() => {
+    return Inertia.on('error', callback);
+  }, []);
+}
+
+export function useFinish(callback) {
+  useEffect(() => {
+    return Inertia.on('finish', callback);
+  }, []);
+}
+
+export function useInvalid(callback) {
+  useEffect(() => {
+    return Inertia.on('invalid', callback);
+  }, []);
+}
+
+export function useProgress(callback) {
+  useEffect(() => {
+    return Inertia.on('progress', callback);
+  }, []);
+}
+
+export function useNavigate(callback) {
+  useEffect(() => {
+    return Inertia.on('navigate', callback);
+  }, []);
+}


reinink:
Hey thanks for your interest in Inertia.js! 🙌

To be honest, I don't really feel like this change is necessary. That said, let's see what @sebastiandedeyne, our React adapter maintainer, says. 👍

sebastiandedeyne:
Hmm I'm not convinced yet. Note that by hardcoding the `useEffect` dependencies to `[]`, you could be introducing stale closure bugs.

```js
const [foo, setFoo] = useState('foo');

useSuccess(() => {
  console.log(foo);
});
```

I believe `foo` will always be the initial value `'foo'` in this case.

The only way to solve this is to require dependencies like `useEffect`.

```js
useSuccess(() => {
  console.log(foo);
}, [foo]);
```

But at this point, the `useEffect` abstraction is so leaky I'd rather keep close to React.

Besides these issues, I don't think the names of these hooks look like event handlers. I wonder how other libraries expose events through hooks? (Are there any that do?)

reinink:
Hey thanks for this feedback @sebastiandedeyne. I think I'm going to close this, as I don't think it's necessary. Thanks either way @kioltk. 👍

kioltk:
> Hmm I'm not convinced yet. Note that by hardcoding the `useEffect` dependencies to `[]`, you could be introducing stale closure bugs.
> 
> ```js
> const [foo, setFoo] = useState('foo');
> 
> useSuccess(() => {
>   console.log(foo);
> });
> ```
> 
> I believe `foo` will always be the initial value `'foo'` in this case.
> 
> The only way to solve this is to require dependencies like `useEffect`.
> 
> ```js
> useSuccess(() => {
>   console.log(foo);
> }, [foo]);
> ```
> 
> But at this point, the `useEffect` abstraction is so leaky I'd rather keep close to React.

I understand how `useEffect` work, but the only reason I made the pull request is to shorthand the bolierplate of creating `useEffect` every time. However your assumption about `'foo'` is wrong, cause if `foo` is changed before `useSuccess` triggered, callback will still use an updated value. That's maybe how Java or PHP scopes work, by finalizing the passed variable and sticking it inside the scope forever, however JavaScript does not finalize references variables. Your proposal by adding dependencies to these hooks makes sense in case there is some action we want to make _every time_ the value updates. Like `useRemember` somehow processes new values. but recreating the event listener will not do a work, cause onSuccess will be called only when `Inertia.event('success')` triggered and and triggered only once.

> Besides these issues, I don't think the names of these hooks look like event handlers. I wonder how other libraries expose events through hooks? (Are there any that do?)

Another example of hooks is gorgeous library [react-use](https://github.com/streamich/react-use), there is [useKeyPressedEvent](https://github.com/streamich/react-use/blob/master/docs/useKeyPressEvent.md). Well, while using babel you can name your imports and make it like, so it makes sense to keep naming as close to functionality it does as possible.
```js 
import { useSuccess as useInertiaSuccess } from 'inertiajs/react'
```
Still it maybe has sense to call it `useSuccessEvent` and etc.

sebastiandedeyne:
I created a CodePen to illustrate the issue I'm talking about: https://codepen.io/sebdd/pen/ExWoXyB?editors=1011

Clicking the randomize button randomizes the view, but the original piece of state is trapped in the effect closure.

> onSuccess will be called only when Inertia.event('success') triggered and and triggered only once.

I don't think this is true. If an `onSuccess` callback is used in a component that never gets removed from the tree (like a persistent layout), it might be called multiple times.

Issue #649, opened 2021-04-24
VisitOptions missing onBefore TS definition?


devpake:
https://github.com/inertiajs/inertia/blob/653ccdda512bdc73ce0491103f9ade0b803cdbb0/packages/inertia/index.d.ts#L27

`onBefore` does not seem to be defined within `VisitOptions` in the definition file but it is referenced in the docs [https://inertiajs.com/manual-visits](https://inertiajs.com/manual-visits).  Should it be added or is `onBefore` being deprecated?

reinink:
Inertia core has now been updated to TypeScript, so this is no longer an issue. 👍

PR #662, opened 2021-05-07
Update core type definition file

ycs77:
* Use the `Record` utility type
* Remove space
* Add the missing visit options


ycs77:
Fixed #649

PR #668, opened 2021-05-11
Extend hasFiles to support FormData

kasperstyre:
### Overview
This change makes the `hasFiles` function able to determine if files are included in a `FormData` object.

### Reasoning
I'm currently developing a project that uses [an unofficial port of Inertia](https://github.com/Nothing-Works/inertia-aspnetcore) in an ASP.NET Core environment. To make sure file lists are correctly translated into view models, i need to make sure they are added with the same name without indices attached to the name, as illustrated below:
```js
const formData = new FormData();

formData.append("files", new File([], "file1"));
formData.append("files", new File([], "file2"));
```
In the current version of Inertia, i *can* add an object like the above as a parameter to a visit, but due to Inertia not knowing there are files included in the form, no progress events are being fired, preventing me from displaying a progress bar in my UI.

### DIFF
--- a/packages/inertia/src/inertia.js
+++ b/packages/inertia/src/inertia.js
@@ -201,7 +201,7 @@ export default {
     [url, data] = mergeDataIntoQueryString(method, hrefToUrl(url), data)
 
     const visitHasFiles = hasFiles(data)
-    if (method !== 'get' && (visitHasFiles || forceFormData)) {
+    if (method !== 'get' && (visitHasFiles || forceFormData) && !(data instanceof FormData)) {
       data = objectToFormData(data)
     }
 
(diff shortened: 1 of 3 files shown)


packages/inertia/src/inertia.js
```diff
@@ -201,7 +201,7 @@ export default {
     [url, data] = mergeDataIntoQueryString(method, hrefToUrl(url), data)
 
     const visitHasFiles = hasFiles(data)
-    if (method !== 'get' && (visitHasFiles || forceFormData)) {
+    if (method !== 'get' && (visitHasFiles || forceFormData) && !(data instanceof FormData)) {
       data = objectToFormData(data)
```
reinink, line 205: I think it would be better to do the check if it's already a `FormData` object within the `objectToFormData` method.

reinink:
Thanks @kasperstyre!

PR #670, opened 2021-05-14
End processing in the onFinish callback

grantholle:
It makes sense to me that processing would end regardless of a successful, canceled, or errored request. Right now if my server throws a 500, the form is still "processing" even though it's not _really_ processing anything. The state of my form is now stuck processing.

In my case I have a loading state. An exception gets thrown and I can display a nice error, but the form state doesn't reset. If all roads lead to `onFinish` it makes sense to return `processing` to false in that callback.

### DIFF
--- a/packages/inertia-react/src/useForm.js
+++ b/packages/inertia-react/src/useForm.js
@@ -51,7 +51,6 @@ export default function useForm(...args) {
           }
         },
         onSuccess: (page) => {
-          setProcessing(false)
           setProgress(null)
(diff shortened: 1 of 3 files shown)


reinink:
Good catch, but we've actually designed it this way intentionally. See this PR: #635

I am working on a fix for this, which will solve this issue, while still taking into consideration the issue in that PR. Stay tuned for that. 👍 

(Also, this PR is missing the Vue 2 implementation.)

Issue #656, opened 2021-05-02
Can't perform a React state update on an unmounted component when performing a http request with useForm methods


amiranagram:
### Versions:

- `@inertiajs/inertia` version: 0.8.7
- `@inertiajs/inertia-react` version: 0.5.12

### Describe the problem:

```
Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

I'm a noobie with react (coming from vue) so bear with me. After some digging I found out that this error happens when trying to modify the state inside a component after the component has been unmounted. In my case after submitting a post request with `post` method from Inertia's `useForm`.

First I thought I messed up something (react noob) then I cloned [pingcrm-react](https://github.com/Landish/pingcrm-react) and the same thing happens there.

### Steps to reproduce:

Have a form that and use `useForm` to manage the state and submit it.

Ex:

```
const {data, setData, post, processing} = useForm({
    email: '',
    password: '',
    remember: false,
});

function submit(e) {
    e.preventDefault();

    post(route('login'));
}
```

Or you can clone [pingcrm-react](https://github.com/Landish/pingcrm-react), set up the project and try to login. After successful login when it navigates away from the login page you'll see the error in console.

Hope I was clear enough.

mrtijn:
Got the same issue. Using `Inertia.post` instead of the `post` from `useForm` seems to fix it.

reinink:
This has been fixed in #680 and will be released soon (hopefully tomorrow). 👍

PR #680, opened 2021-05-17
React: Fix "Can't perform a React state update on an unmounted component" error

vuxorpadd:
A fix for a #656 issue that happens when you redirect away from a form page after a successful form action.

### DIFF
--- a/packages/inertia-react/src/useForm.js
+++ b/packages/inertia-react/src/useForm.js
@@ -1,8 +1,19 @@
 import { Inertia } from '@inertiajs/inertia'
-import { useCallback, useRef, useState } from 'react'
+import {useCallback, useEffect, useRef, useState} from 'react'
 import useRemember from './useRemember'
 
 export default function useForm(...args) {
+
+  const isMounted = useRef(null)
+
+  useEffect(() => {
+    isMounted.current = true
+
+    return () => {
+      isMounted.current = false
+    }
+  }, [])
+
   const rememberKey = typeof args[0] === 'string' ? args[0] : null
   const defaults = (typeof args[0] === 'string' ? args[1] : args[0]) || {}
   const cancelToken = useRef(null)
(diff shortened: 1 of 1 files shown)


reinink:
@sebastiandedeyne I've been seeing this warning in the React adapter even when not using the form helper, when navigating between pages after a form submission. Should we be doing something differently in the React adapter to help prevent these warnings?

To be honest, in Vue, when an old page component is unmounted, you can still make changes to it, and Vue just gracefully ignores it and doesn't throw any warnings. And, I suspect React does the same when in production. So, is this even an issue?

packages/inertia-react/src/useForm.js
```diff
@@ -57,7 +68,13 @@ export default function useForm(...args) {
           setHasErrors(false)
           setWasSuccessful(true)
           setRecentlySuccessful(true)
-          recentlySuccessfulTimeoutId.current = setTimeout(() => setRecentlySuccessful(false), 2000)
+          recentlySuccessfulTimeoutId.current = setTimeout(() => {
+            if (!isMounted.current) {
+              return true
```
reinink, line 73: Curious, why are you returning `true` here?

reinink:
I have made some improvements to this PR, to account for other situations where updating the form component could result in this warning. 👍

sebastiandedeyne:
> And, I suspect React does the same when in production. So, is this even an issue?

This is one of these React things where the authors say it's not an issue today, but might be an issue in the future. So no, in production this doesn't matter much, but the warnings drive me crazy (especially since you don't want the m mixed with your _application's_ warnings) so I think it's worth getting rid of them!

reinink:
@sebastiandedeyne Thanks, that's kind of what I figured, which is why I tweaked this PR and merged it in.

Thanks @vuxorpadd for submitting this. 👍

packages/inertia-react/src/useForm.js
vuxorpadd: Well, the reasoning was: `setTimeout` did its job even if it did nothing, so `return true` :) But it could have been
```
if (isMounted.current) {
	setRecentlySuccessful(false), 2000)
}
```
instead

reinink: Understood. You'll notice I actually wrapped a lot more code within the `isMounted.current` check, since there are other situations that could result in this warning. 👍

vuxorpadd: Yep, saw that. I wasn't sure about those, so only fixed the thing I was able to reproduce. Thanks for your work :+1:

PR #688, opened 2021-05-22
Add head component to React adapter

reinink:
This PR adds the new `<InertiaHead>` component to the React adapter.

```js
import { InertiaHead } from '@inertiajs/inertia-react'

<InertiaHead>
  <title>Dashboard</title>
  <meta type="description" content="Example page description..." />
</InertiaHead>
```

If you only want to set the page title, you can also use this shorthand:

```js
import { InertiaHead } from '@inertiajs/inertia-react'

<InertiaHead title="Dashboard" />
```

And if terseness is your thing, the head component has also been exported as just `Head`:

```js
import { Head } from '@inertiajs/inertia-react'

<Head title="Dashboard" />
```

This also includes SSR support for the head component. This is done via a new `onHeadUpdate() callback prop that you can use within the SSR server. Something like this:

```js
server.post('/render', async (request, response, next) => {
  try {
    let head = []
    const html = ReactDOMServer.renderToString(
      <InertiaApp
        initialPage={request.body}
        resolveComponent={name => require(`${__dirname}/Pages/${name}`).default}
        onHeadUpdate={updated => (head = updated)}
      />,
    )
    response.end(`${head.join("\n")}<!--INERTIA_SSR_BOUNDARY-->${html}`)
  } catch (error) {
    next(error)
  }
})
```


packages/inertia-react/src/Head.js
```diff
@@ -0,0 +1,79 @@
+import React, { useContext, useEffect, useMemo } from 'react'
+import HeadContext from './HeadContext'
+
+export default function InertiaLink({ children, title }) {
```
claudiodekker, line 4:
```suggestion
export default function InertiaHead({ children, title }) {
```

claudiodekker (approved):

PR #692, opened 2021-05-25
Unify `onHeadUpdate` prop across adapters

reinink:
This PR updates the Vue 2, Vue 3 and React adapters to all use an `onHeadUpdate()` prop for handling `<head>` updates in SSR mode. It also tweaks the `createHeadManager()` method to require an `onUpdate()` callback, where previously this was set using `headManager.onUpdate(callback)`.

### DIFF
--- a/packages/inertia-react/src/App.js
+++ b/packages/inertia-react/src/App.js
@@ -18,12 +18,7 @@ export default function App({
   })
 
   const headManager = useMemo(() => {
-    const isServer = typeof window === 'undefined'
-    const headManager = createHeadManager(isServer)
-    if (isServer && onHeadUpdate) {
-      headManager.onUpdate(onHeadUpdate)
-    }
-    return headManager
+    return createHeadManager(typeof window === 'undefined', onHeadUpdate || (() => {}))
   }, [])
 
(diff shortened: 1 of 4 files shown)


packages/inertia-react/src/App.js
```diff
@@ -18,12 +18,7 @@ export default function App({
   })
 
   const headManager = useMemo(() => {
-    const isServer = typeof window === 'undefined'
-    const headManager = createHeadManager(isServer)
-    if (isServer && onHeadUpdate) {
-      headManager.onUpdate(onHeadUpdate)
-    }
-    return headManager
+    return createHeadManager(typeof window === 'undefined', onHeadUpdate || (() => {}))
```
claudiodekker, line 21: Maybe we can split this into multiple lines, so that it's easier to visually see that `x || y` second argument?

claudiodekker (commented):
LGTM 👍

packages/inertia-react/src/App.js
reinink: Hah, I had it that way, and then switched it back. But yeah, you're probably right. I'll make that change elsewhere.

PR #693, opened 2021-05-25
Remove `transformProps()` and `resolveErrors()`

### LINKED ISSUE #343: router/navigate/history: props don't transform when go back or forward

### LINKED ISSUE #598: Existing/unchanged props should not be cloned during partial reloads
### Versions:

- `@inertiajs/inertia` version: 0.8.6

### Describe the problem:

I've been debugging some infinite render/watch loops which started when we updated Inertia to the latest version.

I've boiled it down to the following scenario

- Page loads initially 👌 
- Stuff happens in the ui 👆 
- `Inertia.reload({only: 'foo'})` ♻️ 
- `visit` resolves and `this.page.props` is merged with the updated props 👌 
- `setPage` clones **all** props 🙅 
- All non-primitive props trigger updates in the component since they receive new values 🙅 
- Bad things happen 😭 

I believe the correct behavior is that _only_ the `only` props should be cloned and passed to `transformProps` and the rest should be merged as-is (by reference) since they have already been cloned and transformed previously. This will avoid any unnecessary and unpredictable updates from occurring.

reinink:
Closes #343
Closes #347
Closes #598

This PR removes the `transformProps` and `resolveErrors()` props from the Inertia adapters, and subsequently from the Inertia router (core). There are three reasons for this change:

1. These features are difficult to support in SSR (upcoming)
2. The transform props feature has been wrought with issues from the start (and is still broken)
3. We want to simplify the API (much of this work can be done server-side)

## Transforming props

The transform props feature has been wrought with issues from the start (#278, #295, #297, #343, #347, #598), largely due to the limitations with what kind of data you're able to store in history state. The original goal with this feature was to allow you to "enhance" the props before they are sent to a page component. For example, converting an `errors` prop to an `Errors` class. For example:

```js
transformProps: props => {
  return {
    ...props,
    errors: new Errors(props.errors),
  }
},
```

However (besides the fact that Inertia now has first-class error prop support) the reality is that doing this has caused more issues than it's solved. If you'd like to "enhance" your props in this way, I'd recommend using a helper function within your page components, or even a global computed property, if you're using Vue. Something like this:

```js
Vue.mixin({
  computed: {
    errors() {
      return new Errors(this.$page.props.errors || {})
    }
  },
})
```

## Resolving errors

We originally added the `resolveErrors()` prop for situations where your application shares errors using a different prop than `errors`. Inertia needs to know the errors prop to determine if a request has succeeded or failed.

In reality, there isn't any need for us to support an alternative prop than `errors`. If your server-side application doesn't currently send errors via the `errors` prop, we recommend simply updating your app to do this. Since you control what data is sent to Inertia, this shouldn't ever be an issue.

Think of `page.props.errors` as part of the [Inertia protocol](https://inertiajs.com/the-protocol).


claudiodekker (approved):

albertpratomo:
@reinink @claudiodekker Help needed:

Previously with `transformProps` I was able to enhance some props to be global `Vue.prototype`.

Now I can do this during `setup` in `createInertiaApp()`, but that global prototype is not updated when backend redirect to different page.

Where would you suggest to do enhancing some props to be global Vue property?

Much thanks before..

reinink:
> Previously with `transformProps` I was able to enhance some props to be global `Vue.prototype`.

I honestly don't even know what this means. 😅 

I think my recommendation would be to not do this, and find another way?

albertpratomo:
@reinink Instead of `this.$page.props.$me`, I want to be able to do `this.$me` instead.

At the end I use a global mixin for this:
```
Vue.mixin({
    beforeCreate() {
        if (!this.$page) return;

        Object
            .keys(this.$page.props)
            .filter(key => key[0] === '$')
            .forEach(key => {
                Object.defineProperty(this, key, {
                    get: function() {
                        return this.$page.props[key];
                    },
                });
            });
    },
});
```

reinink:
@albertpratomo You could use a global mixin for this instead. 👍

PR #697, opened 2021-05-27
Initial component

### LINKED ISSUE #696: Code splitting with react isn't working
### Versions:

- `@inertiajs/inertia` version: 0.9.0
- `@inertiajs/inertia-react` version: 0.6.0

### Describe the problem:

[Using `import()` in resolveComponent](https://inertiajs.com/client-side-setup#code-splitting) won't work, results in following error in browser:

```
React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: object. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

I believe this is due to the `component` here is a `Promise` object:

https://github.com/inertiajs/inertia/blob/88bf77eb1f5406d416351532f6deb7b7ea1448c7/packages/inertia-react/src/App.js#L13

### Steps to reproduce:

 Initialize a minimal React / Laravel project, adding following code to `app.js`:

```js
import { App } from '@inertiajs/inertia-react'
import React from 'react'
import { render } from 'react-dom'

const el = document.getElementById('app')

render(
  <App
    resolveComponent={(name) =>
      import(`./Pages/${name}`).then((res) => res.default)
    }
  />,
  el,
)
```

Adding some pages in `Pages` folder, and run the server.

reinink:
Resolves #696

This PR adds a new `initialComponent` prop to the Vue 2, Vue 3 and React adapters. The primary purpose of this prop is for apps using SSR.

When running the app in Node, you no longer need to provide a `resolveComponent` prop. Instead, you now just provide a `initialComponent` instead.

```diff
- resolveComponent: (name) => require(`./Pages/${name}`).default,
+ initialComponent: require(`./Pages/${request.body.component}`).default,
```

When running an SSR-enable app in the browser, when using code-splitting, you can also pass in an `initialComponent` prop to ensure that client-side hydration works. Otherwise, if you don't do this, the first client-side render will not have the component on time, and will render blank, and hydration will fail.

### DIFF
--- a/packages/inertia-react/src/App.js
+++ b/packages/inertia-react/src/App.js
@@ -6,12 +6,13 @@ import { createElement, useEffect, useMemo, useState } from 'react'
 export default function App({
   children,
   initialPage,
+  initialComponent,
   resolveComponent,
   onHeadUpdate,
 }) {
   const [current, setCurrent] = useState({
-    component: resolveComponent(initialPage.component),
-    page: initialPage,
+    component: initialComponent || null,
+    page: initialComponent ? initialPage : null,
     key: null,
   })
 
--- a/packages/inertia-vue/src/app.js
+++ b/packages/inertia-vue/src/app.js
@@ -14,9 +14,13 @@ export default {
       type: Object,
       required: true,
     },
+    initialComponent: {
+      type: Object,
+      required: false,
+    },
     resolveComponent: {
       type: Function,
-      required: true,
+      required: false,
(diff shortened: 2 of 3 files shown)


lvillacin:
Can this be added to the documentation? Had some trouble with hydration since https://inertiajs.com/server-side-rendering#client-side-hydration nor does https://inertiajs.com/code-splitting mention this

PR #698, opened 2021-05-27
Add createInertiaApp() method

reinink:
This PR adds a new `createInertiaApp()` method to the adapters, making Inertia easier to configure. Here is an example using the Vue 2 adapter:

```js
import Vue from 'vue'
import { createInertiaApp } from '@inertiajs/inertia-vue'

createInertiaApp({
  resolve: name => import(`./Pages/${name}`),
  setup({ el, app, props }) {
    new Vue({
      render: h => h(app, props),
    }).$mount(el)
  },
})
```

TODO:

- [x] Vue 2
- [x] Vue 3
- [x] React
- [x] Svelte

### DIFF
--- a/packages/inertia-react/src/createInertiaApp.js
+++ b/packages/inertia-react/src/createInertiaApp.js
@@ -1,0 +1,35 @@
+import App from './App'
+import { createElement } from 'react'
+
+export default async function createInertiaApp({ id = 'app', resolve, setup, page, render }) {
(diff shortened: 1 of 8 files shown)


d8vjork:
Is there any update left on the types file?

Would be good to have autocompletion on the IDE and TS support on this

andrewl64:
How do we add global components to the `createInertiaApp()` method?

PR #714, opened 2021-06-02
fix(vue2): initialComponent can be a function

jaulz:


claudiodekker:
Theoretically, it can also a be a string, and that should also work (as you can call `"".foo` and that'll just give you `undefined`, similar to what happens with this PR when doing `(function() {}).foo`). 

https://vuejs.org/v2/guide/render-function.html#createElement-Arguments:
 ```
createElement(
  // {String | Object | Function}
  // An HTML tag name, component options, or async
  // function resolving to one of these. Required.
  'div',

...
```

Because [we explicitly do this check here](https://github.com/inertiajs/inertia/blob/master/packages/inertia-vue/src/app.js#L62-L75), I don't foresee any problems with what this specific PR tries to add (note: I didn't actually test it), BUT in that case I'd support the string-variant as well..

Thoughts @reinink?

**Comment by jaulz** [comment]
@claudiodekker thanks for the feedback. In my case it just caused some verbose warnings because my components are no objects. To be absolutely correct you are right and the prop validation should also include strings. However, f you really think that the explicit check later is already enough, then the prop validation should be removed completely, shouldn't it?

**Comment by claudiodekker** [comment]
Perhaps, but that would then imply that passing in a Boolean is also fine. I think adding strings explicitly makes more sense.

**Comment by jaulz** [comment]
It's in 😊

**Review by claudiodekker — APPROVED** [review]

**Comment by reinink** [comment]
> because my components are no objects

@jaulz I am curious, can you show us a little more code from your setup? I've never run into this before.

**Comment by jaulz** [comment]
@reinink sorry, I meant to say "no functions" 😉  I use the class based approach with decorators and thus `Function` is not enough for me.

**Comment by reinink** [comment]
@jaulz Are you able to share a simple example of one of these page components?

**Comment by jaulz** [comment]
@reinink sure, here you go:
```
<template>
  <div></div>
</template>

<script lang="ts">
import {
  Vue,
  Component,
  Prop,
  Watch,
  Provide,
  ProvideReactive,
} from 'vue-property-decorator'

@Component
export default class Tabs extends Vue {

}
</script>

```

reinink:
Cool, thanks! 👍

PR #717, opened 2021-06-03
fix: fix types

jaulz:
While working with the newest version I noticed several incorrect typings. If you agree, these types would also need to be adapted in `vue3` and `react` package.

Also, currently, types can only be imported from the types file directly but it would be great to have them all exported on package level.


claudiodekker:
Thanks @jaulz!

PR #718, opened 2021-06-03
fix: use proper types for events

jaulz:


jaulz:
@claudiodekker as discussed in Discord here is my proposal how the event system could be properly typed 😊

jaulz:
@claudiodekker any chance to merge this? 😊

PR #719, opened 2021-06-04
Add support for fragments in Head component in Vue 3 adapter

### LINKED ISSUE #715: Problems with InertiaHead component on Vue 3
### Versions:

- `@inertiajs/inertia` version: 0.9.1
- `@inertiajs/inertia-vue3` version: 0.4.2

### Describe the problem:

Problem using InertiaHead global component on my Vue 3 app:

`<inertia-head title="Test - MyApp" />`

image

### Steps to reproduce:

Just use `<inertia-head>` component on another

reinink:
Fixes #715

Right now if you include a `<slot />` within `<inertia-head>`, it throws an error, since the head component doesn't know how to render a fragment. This fixes that by checking to see if a fragment was provided, and if one was, uses its children instead.

### DIFF
--- a/packages/inertia-vue3/src/head.js
+++ b/packages/inertia-vue3/src/head.js
@@ -41,7 +41,9 @@ export default {
         : node.children.reduce((html, child) => html + this.renderTag(child), '')
     },
     renderTag(node) {
-      if (node.type.toString() === 'Symbol(Text)') {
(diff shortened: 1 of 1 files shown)


reinink:
Hmm, I wonder in hindsight if this work should have been done in the `renderTag()` method instead. 🤔

PR #725, opened 2021-06-05
allow array values to be passed in form

iZaL:
with the new typescript types addition, Its not possible to send the array values to the form data, for example the below code outputs the type error "Type 'CollectionCondition[]' is not assignable to type 'string'."

```ts
export interface CollectionCondition {
  id: number;
  field: CollectionField;
  criteria: string;
  value: string;
}

export interface Collection {
  id: number;
  name: string;
  determiner: CollectionDeterminer;
  conditions: Array<CollectionCondition>;
}

const { data, setData } = useForm<Collection>({
  id: 1,
  name: "Summer collection",
  type: "manual",
  determiner: "all",
  conditions: [],
});

Inertia.post(url, data, {
  onSuccess: () => {
    Inertia.reload();
  },
});
```

with the changes adding Array<any> type in FormDataConvertible type we can pass the array values to the form.

### DIFF
--- a/packages/inertia/src/types.ts
+++ b/packages/inertia/src/types.ts
@@ -3,7 +3,7 @@ import { CancelTokenSource } from 'axios'
 export type Errors = Record<string, string>
 export type ErrorBag = Record<string, Errors>
 
-export type FormDataConvertible = Date|File|Blob|boolean|string|number|null|undefined
+export type FormDataConvertible = Date|File|Blob|Array<any>|boolean|string|number|null|undefined
 
 export enum Method {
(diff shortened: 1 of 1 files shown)


iZaL:
any update regarding this PR ?

claudiodekker:
Hey @iZaL, thanks for this! 

I'll merge this for now as to recognize your contribution, but will also look into solving this more 'correctly', as we don't really want to allow `any` here (as for instance, a `symbol` or object by itself is still part of `any`, but is not a valid Form value)

PR #728, opened 2021-06-08
Fix import of App.svelte in createInertiaApp.js

joburgard:
I am using Rollup and with the current version of the Svelte adapter it aborts with this error:
```
[!] Error: Could not resolve './App' from node_modules\@inertiajs\inertia-svelte\src\createInertiaApp.js
```
This change fixes it.


reinink:
Good catch, thanks! 🙏

reinink:
Released: http://inertiajs.com/releases/inertia-svelte-0.7.2-2021-06-08

PR #737, opened 2021-06-09
Add global typescript Inertia SharedData definition

ycs77:
Add the old global typescript Inertia PageProps definition that was previously removed.

PagePropsBeforeTransform is not added, because `transformProps()` is removed in #693.

Source:
https://github.com/inertiajs/inertia/blob/76a2328c751ce53e0268e211a8e8644094a7a8fc/packages/inertia/index.d.ts#L3-L14

### DIFF
--- a/packages/inertia/src/types.ts
+++ b/packages/inertia/src/types.ts
@@ -4,7 +4,7 @@ declare global {
   // These open interfaces may be extended in an application-specific manner via
   // declaration merging / interface augmentation.
   namespace Inertia {
-    interface PageProps {}
+    interface SharedData {}
   }
 }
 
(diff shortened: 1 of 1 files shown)


ycs77:
I later found out that the interface `PageProps` corresponds to Laravel adapter's `SharedData` part, so to make it easier to understand, I renamed the interface to `SharedData`.

rovansteen:
Would love to see this merged, since there's no way to type the `PageProps` now.

claudiodekker:
Thank you for contributing this PR! With that said, I'll be closing this PR in favor of #757, as that one appears slightly more flexible if the application scope (and with it, the shared data) changes.

ycs77:
PageProps types can now be defined directly:

*resources/js/env.d.ts*
```ts
declare module '@inertiajs/inertia' {
  interface PageProps {
    user: {
      id: number
      name: string
      email: string
    }
    myGlobalProp: string
  }
}

export {}
```

PR #738, opened 2021-06-10
Type options param for visit shortcuts.

Spice-King:
This preserves types, while eliminating options rendered unusable due to the method overriding them, all while not adding new types to keep updated in 7 other places.


claudiodekker:
Thanks @Spice-King!

PR #740, opened 2021-06-10
Disable layout attribute inheritance in Vue 3 adapter

reinink:
Resolves #733

This change is similar to #681, where unused page props (that are passed to the layout component as a result of #602) are being actually rendered as attributes in the Vue 3 adapter. For example, the `errors` prop:

```html
<div errors="[object Object]">
  <!-- ... -->
</div>
```

This PR fixes this by automatically disabling attribute inheritance on the layout components by setting `inheritAttrs` to `false`.

### DIFF
--- a/packages/inertia-vue3/src/app.js
+++ b/packages/inertia-vue3/src/app.js
@@ -65,14 +65,18 @@ export default {
         if (component.value.layout) {
           if (typeof component.value.layout === 'function') {
             return component.value.layout(h, child)
-          } else if (Array.isArray(component.value.layout)) {
-            return component.value.layout
-              .concat(child)
-              .reverse()
-              .reduce((child, layout) => h(layout, { ...page.value.props }, () => child))
           }
 
-          return h(component.value.layout, { ...page.value.props }, () => child)
+          return (Array.isArray(component.value.layout) ? component.value.layout : [component.value.layout])
+            .concat(child)
+            .reverse()
+            .map((layout) => {
+              if (layout.inheritAttrs === undefined) {
+                layout.inheritAttrs = false
+              }
+              return layout
+            })
+            .reduce((child, layout) => h(layout, { ...page.value.props }, () => child))
         }
 
         return child
--- a/packages/inertia-vue3/src/app.js
+++ b/packages/inertia-vue3/src/app.js
@@ -70,12 +70,7 @@ export default {
           return (Array.isArray(component.value.layout) ? component.value.layout : [component.value.layout])
             .concat(child)
             .reverse()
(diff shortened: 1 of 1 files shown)


packages/inertia-vue3/src/app.js
```diff
@@ -65,14 +65,18 @@ export default {
...
           }
 
-          return h(component.value.layout, { ...page.value.props }, () => child)
+          return (Array.isArray(component.value.layout) ? component.value.layout : [component.value.layout])
+            .concat(child)
+            .reverse()
+            .map((layout) => {
+              if (layout.inheritAttrs === undefined) {
+                layout.inheritAttrs = false
+              }
```
claudiodekker, line 76:
```suggestion
              layout.inheritAttrs = !!layout.inheritAttrs
```

packages/inertia-vue3/src/app.js
```diff
@@ -65,14 +65,18 @@ export default {
...
+              return layout
+            })
```
claudiodekker, line 78:
```suggestion
           .map(layout => ({...layout, inheritAttrs: !!layout.inheritAttrs}))
```

packages/inertia-vue3/src/app.js
claudiodekker, line 76:
```suggestion
              layout.inheritAttrs = !!layout.inheritAttrs
```

claudiodekker (approved):

shengslogar:
Reading through discussion https://github.com/inertiajs/inertia/discussions/733 makes it clear why this is desired behavior. However, this bit me in one specific way today, and I wanted to document for posterity.

https://github.com/inertiajs/inertia/blob/4a4f256a3fffe1ddf5dfb2ca5973fc5eba2e126d/packages/inertia-vue3/src/app.js#L70-L75

As intended, this PR will default all layout components without `defineOptions({ inheritAttrs: ... })` specified to `false`, whereas the Vue default is `true`. But if you have multiple layouts and want to take advantage of [async components](https://vuejs.org/guide/components/async) to keep your bundle size small, you will find that Vue's `defineAsyncComponent(() => import('Layout.vue'))` method will behave differently by nature of being a function:

https://github.com/inertiajs/inertia/blob/4a4f256a3fffe1ddf5dfb2ca5973fc5eba2e126d/packages/inertia-vue3/src/app.js#L66-L68

The obvious "fix" is adding `defineOptions({ inheritAttrs: false })` to the underlying layout component you are async-ing, but I found it counterintuitive that regular ol' components and `defineAsyncComponent`s weren't swappable within Inertia like they otherwise are throughout Vue.

PR #741, opened 2021-06-10
fix(vue2): expose all types

jaulz:
I would like to use some of the types in my code but they are not exposed right now.


claudiodekker:
Thanks for this! I've already merged a different PR, and will be making some adjustments soon to make the whole manual management thing unnecessary 👍

PR #752, opened 2021-06-14
Allow defining only a subset of VisitOptions

### LINKED ISSUE #751: [TS] VisitOptions definition forces us to use all options
### Versions:

- `@inertiajs/inertia` version: 0.9.2
- `@inertiajs/inertia-vue3` version: 0.4.7

### Describe the problem:
Using the form helper and trying to pass a visit option such as `onSuccess` triggers TS errors, it expects all of VisitOptions to be passed or none at all.

eg:
```js
form.post('some-url', {
  onSuccess: () => {/*... */}
})

// Argument of type '{ onSuccess: (page: Page) => void; }' is not assignable to parameter of type 'VisitOptions'.
// Type '{ onSuccess: (page: Page) => void; }' is missing the following properties from type 'Visit': method, data, replace, preserveScroll, and 5 more.
```

obennaci:
Current Typescript definition of `VisitOptions` doesn't allow setting only a subset of the options.

This MR makes it possible by wrapping the definition with the TS built-in [Partial Type](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype)

Closes #751


paulohpsrock:
Any plans to integrate this PR?

PR #753, opened 2021-06-15
Add `<title>` callback support

reinink:
This PR adds the ability to globally modify the page `<title>` when set via the `<inertia-head>` component. A great use case for this is automatically adding an app name before or after each page title.

To use this feature, first set the `title` callback in your `createInertiaApp` configuration:

```js
createInertiaApp({
  title: title => `${title} - Ping CRM`,
  // ...
})
```

And then set a title using the `<inertia-head>` component in your pages:

```html
<inertia-head title="Dashboard">
```

This will result in:

```html
<title>Dashboard - Ping CRM</title>
```

Note, this also works if you set the title using a `<title>` tag:

```html
<inertia-head>
  <title>Dashboard</title>
</inertia-head>
```

## TODO

- [x] Vue 2
- [x] Vue 3
- [x] React

### DIFF
--- a/packages/inertia-react/src/App.js
+++ b/packages/inertia-react/src/App.js
@@ -8,6 +8,7 @@ export default function App({
   initialPage,
   initialComponent,
   resolveComponent,
+  titleCallback,
   onHeadUpdate,
 }) {
   const [current, setCurrent] = useState({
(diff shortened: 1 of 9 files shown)


emargareten:
@reinink is it possible to add `title` as a page option like `layout`?

reinink:
@emargareten Nope, sorry. And honestly, I'd like to avoid adding additional properties to the Vue component object. It feels like we're opening ourselves up to future collisions. And the component approach works beautifully. 👍

PR #757, opened 2021-06-18
Add generic argument to Page type to allow for shared props typing

janpantel:
This allows adding an optional type to the `Page` interface to create type safety for shared props.

Example:
```typescript
// before
const { foo } = usePage().props; // no type safety :-(

// now
type MyPage = Page<{foo: string}>;
const { foo } = usePage<MyPage>().props; // yeah type safe 'foo'
```

### DIFF
--- a/packages/inertia/src/types.ts
+++ b/packages/inertia/src/types.ts
@@ -19,9 +19,9 @@ export interface PageProps {
   [key: string]: unknown
 }
 
-export interface Page {
+export interface Page<SharedPropsType = Record<string, unknown>> {
   component: string,
-  props: PageProps & {
+  props: PageProps & SharedPropsType & {
     errors: Errors & ErrorBag;
   }
   url: string,


janpantel:
This goes into the same direction as https://github.com/inertiajs/inertia/pull/737 but allows to add the type param to the definition of the `SharedProps` interface on the app level.

claudiodekker:
Thanks @janpantel!

rovansteen:
@reinink is it correct that this hasn't been released yet?

reinink:
Yes, it's been released now as `@inertiajs/inertia@0.9.3`. 👍

reinink:
Correction, use `@inertiajs/inertia@0.10.0`. I made a mistake releasing it as `@inertiajs/inertia@0.9.4`, as it's a breaking release. You'll also need to be on the latest versions of Vue adapters, if you're using them. They are `@inertiajs/inertia-vue@0.7.0` and `@inertiajs/inertia-vue3@0.5.1`. Sorry for the confusion.

PR #759, opened 2021-06-18
Update modal.ts

ivanmdh:
Force div position to keep iframe centered


reinink:
I've never run into a situation where the error modal wasn't centered. Can you explain the situation that caused this?

PR #774, opened 2021-06-27
Add ability to ignore Inertia default `handlePopstateEvent`

crynobone:
This would allow a SPA application to take control of back button action if necessary. Otherwise, we always bound to Inertia default implementation. 

In my use case, I need to intercept the back button request and display a confirmation message. Not too elegant but this would enable us to build actions similar to `beforeRouteLeave()` with vue-router.

### DIFF
--- a/packages/inertia/src/router.ts
+++ b/packages/inertia/src/router.ts
@@ -13,6 +13,7 @@ export class Router {
   protected activeVisit?: ActiveVisit
   protected visitId: VisitId = null
   protected page!: Page
+  protected ignoreHistoryState: false
 
   public init({
     initialPage,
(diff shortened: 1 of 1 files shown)


reinink:
One note, if you really want to do this on a page, you can do this:

```js
window.addEventListener('popstate', (event) => {
    event.stopImmediatePropagation()
})
```

Calling `event.stopImmediatePropagation()` like this will stop Inertia from picking up this event. However, this event listener must be define BEFORE Inertia is booted, otherwise it will run too late. Meaning you'll need to create some type of global state  in your app that you can check to see if Inertia's `popstate` handler should be ignored.

Keep in mind, however, that your users will still go back in the browser history. That's unavoidable.

Issue #775, opened 2021-06-30
DOMException: Failed to execute 'replaceState' on 'History': #<Object> could not be cloned


iamohd-zz:
### Versions:

- `@inertiajs/inertia` version: 0.9.2
- `@inertiajs/inertia-vue3` version: 0.4.7

### Describe the problem:

I am using [Vue 3 Draggable](https://github.com/SortableJS/vue.draggable.next) package. In the event of dragging an element, I execute Inertia.put method and pass some data to Laravel. This is where this exception occurs.

### Steps to reproduce:

For example, in Laravel pass a project with a list of columns to your view. Then use the Draggable component as following:

```vue
<Draggable
    v-model="project.columns"
    group="columns"
    item-key="id"
    @end="onColumnPositionChanged"
>
    <template #item="{element: column}">
        <KanbanColumn
             :column="column"
             :key="column.id"
         />
    </template>
</Draggable>
```

In event of onColumnPositionChanged, trigger Inertia.put method

```js
const onCardPositionChanged = () => {
    Inertia.put('/some-route');
};
```

reinink:
I suspect this has something to do with the URL you're submitting to, which I assume isn't `/some-route`. What's the *exact* URL that's being used when submitting the request?

iamohd:
> I suspect this has something to do with the URL you're submitting to, which I assume isn't `/some-route`. What's the *exact* URL that's being used when submitting the request?

Thanks for the quick reply

The URL is:
 http://127.0.0.1:8000/back/projects/dummy-slug/columns/sort

iamohd:
I think the issue is because of trying to push a complex object into the page state. Once I replaced the page state with simple data, it worked without any issue. 

This StackOverflow's question is related to this issue
https://stackoverflow.com/questions/50618225/uncaught-domexception-failed-to-execute-pushstate-on-history-function

reinink:
> I think the issue is because of trying to push a complex object into the page state. Once I replaced the page state with simple data, it worked without any issue.

So, how are you getting complex data into your page props? We actually intentionally removed the `transformProps` option because of this very issue. The data that is saved to history state should just be the props (JSON) that come back from the server.

reinink:
Also, can you please provide an example of what specifically you're doing to cause this? I don't see how the draggable package/component should have any bearing on what ends up in history state.

iamohd:
At first, I thought it's because of Vue Draggable, but it has nothing to do with this problem.

[...]

PR #776, opened 2021-07-01
Fix DOMException when replacing or pushing page history state with a complex object

iamohd:
When trying to replace a page history state with a complex object, DOMException will be thrown. #775 

This PR fixes this issue.

### DIFF
--- a/packages/inertia/src/router.ts
+++ b/packages/inertia/src/router.ts
@@ -378,12 +378,12 @@ export class Router {
 
   protected pushState(page: Page): void {
     this.page = page
-    window.history.pushState(page, '', page.url)
+    window.history.pushState(JSON.stringify(page), '', page.url)
   }
 
   protected replaceState(page: Page): void {
     this.page = page
-    window.history.replaceState(page, '', page.url)
+    window.history.replaceState(JSON.stringify(page), '', page.url)
   }
 
   protected handlePopstateEvent(event: PopStateEvent): void {


reinink:
Hmm, this shouldn't ever be necessary, as the `page` object that's received from a server response is what's saved in history state, and that comes from a JSON response. Meaning, there shouldn't ever be a complex JavaScript object within the `page` object. I'm honestly sure how that would be possible, without doing something weird/hacky.

rkyoku:
@reinink could you please remove function before serializing?

Maybe this is what you call "weird" or "hacky", but I am dynamically adding functions to my object/prop.

Sure it's a different philosophy from using events for everything, but the advantages of this are:

- more intuitive (a mere function call),
- less weird and unnecessary code to communicate from child to parent,
- fewer bugs (e.g. I forgot to attach the handler to the child comp, or misspelled the event name),
- last but not least: it also works BEFORE the child component gets mounted (which doesn't work with events).

Events are golden when props are not passed to components. But when they are, they become irrelevant.

Compare this:

```js
// ParentComp.svelte
export let myobj // retrieved from page props
myobj.onMyEvent = e => {/*do stuff*/}
<ChildComp {myobj} />

// ChildComp.svelte
export let myobj // passed from parent
myobj.onMyEvent()
```

With this:

```js
// ParentComp.svelte
export let myobj
const onMyEvent = e => {/*do stuff*/}
<ChildComp {myobj} on:myevent={onMyEvent} />

// ChildComp.svelte
import { createEventDispatcher, onMount } from 'svelte'
export let myobj
const dispatch = createEventDispatcher()
onMount(() => {dispatch('myevent', {})})
```

Sure, you don't save 300 lines of code, but it's still a shorter syntax, more intuitive too, and with better separation of concerns and also less imports, and less error-prone (e.g. "oh fudge, I've been investigating this bug for at least 15 minutes but I just noticed I forgot to attach my event handler" kind of thing).

All you have to do in Inertia is to remove the functions from objects that get serialized, it's not a big deal, and it will give a lot more flexibility to devs.

PR #801, opened 2021-07-12
feat(vue-3): `components` property for `createInertiaApp`

innocenzi:
# Summary

This PR adds a new non-breaking feature that allows passing a record of components to the `createInertiaApp` factory via the `components` property: 

```js
createInertiaApp({
	components: import.meta.glob('../views/pages/**/*.vue'), // <--- this is new
	setup({ el, app, props, plugin }) {
			.use(plugin)
			.mount(el)
	},
})
```

# Behavior

The behavior is the following: 
- If `components` is defined, `resolve` is replaced by a function that looks up the given record and returns the corresponding page component
- If the page component is not found, a clean error is printed in the console. 
- If `components` was not given, `resolve` behaves just like before.

# Interface

The record must be of this form: 

```ts
interface ComponentRecord {
  [path: string] ComponentResolvable
}
```

For instance, Vite's [`import.meta.glob`](https://vitejs.dev/guide/features.html#glob-import) returns something like this: 

```js
// import.meta.glob('../views/pages/**/*.vue')
{
  '../views/pages/index.vue': function() {},
  '../views/pages/about.vue': function() {},
  '../views/pages/dashboard/index.vue': function() {}
}
```

# Use cases

This may not be super helpful with Webpack, since ``resolve: name => require(`./Pages/${name}`)`` is available and clear enough, but it's very handy with Vite and reduces the boilerplate a lot. 

The feature is designed to not be only specific to Vite though, since `components` is just a map of paths/components. 

A nice bonus is the console message when the component is not found.

# Alternative

Alternatively, we could export the resolve logic from the package and use it like the following: 

```ts
import { createInertiaApp, withComponents } from '@inertiajs/inertia-vue3'

createInertiaApp({
	resolve: withComponents(import.meta.glob('../views/pages/**/*.vue')),
	setup({ el, app, props, plugin }) {
			.use(plugin)
			.mount(el)
	},
})
```

The only benefit I see of this approach would be the introduced tree-shakability, though.

### DIFF
--- a/packages/inertia-vue3/index.d.ts
+++ b/packages/inertia-vue3/index.d.ts
@@ -9,6 +9,10 @@ export interface InertiaAppProps {
 }
 
 type InertiaApp = DefineComponent<InertiaAppProps>
+type ComponentResolvable = DefineComponent 
+  | Promise<DefineComponent> 
+  | { default: DefineComponent }
+  | (() => Promise<{ [key: string]: any; }>)
 
 export declare const App: InertiaApp
 
@@ -16,10 +20,8 @@ export declare const plugin: Plugin
 
 export interface CreateInertiaAppProps {
   id?: string
-  resolve: (name: string) => 
-    DefineComponent |
-    Promise<DefineComponent> |
-    { default: DefineComponent }
+  components: Record<string, ComponentResolvable>
+  resolve?: (name: string) => ComponentResolvable
   setup: (props: {
     el: Element
     app: InertiaApp
--- a/packages/inertia-vue3/src/createInertiaApp.js
+++ b/packages/inertia-vue3/src/createInertiaApp.js
@@ -1,10 +1,25 @@
 import { createSSRApp, h } from 'vue'
 import App, { plugin } from './app'
 
-export default async function createInertiaApp({ id = 'app', resolve, setup, title, page, render }) {
+export default async function createInertiaApp({ id = 'app', resolve, components, setup, title, page, render }) {
   const isServer = typeof window === 'undefined'
   const el = isServer ? null : document.getElementById(id)
   const initialPage = page || JSON.parse(el.dataset.page)
+
+  if (components) {
+    resolve = (name) => {
+      for (const path in components) {
+        if (path.endsWith(`${name.replace('.', '/')}.vue`)) {
+          return typeof components[path] === 'function'
+            ? components[path]()
+            : components[path]
+        }
+      }
+
+      throw new Error('Page component not found: ' + name)
+    }
+  }
+
   const resolveComponent = name => Promise.resolve(resolve(name)).then(module => module.default || module)
 
   let head = []


reinink:
So, `import.meta.glob` returns an object, correct? Would it maybe make sense to update the `resolve` property to accept an object, so that we don't have to introduce another property here? That way you could do this:

```js
createInertiaApp({
  resolve: import.meta.glob('../views/pages/**/*.vue'),
  setup({ el, app, props, plugin }) {
      .use(plugin)
      .mount(el)
  },
})
```

> The feature is designed to not be only specific to Vite though, since `components` is just a map of paths/components.

Is there any other libraries or tools that use this same format? Or is it, practically speaking, just a Vite format?

innocenzi:
> Would it maybe make sense to update the `resolve` property to accept an object, so that we don't have to introduce another property here? That way you could do this

Yes, I think it would make sense - although I initially thought of a new property because of the clarity of the name. 

> Is there any other libraries or tools that use this same format? Or is it, practically speaking, just a Vite format?

Short answer is: it's a Vite format. 

I don't know any tool that does the same, the closest I'm aware of would be `require.context` but it's not exactly the same one. Admittedly, I can't see any other use for it right now, but the point was that it's just a record and not some very specific interface.

Want me to adapt the PR to only use `resolve` instead of adding `components`?

reinink:
> Short answer is: it's a Vite format.

I'm okay with that. I want Inertia to work awesome with Vite. My primary concern is with setting default layouts. Right now the way you do this is via the `resolve` callback. Something like this:

```js
createInertiaApp({
  resolve: name => {
    const page = require(`./Pages/${name}`).default
    page.layout = page.layout || Layout
    return page
  },
  // ...
})
```

Any thoughts on how this would be done if `components` can optionally be an object?

Interestingly, this maybe also connects to the issues with setting the layout via the new `<script setup>` syntax, which you explored here (#651).

innocenzi:
I actually thought about that prior to the PR, but I was thinking about discussing a simple component implementation first before deciding on how to implement layouts. 

I thoughts of two simple ways:

**Adding a new `layout` property**

```ts
import { createInertiaApp, withComponents } from '@inertiajs/inertia-vue3'
import layout from '@/views/layouts/default.vue'

createInertiaApp({
	layout,
	components: import.meta.glob('../views/pages/**/*.vue'),
	setup({ el, app, props, plugin }) {
			.use(plugin)
			.mount(el)
	},
})
```

**Using the alternative solution's `withComponent` with an additional parameter**

```ts
import { createInertiaApp, withComponents } from '@inertiajs/inertia-vue3'
import layout from '@/views/layouts/default.vue'

createInertiaApp({
	resolve: withComponents(import.meta.glob('../views/pages/**/*.vue'), layout),
	setup({ el, app, props, plugin }) {
			.use(plugin)
			.mount(el)
	},
})
``` 

In either case, the implemention would be similar than previously: 

```ts
// in the resolve function
const component = typeof components[path] === 'function'
	? components[path]()
	: components[path]

if (layout) {
	component.layout &&= layout
}

return component
```

winter-ice:
This is my way for use `Vite` and default layout:
```
import { createInertiaApp,Head,Link } from '@inertiajs/inertia-vue3'
import Layout from './Layouts/App/Fixed/Index.vue'

function withVite (pages, name) {
  for (const path in pages) {
    if (path.endsWith(`${name.replace('.', '/')}.vue`)) {
      return typeof pages[path] === 'function'
        ? pages[path]()
        : pages[path]
    }
  }

  throw new Error('Page not found: ' + name)
}

createInertiaApp({
  resolve: (name) => withVite(import.meta.glob('./Pages/**/*.vue'), name)
    .then((page) => {
      page.default.layout = page.default.layout || Layout
      return Promise.resolve(page)
    }),
  setup ({ el, app, props, plugin }) {
      .use(plugin)
      .component('Head',Head)
      .component('Link',Link)
      .mount(el)
  },
})

```

It would be great if there was a better official way to integrate for `Vite` users.

PR #809, opened 2021-07-14
[inertia-vue3] Deprecate `plugin` in favor of `inertia`

innocenzi:
## Summary

This PR deprecates the `plugin` export of `createInertiaApp` in favor of `inertia`: 

```ts
import { createInertiaApp } from '@inertiajs/inertia-vue3'

createInertiaApp({
  setup({ el, app, props, inertia }) {
      .use(inertia) // <-- instead of `plugin`
      .mount(el)
  },
})
```

## Reasoning

Quite often, we use different Vue plugins, and the plugin variables are named upon the plugin name. For instance, using `vue-i18n` would look like that: 

```ts
createApp().use(i18n)
```

Currently, Inertia's plugin is named `plugin`, which doesn't really make sense in this context. Imagine you have multiple plugins: 

```ts
createInertiaApp({
  setup({ el, app, props, inertia }) {
      .use(plugin)
      .use(i18n)
      .use(toast)
      .use(calendar)
      .mount(el)
  },
})
```

For clarity, it should be named `inertia`. 

While the variable can be renamed in the destructuration (`{ plugin: inertia }`), I think it should just be the default, for consistency's sake and so the boilerplate can be reduced.

### DIFF
--- a/packages/inertia-vue3/index.d.ts
+++ b/packages/inertia-vue3/index.d.ts
@@ -24,7 +24,9 @@ export interface CreateInertiaAppProps {
     el: Element
     app: InertiaApp
     props: InertiaAppProps
+		/** @deprecated Use `inertia` instead */
     plugin: Plugin
+    inertia: Plugin
   }) => void | VueApp
   page?: Inertia.Page
   render?: (app: VueApp) => Promise<string>
--- a/packages/inertia-vue3/src/createInertiaApp.js
+++ b/packages/inertia-vue3/src/createInertiaApp.js
@@ -21,7 +21,9 @@ export default async function createInertiaApp({ id = 'app', resolve, setup, tit
         titleCallback: title,
         onHeadUpdate: isServer ? elements => (head = elements) : null,
       },
+      /** @deprecated Use `inertia` instead **/
       plugin,
+      inertia: plugin,
     })
   })
 


reinink:
So yeah, I think I struggled with this very decision when coming up with this API. It's tricky, because strictly speaking the term `plugin` makes a lot of sense coming from the `setup()` method. It's clearly indicates what it is. However, I agree, the term `inertia` would be more appropriate when you consider how you typically register Vue plugins.

The reason we decided to go with `plugin` in the end was exactly for the reason you mentioned—the fact that you can alias it easily enough with `{ plugin: inertia }`. I think for that reason, I'd like to leave it as it.

Thanks either way for this suggestion @innocenzi, and for helping to improve Inertia! 💜

innocenzi:
@reinink Wouldn't it be better to make it make sense from the user point of view instead of the internals point of view? It can be aliased in the export (like this PR does) and we don't even have to deprecate `plugin` if you prefer. 

It really bugs me actually 😬 

I don't think there's a downside to adding an `inertia` export, is there?

reinink:
Well, it's not really internal only. If it was, I'd agree. This is a property that's available alongside a bunch of other properties (`el`, `app`, `props`) within your `app.js` file. Why does it make sense to call the plugin `inertia` and not, for example, the `app` property? I think an argument could totally be made for renaming `app` to `inertia` as well. But obviously they can't both be called `inertia`. 😂 

Plus, you basically never interact with the `createInertiaApp` code beyond the initial work setting it up.

Because of these reasons, and the above reasons, I just don't think this needs to be changed.

Thanks again for your input. 👍

innocenzi:
Right, that makes sense. 

> I think an argument could totally be made for renaming `app` to `inertia` as well

Totally haha, and I've thought about it for a while before giving up on a better name. The difference with `plugin` is that it is pretty visible that the name is confusing when mixed up with other plugins. But I understand your point of view too.

Thanks for putting up with me 👍

reinink:
All good buddy. And again, definitely just do `{ plugin: inertia }` if you're finding it confusing in your plugin registration. 👍

PR #816, opened 2021-07-16
Update inertia-vue definitions

TullariS:
Added missing InertiaHead definition. Based on inertia-vue3 definitions.


dacoto97:
Duplicated with #819

reinink:
Thanks for this, but I'm closing this in favour of #819. ❤️

PR #817, opened 2021-07-17
Add `isDirty` definition for `useForm` hook

digitaltoad:
Just adding `isDirty` type definition for the `useForm` hook in the React adapter.


reinink:
Thanks @digitaltoad!

PR #821, opened 2021-07-23
fire navigate event after resolving initial component

rovansteen:
This PR changes the behaviour of the initial `navigate` event to be fired *after* the initial component has been resolved. This is similar to the behaviour on subsequent visits in Inertia.

The use case I have for this is that I want to set a value to be remembered after a `navigate` event. However due to the event being fired before the component has been resolved the history state will be overwritten.

```js
Inertia.on('navigate', () => {
	Inertia.remember(Date.now(), 'key');
});
```

Now if you'd check the history after the component has been loaded, you see the `rememberedState` is empty.

```js
window.history.state.rememberedState
> {}
```

### DIFF
--- a/packages/inertia/src/router.ts
+++ b/packages/inertia/src/router.ts
@@ -26,21 +26,21 @@ export class Router {
     this.page = initialPage
     this.resolveComponent = resolveComponent
     this.swapComponent = swapComponent
-    this.handleInitialPageVisit()
-    this.setupEventListeners()
-  }
 
-  protected handleInitialPageVisit(): void {
     if (this.isBackForwardVisit()) {
       this.handleBackForwardVisit(this.page)
     } else if (this.isLocationVisit()) {
       this.handleLocationVisit(this.page)
     } else {
-      this.page.url += window.location.hash
-      this.setPage(this.page, { preserveState: true }).then(() => {
-        fireNavigateEvent(this.page)
-      })
+      this.handleInitialPageVisit(this.page)
     }
+
+    this.setupEventListeners()
+  }
+
+  protected handleInitialPageVisit(page: Page): void {
+    this.page.url += window.location.hash
+    this.setPage(page, { preserveState: true }).then(() => fireNavigateEvent(page))
   }
 
   protected setupEventListeners(): void {
--- a/packages/inertia/src/router.ts
+++ b/packages/inertia/src/router.ts
@@ -37,9 +37,10 @@ export class Router {
       this.handleLocationVisit(this.page)
     } else {
       this.page.url += window.location.hash
-      this.setPage(this.page, { preserveState: true })
+      this.setPage(this.page, { preserveState: true }).then(() => {
+        fireNavigateEvent(this.page)
+      })
     }
-    fireNavigateEvent(this.page)
   }
 
   protected setupEventListeners(): void {
@@ -102,6 +103,7 @@ export class Router {
     window.history.state.version = page.version
     this.setPage(window.history.state, { preserveScroll: true, preserveState: true }).then(() => {
       this.restoreScrollPositions()
+      fireNavigateEvent(page)
     })
   }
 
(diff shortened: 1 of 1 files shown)


RobertBoes:
Firing the event on load seems like unintended behaviour. Inertia doesn't navigate in this case, so the event shouldn't be fired by Inertia imo.
I'm using this event for analytics to track Inertia page views, which would result in double entries, because it logs when the analytics load and then fires another event.

If you want to store something when the page loads, wouldn't it be possible to just call this inside your setup? (example from Vue 3)
```js
import { createInertiaApp } from '@inertiajs/inertia-vue3'

createInertiaApp({
  resolve: name => require(`./Pages/${name}`),
  setup({ el, app, props, plugin }) {
    Inertia.remember(Date.now(), 'key');

      .use(plugin)
      .mount(el)
  },
})
```

rovansteen:
@RobertBoes the event is already fired on load. This PR just changes it so it's fired _after_ the component is loaded, so that it's safe to write something to the history state. Your example suffers from the same problem, calling `Inertia.remember` will write to the history state, but once the component is loaded it will be overridden.

RobertBoes:
> @RobertBoes the event is already fired on load by Inertia. This PR just changes it so it's fired _after_ the component is loaded, so that it's safe to write something to the history state. Your example suffers from the same problem, calling `Inertia.remember` will write to the history state, but once the component is loaded it will be overridden.
> 
> If you use the `navigate` event for analytics you'll already have double entries if you also fire an event on load e.g. in the setup method.

Interesting, after some debugging I see this behaviour has been changed/added with the `createInertiaApp()` setup method. I'm using `@inertiajs/inertia-vue3@^0.4.7`, but haven't converted to the new setup method, because it's only "to make configuring Inertia easier". After changing to `createInertiaApp()`, like described [here](https://inertiajs.com/releases/inertia-vue3-0.4.1-2021-05-31) the `navigate` event is also triggered on page load, which seems to me like a breaking change that has been introduced somewhere in `0.4.1` to `0.4.7`

rovansteen:
It seems to be added in this commit, almost a year ago already: https://github.com/inertiajs/inertia/commit/73a960b614918556f45c61a2751b268ac7874886#diff-467a987931305e1cb6da2436149dad1d780e881e4df994731a88c53bbdf53b08R26

reinink:
@RobertBoes We've fired the `navigate` event on the initial page load ever since we introduced the event system. It's fired on the initial page load, on Inertia visits, and on history state navigation. This is the desired behaviour, since this event is intended to allow for page visit tracking, which you want in all these situations.

That said, it definitely should not be firing twice on the initial load, and if it is, we've done something wrong somewhere along the lines. Let me look into this. 😅 

And I like this change @rovansteen. I agree, the event should be fired at the same time...after the page is resolved/rendered. Let me pull this down and do some testing, and we can hopefully get this merged in soon. 👌 

Thanks to both of you for digging into this! 👍

reinink:
Okay, I just tested this, and it looks good to me. The `navigate` event fires properly, at the desired time. I'm also not seeing any duplicate events being fired.

I did make some minor formatting changes.

I'm going to wait for @claudiodekker's 👀 on this one, and once it has his approval, we'll merge it in.

RobertBoes:
> @RobertBoes We've fired the `navigate` event on the initial page load ever since we introduced the event system. It's fired on the initial page load, on Inertia visits, and on history state navigation. This is the desired behaviour, since this event is intended to allow for page visit tracking, which you want in all these situations.

@reinink whoops, my bad 😅 
I think I have unintentionally "abused" the events behaviour. I always had the event registration áfter Vue's `createApp`, so the app would be mounted before the handler would be registered. Now that `createInertiaApp` is async the handler is registered before the app is mounted. So to keep my hacky workaround I'm now using this
```js
createInertiaApp({
  setup({ el, app, props, plugin }) {
      .use(plugin)
      .mount(el)

    Inertia.on('navigate', () => { ... })
  },
})
```

claudiodekker (approved):

reinink:
Thanks everyone!

PR #825, opened 2021-07-24
Add typings for createInertiaApp

### LINKED ISSUE #823: Module '"@inertiajs/inertia-react"' has no exported member 'createInertiaApp'
### Versions:

- `@inertiajs/inertia` version: 0.10.0
- `@inertiajs/inertia-react` version: 0.7.0

### Describe the problem:

As the title implies, there is no exported createInertiaApp member from inertia-react. Also, react types.d.ts don't have it as well, even though, vue3 types.d.ts do have it. 

### Steps to reproduce:
Try to follow the client-side installation instructions from inertia.js for react.

anubra266:
This fixes #823

### DIFF
--- a/packages/inertia-react/index.d.ts
+++ b/packages/inertia-react/index.d.ts
@@ -1,20 +1,21 @@
 import * as Inertia from '@inertiajs/inertia'
 import * as React from 'react'
+import { renderToString } from "react-dom/server";
 
-type App<
-  PagePropsBeforeTransform extends Inertia.PagePropsBeforeTransform = Inertia.PagePropsBeforeTransform,
-  PageProps extends Inertia.PageProps = Inertia.PageProps
-> = React.FunctionComponent<{
+export type ReactInstance = React.ReactElement;
+export type ReactComponent = React.ReactNode;
+
+export type PageResolver = (name: string) => ReactComponent|Promise<ReactComponent>|NodeRequire // TODO: When shipped, replace with: Inertia.PageResolver<ReactComponent>
+export type HeadManagerOnUpdate = (elements: string[]) => void // TODO: When shipped, replace with: Inertia.HeadManagerOnUpdate
+export type HeadManagerTitleCallback = (title: string) => string  // TODO: When shipped, replace with: Inertia.HeadManagerTitleCallback
+
+export type AppType<SharedProps = Inertia.PageProps> = React.FunctionComponent<{
   children?: (props: {
     Component: React.ComponentType
     key: React.Key
-    props: PageProps
+    props: (Inertia.Page<SharedProps>)['props']
   }) => React.ReactNode
-  initialPage: Inertia.Page<PageProps>
-  resolveComponent: (
-    name: string
-  ) => React.ComponentType | Promise<React.ComponentType>
-}>
+} & SetupOptions<unknown, SharedProps>['props']>
 
 interface BaseInertiaLinkProps {
   as?: string
@@ -57,9 +58,9 @@ export const InertiaLink: InertiaLink
 
 export const Link: InertiaLink
 
-export const InertiaApp: App
+export const InertiaApp: AppType
 
-export const App: App
+export const App: AppType
 
 type setDataByObject<TForm> = (data: TForm) => void
 type setDataByMethod<TForm> = (data: (previousData: TForm) => TForm) => void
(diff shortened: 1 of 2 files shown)


packages/inertia-react/index.d.ts
```diff
@@ -86,3 +86,29 @@ export interface InertiaFormProps<TForm = Record<string, any>> {
...
+    initialPage: Inertia.PageProps,
+    initialComponent: any,
+    resolveComponent: (name: string) => Promise<any>,
+    titleCallback: (title: string) => string,
+    onHeadUpdate: ((elements: any[]) => boolean) | null,
+}
+
+export interface AppSetupProps {
+    el: HTMLElement | null
+    App: (props: AppSetupPageProps) => JSX.Element
```
RobertBoes, line 100:
```suggestion
    App: App
```

packages/inertia-react/index.d.ts
```diff
@@ -86,3 +86,29 @@ export interface InertiaFormProps<TForm = Record<string, any>> {
...
+export interface AppSetupPageProps {
+    initialPage: Inertia.PageProps,
+    initialComponent: any,
+    resolveComponent: (name: string) => Promise<any>,
+    titleCallback: (title: string) => string,
+    onHeadUpdate: ((elements: any[]) => boolean) | null,
+}
+
+export interface AppSetupProps {
+    el: HTMLElement | null
```
RobertBoes, line 99:
```suggestion
    el: HTMLElement
```
Think this doesn't have to be null

packages/inertia-react/index.d.ts
```diff
@@ -86,3 +86,29 @@ export interface InertiaFormProps<TForm = Record<string, any>> {
 }
 
 export function useForm<TForm = Record<string, any>>(initialValues: TForm): InertiaFormProps<TForm>;
+
+export interface AppSetupPageProps {
+    initialPage: Inertia.PageProps,
+    initialComponent: any,
+    resolveComponent: (name: string) => Promise<any>,
```
RobertBoes, line 93:
```suggestion
    resolveComponent: (name: string) => React.ComponentType | Promise<React.ComponentType>,
```
See https://github.com/inertiajs/inertia/blob/master/packages/inertia-react/index.d.ts#L16

packages/inertia-react/index.d.ts
```diff
@@ -86,3 +86,29 @@ export interface InertiaFormProps<TForm = Record<string, any>> {
...
+    titleCallback: (title: string) => string,
+    onHeadUpdate: ((elements: any[]) => boolean) | null,
```
RobertBoes, line 95:
```suggestion
    onHeadUpdate: (elements: string[]) => void,
```
See the `onUpdate` parameter at https://github.com/inertiajs/inertia/blob/master/packages/inertia/src/head.ts#L56

RobertBoes (changes requested):
There's a few `any` types remaining, but I don't know what should be returned without diving into React

anubra266:
Thanks for the review @RobertBoes. I've made the changes.

reinink:
@sebastiandedeyne @claudiodekker Gents, can I get your 👀 on this? 🙏

packages/inertia-react/index.d.ts
```diff
@@ -86,3 +86,29 @@ export interface InertiaFormProps<TForm = Record<string, any>> {
 }
 
 export function useForm<TForm = Record<string, any>>(initialValues: TForm): InertiaFormProps<TForm>;
+
+export interface AppSetupPageProps {
+    initialPage: Inertia.Page<Inertia.PageProps>,
```
claudiodekker, line 91: `Inertia.PageProps` is unnecessary here, as now it's being double defined and merged (directly overriding the defaults). This option is really only ever necessary in userland to include shared props.

```suggestion
    initialPage: Inertia.Page,
```

anubra266:
Thanks @claudiodekker

claudiodekker:
@anubra266 there's still some other issues with this PR (not your fault, it's SSR-related), but I'll push those fixes to your PR in an hour or so, so we can merge this after that. 👍

anubra266:
ok...cool

packages/inertia-react/index.d.ts
```diff
@@ -1,20 +1,21 @@
 import * as Inertia from '@inertiajs/inertia'
 import * as React from 'react'
+import { renderToString } from "react-dom/server";
 
-type App<
-  PagePropsBeforeTransform extends Inertia.PagePropsBeforeTransform = Inertia.PagePropsBeforeTransform,
-  PageProps extends Inertia.PageProps = Inertia.PageProps
-> = React.FunctionComponent<{
+export type ReactInstance = React.ReactElement;
+export type ReactComponent = React.ReactNode;
```
sebastiandedeyne, line 6: Why are these aliased and re-exported?

claudiodekker: @sebastiandedeyne This way, it's easier to work with these / swap and test out the changes in case we're referencing the wrong (sub)type somewhere in the code. It's basically a little shortcut that's relevant as long as the whole thing isn't converted to 100% TS (yet), as apart from the team there isn't really anyone actively helping out with this.

sebastiandedeyne (approved):

reinink:
Thanks all! 👍

PR #831, opened 2021-07-28
Check if scroll region's position is available before restoring

rovansteen:
While using Inertia the following error sometimes pops up:
```
Uncaught (in promise) TypeError: Cannot read property 'top' of undefined
    at router.ts:89
    at NodeList.forEach (<anonymous>)
    at e2.n2.restoreScrollPositions (router.ts:88)
    at router.ts:401
```
Relating to this line: https://github.com/inertiajs/inertia/blob/master/packages/inertia/src/router.ts#L88

I am not entirely sure why the scroll regions are not properly saved sometimes so perhaps this does't fix the underlying issue, but it wouldn't hurt to have a little sanity check here that checks if the value is present before applying it.

### DIFF
--- a/packages/inertia/src/router.ts
+++ b/packages/inertia/src/router.ts
@@ -86,7 +86,6 @@ export class Router {
     if (this.page.scrollRegions) {
       this.scrollRegions().forEach((region: Element, index: number) => {
         const scrollPosition = this.page.scrollRegions[index]
-
         if (scrollPosition) {
           region.scrollTop = scrollPosition.top
           region.scrollLeft = scrollPosition.left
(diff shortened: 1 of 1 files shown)


reinink:
Good call @rovansteen, thanks for submitting this PR. 👍

claudiodekker:
Yeah, if I remember correctly, this occurs because not everything on the page is rendered/mounted yet at the time of restore (the app/layout itself, child components, and child components of child components). I had a really ugly fix/hack for this in the past, that if I believe only partly worked: 

telegram-cloud-photo-size-4-5864249115490235630-y

I believe the way to 'fix' this correctly (for future reference) is to somehow determine that all components are mounted first (app 'ready'), and only then run the scroll restoration. Or, always run restoration whenever a 'restored' component is re-mounted or something.

In either case, thanks for this!

reinink:
Thanks for that context @claudiodekker. That's dark. Makes me wonder if simply adding a slight timeout (<10ms) would be enough to solve this in most cases? 😬

reinink:
@claudiodekker Oh lol, I didn't even read your code close enough to see that's exactly what you're suggesting! 😂

PR #836, opened 2021-07-30
Fix wrong type in InertiaFormProps.submit method

### LINKED ISSUE #837: Wrong type used for submit method in InertiaFormProps
### Versions:

- `@inertiajs/inertia` version: 0.10.0
- `@inertiajs/inertia-react` version: 0.7.0

### Describe the problem:

It seems like there is a wrong type used for `submit` method in `InertiaFormProps`. Looking to the source of useForm hook, it should accept one of `get`, `post`, `patch`, `put`, `delete`, but it is actually typed to expect `() => void`. 

It makes this method pretty unusable when working with TypeScript.

rafalkrawiec:
Proposes a fix for a wrong type used for `method` argument in `InertiaFormProps.submit` method, which actually should be one of `Inertia.Method` enum values, or maybe a union type like:

```ts
type submitMethod = 'get' | 'patch' | 'post' | 'put' | 'delete'
```

fixes #837

### DIFF
--- a/packages/inertia-react/index.d.ts
+++ b/packages/inertia-react/index.d.ts
@@ -79,7 +79,7 @@ export interface InertiaFormProps<TForm = Record<string, any>> {
 	transform: (callback: (data: TForm) => TForm) => void
 	reset: (...fields: (keyof TForm)[]) => void
 	clearErrors: (...fields: (keyof TForm)[]) => void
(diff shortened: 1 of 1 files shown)


reinink:
Seems reasonable, from my limited TS knowledge. Thanks! 👍

PR #845, opened 2021-08-03
feat: allow null value for replace parameter

jaulz:
This PR will allow users to pass `null` to `replace` parameter. A null value means that neither a new state will be pushed nor a new state will be set. This can be handy for infinite scroll pages where the URL should not change but anyway new items should be loaded.

### DIFF
--- a/packages/inertia/src/router.ts
+++ b/packages/inertia/src/router.ts
@@ -373,6 +373,8 @@ export class Router {
           this.replaceState(page)
         } else if (replace === false) {
           this.pushState(page)
+        } else {
+          this.keepState(page)
         }
         this.swapComponent({ component, page, preserveState }).then(() => {
           if (!preserveScroll) {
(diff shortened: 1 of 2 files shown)


claudiodekker:
Hey @jaulz,

I'm a bit confused. Isn't this why the `preserveState` component exists? History-wise, you can either replace the current state, or push a new entry, but to 'keep' it sounds like an intermediate state without any state at all, which (to me) sounds problematic as it more or less 'disconnects' from reality?

jaulz:
@claudiodekker yep, you are right. It's an intermediate state that actually only exists at the time of scrolling down. Currently, there are two possibilities to solve the infinite scrolling:
1. Navigate to "next" link (with preserveState) via Inertia but that will update the URL. When the user refreshes the page he will land somewhere in the middle which is fine for search results but not for collections sorted by a time based value (e.g. hotness factor)
2. Use an AJAX request to retrieve the results of the "next" link which will not update the URL. Since this would technically work, you would lose all the Inertia magic (e.g. update other props, auth based redirect).

Thus, it would be great if we could navigate with Inertia without updating the URL to achieve this "intermediate" state.

reinink:

PR #770, opened 2021-06-24
Convert @inertiajs/inertia-vue to TypeScript 

claudiodekker:

PR #846, opened 2021-08-03
Update $inertia typescript type in vue 3.

Esirei:
Add missing form helper's type to $inertia object in vue 3.


claudiodekker:
Hi,

First of all, thanks for this PR! Unfortunately I think the best way forward with this specific change is to close it, as we're already working on introducing this fix through https://github.com/inertiajs/inertia/pull/770. Merging this fix will just complicate things, as it'll cause more merge conflicts with the already work-in-progress branch.

PR #855, opened 2021-08-09
Add typings for InertiaHead component

zlokomatic:


packages/inertia-react/index.d.ts
```diff
@@ -58,6 +64,8 @@ export const InertiaLink: InertiaLink
 
 export const Link: InertiaLink
 
+export const InertiaHead: InertiaHead
```
danielbrodin, line 67: I think you are missing a `export const Head: InertiaHead` since the InertiaHead component is also exported as Head

zlokomatic: You are totally right. I added it now

TitasGailius:
Until this is merged you may add your own type definitions for the `Head` component.

Create `types/global.d.ts` and declare your types:
```ts
import React from "react"

/**
 * InertiaJS is missing type definitions for the Head component.
 * Here we define our own types for the component.
 *
 * @see https://github.com/inertiajs/inertia/issues/824
 * @see https://github.com/inertiajs/inertia/pull/855
 */
declare module "@inertiajs/inertia-react" {
    interface HeadProps {
        title?: string
    }

    let Head: React.FC<HeadProps>
}
```

claudiodekker:
Thanks!

PR #876, opened 2021-08-21
Append empty array to formdata

RobertBoes:
## Summary

Currently empty arrays aren't appended to the FormData object, this means sending an empty array won't reach the server at all. It can be confusing when you have a form with an optional file, like so:
```js
const form = useForm({
  image: null,
  tags: [],
});
```
When the image is null, the tags property is received by Laravel as an empty array `[]`. However, when a file is present Inertia converts the object to a FormData object. Now the `tags` property won't be sent to the server at all, not even as a null value.

## FormData limitations

With FormData it seems to be impossible to send an empty array, at least with Laravel. For example doing `form.append('mykey', [])` results in the following:
```php
// $request->all();
[
  "mykey" => "[]"
]
```
Doing `form.append('mykey[]', '')` results in:
```php
// $request->all();
[
  "mykey" => [
    0 => null,
  ]
]
```

Sending a "normal" XHR request with a JSON payload would contain the following and results in an empty array: `{ "mykey": [] }`

This implementation send the key as an empty string, same as null and undefined values, so `$request->all()` will result in:
```php
[
  "mykey": null,
]
```

### DIFF
--- a/packages/inertia/src/formData.ts
+++ b/packages/inertia/src/formData.ts
@@ -35,7 +35,7 @@ function append(form: FormData, key: string, value: FormDataConvertible): void {
     return form.append(key, value)
   } else if (typeof value === 'number') {
     return form.append(key, `${value}`)
-  } else if (value === null || value === undefined) {
+  } else if (value === null || value === undefined || (Array.isArray(value) && !value.length)) {
     return form.append(key, '')
   }
 
--- a/packages/inertia/src/formData.ts
+++ b/packages/inertia/src/formData.ts
@@ -22,7 +22,10 @@ function composeKey(parent: string|null, key: string): string {
 
 function append(form: FormData, key: string, value: FormDataConvertible): void {
   if (Array.isArray(value)) {
-    return Array.from(value.keys()).forEach(index => append(form, composeKey(key, index.toString()), value[index]))
+    if (value.length) {
+      return Array.from(value.keys()).forEach(index => append(form, composeKey(key, index.toString()), value[index]))
+    }
+    return form.append(key, '')
   } else if (value instanceof Date) {
     return form.append(key, value.toISOString())
   } else if (value instanceof File) {
@@ -35,7 +38,7 @@ function append(form: FormData, key: string, value: FormDataConvertible): void {
     return form.append(key, value)
   } else if (typeof value === 'number') {
     return form.append(key, `${value}`)
-  } else if (value === null || value === undefined || (Array.isArray(value) && !value.length)) {
+  } else if (value === null || value === undefined) {
     return form.append(key, '')
   }
 


claudiodekker:
Hey @RobertBoes,

I'm not sure what the correct behaviour here would be, actually. I don't want to do something just because Laravel does or does not do it, but I also do agree that receiving it conditionally based on some arbitrary condition such as there being a file is a bit strange.

As such, I'm going to leave this one open for now with the given labels attached, so we can think this over at a later point.

Thanks!

HassanZahirnia:
Thanks Robert for the PR ! I was watching this thread since I've reported this issue and was hoping the fix would go through.
Anyway I'm currently using a hack to get around this issue. Which is right before validation I check for the array and set it if it doesn't exist :
```php
$this->request->input('tags') ? false : $this->request->merge(['tags' => []]);
```
It's a bit unpractical but it gets the job done, I guess.

LasseRafn:
I think ideally empty arrays would still be sent (as an empty array, not as a null/empty string), since they're not NOTHING, just empty. 

If I have a PUT endpoint which updates only form fields that are sent in the request, and leave the remaining fields untouched (partial modifications) then without this, it would be impossible to set an array to empty, since the backend would never receive the empty array and therefor know to update it to `[]`. Checking for existence of the field and setting to empty if not set would mean its always required to send the full array even if you're attempting to update a completely different field.

Consider this pseudo-api for users and teams:

```
PUT /api/users/1

{
    "name": "Test"
}
```

Should only update the user name, but if we check for existence of for example `teams` and resetting if not set, then when updating the user name, we reset their teams unintentionally.


Instead this would be the correct payload to remove the user from all `teams`.
```
PUT /api/users/1

{
    "teams": []
}
```

RobertBoes:
> I think ideally empty arrays would still be sent (as an empty array, not as a null/empty string), since they're not NOTHING, just empty. 
> 
> If I have a PUT endpoint which updates only form fields that are sent in the request, and leave the remaining fields untouched (partial modifications) then without this, it would be impossible to set an array to empty, since the backend would never receive the empty array and therefor know to update it to `[]`. Checking for existence of the field and setting to empty if not set would mean its always required to send the full array even if you're attempting to update a completely different field.
> 
> Consider this pseudo-api for users and teams:
> 
> ```
> PUT /api/users/1
> 
> {
>     "name": "Test"
> }
> ```
> 
> Should only update the user name, but if we check for existence of for example `teams` and resetting if not set, then when updating the user name, we reset their teams unintentionally.
> 
> 
> Instead this would be the correct payload to remove the user from all `teams`.
> ```
> PUT /api/users/1
> 
> {
>     "teams": []
> }
> ```

This PR was about Inertia's automatic conversion to FormData when the form contains a file. Empty arrays are sent by Inertia, unless it's being converted to FormData. There is also a limitation there, because FormData doesn't have something as an empty array, you simply can't send an empty array, so that value won't be sent at all. My proposed solution was to send it as null value so the server would still receive the key, instead of not receiving it at all. But again, this only applies when the form contains a file and Inertia converts the request to FormData

LasseRafn:
@RobertBoes ahh got it, so basically to replicate the default (no-file in form) behavior as closely as possible

reinink:

PR #877, opened 2021-08-21
Fix useForm progress type

guihigashi:
`useForm`'s `progress` property is typed as a number in react package. I noticed it is actually a 
[`ProgressEvent`](https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent) extended with the `percentage` in:
https://github.com/inertiajs/inertia/blob/1dc7fef4cdbd7c506b7213235a98ddb21d4f4db9/packages/inertia/src/router.ts#L287-L289

Based on this, I propose a new `Inertia.Progress` type so it can be used by the other packages.


claudiodekker:
Thanks @guihigashi!

Issue #590, opened 2021-03-11
Form helper not preserving state and errors


piljac1:
Hi guys! First of all, this is my first project using the Vue 3/Inertia combo and I gotta say I love it so far! However, I encountered what might not be a bug (could be, but the problem is probably on my side).

So I have this data setup

```js
data() {
  return {
    form: this.$inertia.form({
      redirect_to_page: 'ParentCategorySelection',
      product_option_value_id: '',
      product_color_id: null,
      area: 0,
      option_ids: []
    })
  }
},
```

So far, everything works alright, all v-models are working and data was validated using Vue devtools, as well as by outputting them directly in the page.

Now comes the time to post the form. Here's what I have:

```js
addProduct(page) {
  this.form.redirect_to_page = page;

  // P.S. this.postRoute comes from a prop
  this.form.post(this.postRoute);
}
```

When I try to post the form while having invalid fields (according to my form request validation), the page is reloaded because validation errors were encountered (tried removing all validations to see if the reload could be due to something else, but when I do, the behavior is as intended and the user gets redirected on another page as desired).

At this point, this is the intended behavior. The page should be reloaded upon validation errors, like it successfully does right now.

But... the problem is that in the doc (and when I inspect the form object), it mentions that we should be able to access errors from the form using `form.error.the_field`. However, the error object is always empty, despite the fact that the global `$form.props.errors` contain the appropriate errors.

Also, reading the doc, the state should be preserved automatically when using the post method. It is not the case for me. All form attributes fall back to their default values. I tried explicitly setting the `preserveState` flag to true, but no dice.

Let me know if you need more info. Thanks in advance for your input!

### EDIT : 

My present workaround is to retrieve all inputs in my controller create method and send them as an `oldForm` prop. Then, in my Vue component, I do the following:

```js
form: this.$inertia.form(
  _.isEmpty(this.oldForm)
    ? {
      redirect_to_page: 'ParentCategorySelection',
      product_option_value_id: '',
      product_flodeco_color_id: null,
      area: 0,
      option_ids: []
    }
    : this.oldForm
),
```

This isn't ideal because I still don't know why the form helper doesn't send back errors and doesn't preserve the state, but it fixes my issue for now because I need to meet the deadline.

[...]

PR #884, opened 2021-08-27
Watch for end of processing and reset form state with server response

pipoprods:
(partially) Fixes inertiajs/inertia#590

Based on @msonowal proposal here: inertia/issues/590#issuecomment-878075166

To be reactive, the `useForm` has to be passed a reactive structure, eg:

```
const form = useForm(computed(() => props.user));
```


Propaganistas:
This PR is opinionated. It assumes the server response is exactly in the form's structure, but that may not always be the case.

I've taken a different take on this in the PR mentioned above.

PR #902, opened 2021-09-15
Change return type of transform helper on useForm to unknown

rovansteen:
Currently the `transform` method should always return `TData` however that's problematic because that basically would prevent the user from transforming the data to any other structure than the original structure of the form.

I choose to replace it with `unknown` but it might be better to use a dedicated type for which data can be submitted with Inertia. Is `Record<string, FormDataConvertible>` from `inertia/formData.ts` the correct type for this? In that case it might also be good to replace all the other types like `InertiaFormProps<TForm = Record<string, any>>` to `Record<string, FormDataConvertible>` as well. If you'd like to have that I can do that in a follow up PR.

### DIFF
--- a/packages/inertia-react/index.d.ts
+++ b/packages/inertia-react/index.d.ts
@@ -76,7 +76,7 @@ export interface InertiaFormProps<TForm = Record<string, any>> {
 	wasSuccessful: boolean
 	recentlySuccessful: boolean
 	setData: setDataByObject<TForm> & setDataByMethod<TForm> & setDataByKeyValuePair<TForm>
-	transform: (callback: (data: TForm) => TForm) => void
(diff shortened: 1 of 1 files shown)


claudiodekker:
Hey Roy,

Yeah, I think you're right. This is actually one of the reasons that we're re-doing the entire Inertia codebase (including all adapters) in TypeScript. Not only should it help with proper type definitions, but it should also make issues like these more clear.

Anyway, I think I'm just going to close this PR for now, as I do think that while this might be more correct, it would still (I think?) not entirely be correct with what it should be eventually, and as such I'd rather choose the 'known wrong but everyone is already using it that way' version over the 'still wrong but unknown'.

Hope this makes sense, and thanks again for contributing this PR!

PR #926, opened 2021-10-07
feat(link): support `disabled` property

innocenzi:
This PR adds support for the `disabled` property on Inertia links. The current behavior is to ignore it and still call `Inertia.visit`. This PR changes that behavior to avoid having to manually handle that scenario.


RomanSarvarov:
I don't think it's a good idea.

For me, when I work with Link component I want to think I'm working with `<a>` tag.
As I see `<a>` tag does not respect `disabled` attribute.
https://jsfiddle.net/ujc1973b/

Probably you need just use @click.prevent.

claudiodekker:
Hey @innocenzi 👋,

While I do like this, I do think we'll want to follow the HTML5 standards and only allow this when the `as` is set to a `<button>`, `<command>`, `<fieldset>`, `<keygen>`, `<optgroup>`, `<option>`, `<select>`, `<textarea>` or <input>. (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled)

Given that pretty much all of those also don't really make sense/apply to an Inertia Link we'll probably want to only respect this property when it's a button, so if you could make this change then that would be great!

Furthermore, it would be great if you could port this to React & Svelte. No worries if you can't or don't have time, we'll just pick it up and merge this PR later then, as we want to maintain feature parity 👍

innocenzi:
I think it's not really a necessary restriction to allow the property to work only on `button`s, but let's follow the standard. I made the changes for the Vue adapters. 

I also tried to port this to Svelte and React, but I have zero knowledge in these two frameworks and I did not test my changes. I think it should be working and if not, it should be pretty close, but it should still be tested by someone who has knowledge about them.

claudiodekker:
Yeah, I understand what you're saying, but a-links normally shouldn't be expected to respect the disabled property, and given that we're using it here in an almost HTML-like fashion, I really want to make sure we're sticking to the spec as much as possible.

In either case, thanks for updating this! I'll check this out later 👍

PR #944, opened 2021-10-18
Expose "setErrors" method in useForm

stephenfrank:
Gives more flexibility and options to customise the errors displayed when using other API calls or methods to update the form (ie. not alway 100% inertia requests)

### DIFF
--- a/packages/inertia-react/src/useForm.js
+++ b/packages/inertia-react/src/useForm.js
@@ -137,6 +137,7 @@ export default function useForm(...args) {
     isDirty: !isEqual(data, defaults),
     errors,
     hasErrors,
+    setErrors,
     processing,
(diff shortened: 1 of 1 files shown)


claudiodekker:
Hey @stephenfrank,

Apologies for that last comment, I was a bit too close with closing this, as I forgot we actually had this method internally.

All in all, this seems like it might actually be a good change that aligns with something we're planning to do in the future, so I'm probably going to merge this once we make some additional changes (other adapters, fixing the `setHasErrors` state correctly etc.)

Thanks!

claudiodekker:
Merged this into #999, where this PR is re-implemented and aligned with the other adapters. Thanks again! 👍

Issue #866, opened 2021-08-17
rememberedState of undefined - React


yashiroiori:
### Versions:

- `@inertiajs/inertia` version: ^0.10.1
- `@inertiajs/inertia-react` version: ^0.7.1

### Describe the problem:
Try use useRemember hook but return error "rememberedState of undefined"

### Steps to reproduce:
Into component try use an example displayed in section "Remembering state"

![image](https://user-images.githubusercontent.com/17031855/129791936-0f98ce71-a440-40a2-b66d-cb0b56f46329.png)

Error
![image](https://user-images.githubusercontent.com/17031855/129792004-a367db90-4a9c-4769-9f99-2febe606eead.png)

BjornDCode:
We experienced the same issue so I did some digging. It only happens on the first page load. If I navigate to the page using `useRemember` from another page it loads fine. 

The issue appeared when we switched from the old rendering method (directly rendering App) to using `createInertiaApp`.

It appears the `useRemember` hook runs before Inertia is initialised. `useRemember` calls `Inertia.remember` internally.  The `remember` method relies on `this.page` being set on the Inertia instance. This happens when the `init` method is called, which happens in `createInertiaMethod`. When I log from `useRemember` and `init` it seems like `useRemember` is called before the `init` method.

<img width="658" alt="Screenshot 2021-08-31 at 13 16 50" src="https://user-images.githubusercontent.com/14328774/131493396-72655d50-0897-4895-82e9-61824b15cfbe.png">

yashiroiori:
> We experienced the same issue so I did some digging. It only happens on the first page load. If I navigate to the page using `useRemember` from another page it loads fine.
> 
> The issue appeared when we switched from the old rendering method (directly rendering App) to using `createInertiaApp`.
> 
> It appears the `useRemember` hook runs before Inertia is initialised. `useRemember` calls `Inertia.remember` internally. The `remember` method relies on `this.page` being set on the Inertia instance. This happens when the `init` method is called, which happens in `createInertiaMethod`. When I log from `useRemember` and `init` it seems like `useRemember` is called before the `init` method.
> 
> <img alt="Screenshot 2021-08-31 at 13 16 50" width="658" src="https://user-images.githubusercontent.com/14328774/131493396-72655d50-0897-4895-82e9-61824b15cfbe.png">

I try using Inertia.remeber and it's same error

BjornDCode:
[...]

PR #949, opened 2021-10-21
fix useForm with useRemember

mrtijn:
Useform did not work in combination with useRemember. The types were wrong & the function used an odd way to trigger the useRemember function. This adds an extra param to the useForm to add an rememberkey (just like useRemember works).

You can now add an extra param to trigger the useRemember function.
If you do not add the param, it will use useState instead of useRemember.
```
const initialValues = {
      first_name: '',
      last_name: ''
}

const { data, setData } = useForm(initialValues, 'registerForm');
```

RememberedState also threw an error mentioned in https://github.com/inertiajs/inertia/issues/866 , made this optional to fix it.

### DIFF
--- a/packages/inertia-react/src/useForm.js
+++ b/packages/inertia-react/src/useForm.js
@@ -3,11 +3,11 @@ import useRemember from './useRemember'
 import { Inertia } from '@inertiajs/inertia'
 import { useCallback, useEffect, useRef, useState } from 'react'
 
-export default function useForm(initialData, rememberKey) {
+export default function useForm(rememberKeyOrInitialData, initialData) {
   
   const isMounted = useRef(null)
-  const useRememberKey = typeof rememberKey === 'string' ? rememberKey : null
-  const defaults = initialData || {}
+  const useRememberKey = typeof rememberKey === 'string' ? rememberKeyOrInitialData : null
+  const defaults = useRememberKey === null ? rememberKeyOrInitialData:  initialData
   const cancelToken = useRef(null)
   const recentlySuccessfulTimeoutId = useRef(null)
   const [data, setData] = useRememberKey ? useRemember(defaults, `${useRememberKey}:data`) : useState(defaults)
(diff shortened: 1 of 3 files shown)


packages/inertia-react/src/useForm.js
```diff
@@ -3,14 +3,15 @@ import useRemember from './useRemember'
 import { Inertia } from '@inertiajs/inertia'
 import { useCallback, useEffect, useRef, useState } from 'react'
 
-export default function useForm(...args) {
+export default function useForm(initialData, rememberKey) {
```
claudiodekker, line 6: Can we flip the arguments of this back to how it was? Essentially, `rememberKey` first, and the `initialData` second?

The main reason we've decided to do the `rememberKey` first, has to do with how it looks/reads when you define your form:

```js
form: this.$inertia.form({
  provider: '',
  name: '',
  size: '2048',
  region: null,
  seller: 'online',
  recipe: '',
  is_public: false,
}, 'remember-key'),

```

vs:

```js
form: this.$inertia.form('remember-key', {
  provider: '',
  name: '',
  size: '2048',
  region: null,
  seller: 'online',
  recipe: '',
  is_public: false,
}),
```

On larger forms, the 2nd is a lot more identifiable/readable by the first line.

Apart from this 'preference', we also want to keep all adapters on-par in terms of the features they offer, and how the feature works, so that anyone that works in multiple languages can rely on their muscle memory as much as possible when working with Inertia.

Thanks!

claudiodekker (changes requested):
Hey @mrtijn 👋,

Thank you for this pull request! 

While I'm glad this problem is being solved, there are some concerns that I have with the changes made here. Could you make the necessary changes so we can merge it?

(Furthermore, you're right that a lot of the types are currently a problem, which is in part why we're working on converting the library to be TypeScript from the ground up 🙂)

Thanks again!

packages/inertia-react/src/useForm.js
mrtijn: Hi Claudio!

This does generate a few issues. The first param can be either a rememberKey or the form you want to remember. This can be solved in the code, but makes the typing confusing.

I updated the pull request with your comments processed, but maybe this is something that should be looked at in future releases.

Issue #963, opened 2021-11-04
`<Link>` Component automatically adds `http://localhost` as a prefix to relative path


istiak-tridip:
### Versions:

- `@inertiajs/inertia` version: 0.10.1
- `@inertiajs/inertia-vue3` version: 0.5.2

### Describe the problem:

Passing a relative path to the `<Link>` component which contains the word `http` will cause it to automatically add `http://loclahost` as a prefix.

**Example:**
- Code: `<Link href="/tools/check-http-header">Check</a>`
- Expected Result: `<a href="/tools/check-http-header">Check</a>`
- Current Result: `<a href="http://localhost/tools/check-http-header">Check</a>`

### Steps to reproduce:

Passing any relative path which contains the word `http` to the `<Link>` component will reproduce this issue.

istiak-tridip:
Closing issue as pr #964 was merged

PR #964, opened 2021-11-04
fix: system to check whether a URL has host

istiak-tridip:
Right now `@inertiajs/inertia` detects whether a URL has `host` or not by simply checking if it includes the word `http`. Which causes an error when using a relative URL that includes the word `http` (see issue #963). This PR attempts to solve the issue by checking if the URL starts with `http://` or `https://`.


claudiodekker:
Hey @istiak-tridip 👋,

Thank you for this pull request!

PR #994, opened 2021-12-02
Query String: Allow different 'Array' formatters

claudiodekker:
This PR enables the ability to provide a different method of formatting query string when making GET-requests with URL params (data).

It allows for the following options, available with _all_ Inertia visit types. (The default is the "brackets" type):
```html
<Link href="/foo" :data="{ a: ['b', 'c'] }" queryStringArrayFormat="brackets">Example</Link>
<!-- Generated href: /foo?a[]=b&a[]=c -->

<Link href="/foo" :data="{ a: ['b', 'c'] }" queryStringArrayFormat="indices">Example</Link>
<!-- Generated href: /foo?a[0]=b&a[1]=c -->
```

### DIFF
--- a/packages/inertia-react/src/Link.js
+++ b/packages/inertia-react/src/Link.js
@@ -14,6 +14,7 @@ export default forwardRef(function InertiaLink({
   replace = false,
   only = [],
   headers = {},
+  queryStringArrayFormat = 'brackets',
   onClick = noop,
   onCancelToken = noop,
   onBefore = noop,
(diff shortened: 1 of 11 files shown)


adrianb93:
Nice! Would be great if the default could be configured. I like doing commas.

claudiodekker:
@adrianb93 Stay tuned. 😎

pedroborges:
@claudiodekker I'll be testing the Svelte adapter sometime this week.

claudiodekker:
@pedroborges tested in Svelte, seems to be working fine 👍

Tofandel:
The default makes much more sense as `indices` because serialization is not correct with brackets for array of objects

See #1136

PR #997, opened 2021-12-08
Prevent double submit when processing

khalilst:
This PR prevents double submit when processing the previous submit.

### DIFF
--- a/packages/inertia-react/src/useForm.js
+++ b/packages/inertia-react/src/useForm.js
@@ -27,6 +27,10 @@ export default function useForm(...args) {
 
   const submit = useCallback(
     (method, url, options = {}) => {
+      if (this.processing) {
+        return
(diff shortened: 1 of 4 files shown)


claudiodekker:
Hey 👋,

Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the library, we need to be very careful regarding the amount and kind of code/features we include. Specifically, we believe that there are use-cases in which the user might want to cancel the existing submission and re-submit the form. In part, this is why we have a method of telling whether the form is processing, so that you can manually 'prevent' submission by disabling the submit button or checking for this right before re-submission.

If you feel absolutely certain that this code corrects a bug in the library, please "@" mention me in a follow-up comment with further explanation so that GitHub will send me a notification of your response.

PR #1009, opened 2021-12-14
feat: Add `preserveUrl` option.

xiCO2k:
Hey guys, this PR adds an option to preserve the current url, this is specially useful when there is an infinite scroll pagination.

# Example:

```js
const loadMoreIntersect = ref();

const load = () => {
        return;
    }

    Inertia.get(props.posts.next_page_url, {}, {
        preserveState: true,
        preserveScroll: true,
        preserveUrl: true,
        only: ['posts'],
    })
}

onMounted(() => {
    const observer = new IntersectionObserver(
        { rootMargin: '-150px 0px 0px 0px' }
    );

    observer.observe(loadMoreIntersect.value)
});
```

Thanks,
Francisco.


yizen:
This is a great PR : I just stumbled into that case recently when implementing an infinite scroll, and had to move to an axios call which, even if not that bad, was less elegant than using Inertia.visit for subsequent page loading. Thank you @xiCO2k !

ajnsn:
Hey @xiCO2k 
Interesting! 

Just to add an option to preserve the current url could be too short sighted though. What this issue with the page-query string might also about is to give the developer more control about the `url` coming from the backend. Especially when using the `replace` option which is already correctly preventing pushing the visit into the browsers history.

Some time ago I created a [PR for this in the Laravel's adapter repo](https://github.com/inertiajs/inertia-laravel/pull/206).  This could also be solved and gives you more options and control, even that mentiond in the original PR.

packages/inertia/src/router.ts
```diff
@@ -360,18 +363,24 @@ export class Router {
...
   } = {}): Promise<void> {
     return Promise.resolve(this.resolveComponent(page.component)).then(component => {
       if (visitId === this.visitId) {
         page.scrollRegions = page.scrollRegions || []
         page.rememberedState = page.rememberedState || {}
         replace = replace || hrefToUrl(page.url).href === window.location.href
-        replace ? this.replaceState(page) : this.pushState(page)
+
+        if (! preserveUrl) {
+          replace ? this.replaceState(page) : this.pushState(page)
```
ajnsn, line 381: Hey @xiCO2k 
the current page object is also set inside the `replaceState` and `pushState` methods which could introduce side effects. What you probably want to avoid is the `window.history` inside those methods?

xiCO2k: First I did exactly that, and set the `the.page = page` also with the `preserveUrl`, but actually does not work well.

When you load the `page=2` it does not change the url on top but when you load the `page=3` it changes the url to `page=2`, its related to the `handlePopstateEvent` method.

xiCO2k:
Hey @claudiodekker, any luck checking this PR.

I'm doing all types of workarounds to support infinite scroll paginations. :(

Thanks.

xiCO2k:
Added also a test to this feature, let me know if it looks good to you @claudiodekker.

Thanks.

packages/inertia/src/router.ts
```diff
@@ -374,18 +376,24 @@ export class Router {
...
+        }
```
innocenzi, line 395: This option should not be called `preserveUrl` but something like `saveState`. What it actually does is skip the addition of the current page data to the history, so it's not just skipping replacing the URL. 

If the wanted behavior is to keep the current URL, wouldn't it be better to do this?

```suggestion
        if (preserveUrl) {
        	page.url = window.location.href
        }
        
        replace ? this.replaceState(page) : this.pushState(page)
```

xiCO2k:
@innocenzi that works, just finish testing it, and it works great for infinite scroll stuff.

Hoping @reinink and @claudiodekker can find the time to check this out. 💪💪

ijpatricio:
Hey guys!

This would be so awesome, and this PR looks clean and to the point.

Any status on this??

xiCO2k:
Any luck checking this PR?

PR #1014, opened 2021-12-20
Allows rich text information to be rendered when the page is refreshed

hmingv:
When refreshing the page, the returned content contains rich text information, which will result in JSON.parse() error.

### DIFF
--- a/packages/inertia-react/src/createInertiaApp.js
+++ b/packages/inertia-react/src/createInertiaApp.js
@@ -4,7 +4,7 @@ import { createElement } from 'react'
 export default async function createInertiaApp({ id = 'app', resolve, setup, title, page, render }) {
   const isServer = typeof window === 'undefined'
   const el = isServer ? null : document.getElementById(id)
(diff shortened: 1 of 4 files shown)


claudiodekker:
Hey 👋,

Could you please provide an example of when this problem occurs? It is difficult for us to consider pull requests without this, as it otherwise may introduce breaking changes to the library, 

Thanks!

hmingv:
Please pay attention to the html escape symbol `&quot;`, when I use a rich text editor, the content may contain this value.

When rendering with Inertia::render, as long as this escape symbol is included in the value of the $props variable, the js escape will fail.

E.g.

```
public function index(): Response
{
    return Inertia::render('KnowledgeBase', ['content' => '&quot;']);
        
    OR

    return Inertia::render('KnowledgeBase', ['content&quot;'' => 'data']);
}
```

Access the routing of the above controller method through the browser

> ERROR information
image

hmingv:
The page opened with the Link component will not report this error.

reinink:

PR #1017, opened 2021-12-22
Allow vite dev server to find the entrypoint

buhrmi:
Vite dev server needs the `module` entry in package.json, otherwise it uses dist/index.js, which is not an ES module, resulting in an error when trying to load that file as a module in the browser


buhrmi:
I don't know why, but now vite dev server is happy with dist/index.js and is able to compile it to to an es module upon request.

PR #1052, opened 2022-01-17
Add Global "Visit" Options Hook

claudiodekker:
This PR adds a global hook that's executed prior to any visit. 
This allows for modifying headers etc. in a single place, instead of having to apply them globally.

## Usage

While the URL itself is only available to conditionally apply actions (e.g. on a specific path/prefix) and **cannot be modified**, the options object itself can be adjusted. This options object essentially reflects your Visit options (everything available from an `$inertia.get('/url', options)` etc.).

```js
createInertiaApp({
  // ...
  visitOptions(options, url) {
    options.headers = {
      'some-global-header': 'some-global-header-value'
    }
  
    return options
  }
})
```


sebsobseb:
@claudiodekker Nice. Is this already in the documentation? Couldn't find it.

PR #1054, opened 2022-01-18
Enable preserveState for partial reloaded links

HassanZahirnia:
I noticed Inertia's `Link` component currently does not set `preserveScroll` and `preserveState` to `true` when the `only` prop is provided ( which indicates they're doing a partial reload ). Since `preserveScroll` and `preserveState` are enabled by default for the `Inertia.reload()` method, I believe the same behavior should be applied to the `Link`.

### DIFF
--- a/packages/inertia-vue3/src/link.js
+++ b/packages/inertia-vue3/src/link.js
@@ -25,7 +25,7 @@ export default {
     },
     preserveScroll: {
       type: Boolean,
-      default: false,
+      default: null,
     },
     preserveState: {
       type: Boolean,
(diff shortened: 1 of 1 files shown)


claudiodekker:
Hey @HassanZahirnia 👋,

So, the `only` method doesn't mean you're doing a partial reload per-se, but apart from that, I don't think we'd want some options to change automatically as the result of other options changing, but I'll check with @reinink to get a second-opinion on this.

HassanZahirnia:
Hey Claudio! 😄 Thanks for giving your input on this!
I personally also had concerns about this being sort of a breaking change for everyone using the Link component. But I don't quite understand why using `only` with the Link component is not a partial reload ? The [docs](https://inertiajs.com/links#partial-reloads) kinda indicate they're directly related 🤔

claudiodekker:
Yeah, the docs indeed imply this, but it doesn't necessarily limit itself to this 👍

claudiodekker:
Hey @HassanZahirnia,

So, I've just discussed this with @reinink, and while we both understand why this might be nice, these are very opinionated defaults (you might not always want to preserve scrolling).

More importantly, these changes aren't predictable / easy to remember. It's a lot easier to work with something when it's the same in every place, instead of always having to check the docs to check what something does in situation X compared to situation Y, as that just leads to people getting frustrated and instead manually setting the options to 'force enable' them everywhere even though those options _are already_ the default.

As such, I'll be closing this pull request. Thanks again for contributing to Inertia!

PR #1061, opened 2022-01-23
Add Realtime Validation feature

HassanZahirnia:
This PR which works in junction with this [PR](https://github.com/inertiajs/inertia-laravel/pull/358) is to add real-time ⚡ validation feature to Inertia's form. Note that the validation is done on the backend.
Basically after you initialize the form, you call the `validate()` method and after that, every time you change the form data, it will send a request to the backend to check if that specific input is passing the validation rules.

Here is how the api looks like at a glance:
Controller (2) (Small)

Here is the demo:
Inertia-Realtime-Validation-Demo2

```js
// Validate specific fields
form.validate({
   data: ['first_name', 'last_name'], // Form data that needs to be validated
   url: '/contacts', // Route to check the validation against, needs the backend implementation to work
   method: 'post', // Request method
})

// Validate all fields
form.validate({ url: '/contacts', method: 'post'})

// Full options you can access
form.validate({
   data: ['first_name', 'last_name'],
   url: '/contacts',
   method: 'post',

   enabled: true, // Option to enable/disable the validation, default: true
   processing: true, // Updates form.processing, default: false
   progress: true, // Updates form.progress, default: false
})
// You only call the validate function once
// Currently validate has a default debounce delay of 400ms
```
I've tested this with the PingCRM example and it works just fine ✅👍

- [x] Add the backend implementation PR
- [x] Improve the arguments/options of `validate`
- [x] Improve `validate` method to avoid clearing existing `page.errors`
- [x] Add `delete` method support
- [x] Write logic to check if provided data to `validate` is in the `form.data()`
- [x] Copy `submit` events behavior (selectively) to make `validate` behave similar
- [x] Add option to control whether the form's `processing` and `progress` should be updated during validation
- [x] Add ability to validate all form data at once ( without the need to specify each input name )
- [ ] Add ability to define custom debounce delay
- [ ] Try to add support for `keyup` or `blur` event on inputs as a way to trigger the validation method, instead of watching the `form.data()`

Any feedbacks would be appreciated 💜😄

### DIFF
--- a/packages/inertia-vue3/package.json
+++ b/packages/inertia-vue3/package.json
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "lodash.clonedeep": "^4.5.0",
+    "lodash.debounce": "^4.0.8",
     "lodash.isequal": "^4.5.0"
   }
 }
--- a/packages/inertia-vue3/src/useForm.js
+++ b/packages/inertia-vue3/src/useForm.js
@@ -211,17 +211,26 @@ export default function useForm(...args) {
   }, { immediate: true, deep: true })
 
   const runRealtimeValidation = debounce((newValue, prevValue) => {
+    // Check if realtimeValidation options are not empty.
     let optionsNotEmpty = form.realtimeValidationOptions !== {}
+    // Check if realtime validation is enabled.
     let enabled = typeof form.realtimeValidationOptions.enabled == 'boolean' ? form.realtimeValidationOptions.enabled : optionsNotEmpty
+    // Check if a valid method is provided.
     let method = typeof form.realtimeValidationOptions.method == 'string' && ['get', 'post', 'put', 'patch'].includes(form.realtimeValidationOptions.method) ? form.realtimeValidationOptions.method : undefined
+    // Set the url.
     let url = typeof form.realtimeValidationOptions.url == 'string' ? form.realtimeValidationOptions.url : undefined
+    // Set the data. The data needs to exist in the form.
     let data = form.realtimeValidationOptions.data?.length ? form.realtimeValidationOptions.data.filter((element) => Object.keys(newValue).includes(element)) : undefined
+    // Check if all necessary arguments are provided.
     if (optionsNotEmpty && enabled === true && method && url && data?.length) {
+      // Find the form data that has changed since last watch update.
       let changedData = data.filter((element) => newValue[element] != prevValue[element])
+      // Check if the changed data is not empty.
       if(changedData?.length) {
         Inertia[method](url, {
           _realtimeValidation: true,
           ...data.reduce((carry, key) => {
+            // Only set the key/value pairs that have changed.
             if(newValue[key] != prevValue[key]) carry[key] = newValue[key]
             return carry
           }, {})
@@ -240,7 +249,7 @@ export default function useForm(...args) {
       }
     }
   }, 150)
-
+  // Watcher for realtime validation.
   watch(() => form.data(), (newValue, prevValue) => runRealtimeValidation(newValue, prevValue), { immediate: false, deep: true })
 
   return form
--- a/packages/inertia-vue3/src/useForm.js
+++ b/packages/inertia-vue3/src/useForm.js
@@ -215,8 +215,8 @@ export default function useForm(...args) {
     let enabled = typeof form.realtimeValidationOptions.enabled == 'boolean' ? form.realtimeValidationOptions.enabled : optionsNotEmpty
     let method = typeof form.realtimeValidationOptions.method == 'string' && ['get', 'post', 'put', 'patch'].includes(form.realtimeValidationOptions.method) ? form.realtimeValidationOptions.method : undefined
     let url = typeof form.realtimeValidationOptions.url == 'string' ? form.realtimeValidationOptions.url : undefined
-    let data = form.realtimeValidationOptions.data?.length != 0 ? form.realtimeValidationOptions.data : undefined
-    if (optionsNotEmpty && enabled === true && method && url && data) {
+    let data = form.realtimeValidationOptions.data?.length != 0 ? form.realtimeValidationOptions.data.filter((element) => Object.keys(newValue).includes(element)) : undefined
+    if (optionsNotEmpty && enabled === true && method && url && data?.length != 0) {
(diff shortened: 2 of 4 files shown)


ajmeireles:
I believe it's a good idea, but it should be used with caution and we need to preserve this information. The custom debounce idea is brilliant too. I approve!

HassanZahirnia:
> I believe it's a good idea, but it should be used with caution and we need to preserve this information. The custom debounce idea is brilliant too. I approve!

Hey @ajmeireles 😄 Thanks for your feedback! 
Which areas/situations do you think we should use "caution" ? Let me know and maybe it's something we can improve!

ninjaparade:
I think this is a really good idea but for me, and maybe others, this kind of circumvents the patterns that Inertia brings which is to leave the "logic" to the backend. 

Similar to how we no longer have to worry about permissions on the frontend, leaving that to the backend to handle and not be duplicated, validation is another thing that I think needs to stay on the backend.

ajmeireles:
> > I believe it's a good idea, but it should be used with caution and we need to preserve this information. The custom debounce idea is brilliant too. I approve!
> 
> Hey @ajmeireles smile Thanks for your feedback! Which areas/situations do you think we should use "caution" ? Let me know and maybe it's something we can improve!

I think that for some it can be negative when analyzing the context of other problems that assume the same proposal (like Livewire): excess requests.

Maybe... you add hooks to realtime validation? Something like: only submit the request after 3 seconds of input (like Debounce) of the value or if value has more than a specific size.

HassanZahirnia:
@ajmeireles This PR is already making use of the `debounce` function ( with a default value of `150ms` ). I also believe anyone who wants to get into real time validation should understand that this feature generally comes with a resource cost ( a roundtrip to the backend with every input change ).

ajmeireles:
> @ajmeireles This PR is already making use of the `debounce` function ( with a default value of `400ms` ). I also believe anyone who wants to get into real time validation should understand that this feature generally comes with a resource cost ( a roundtrip to the backend with every input change ).

I understand, but it would be interesting to allow at least diversification of uses, not just a fixed pattern. At least that's what I think.

RobertBoes:
To share my two cents, I don't think this should be part of Inertia to be honest. The reason being, this is a very opinionated approach, which imo doesn't provide good UX, so I'd end up creating my own approach anyway. To give some context:
- There are already approaches out there, like [Juhlinus/kakunin](https://github.com/Juhlinus/kakunin)
- Everything is validated at once. If you have a form with 3 fields, after typing just a single letter in the first field the validation is going to be fired off, marking at least 2 fields with errors (this is "fixed" by altering the validation rules, see below)
- Instead of watching the form object, devs might want to use a `keyup` or `blur` event on inputs, hence this won't be useable for me (and perhaps more people) because imo it's a UX nightmare

As for the Laravel PR:
- It does not work with FormRequests, so if you want to use this feature you'd have to ditch FormRequests
- The `Inertia::validate()` does not validate, so you'd need an additional validator to actually validate your requests (duplicate code)
- The rules are modified, the `sometimes` rule is appended. Likely to prevent the validation to show warnings for fields that have not yet been filled in. However, that also means it doesn't have to be filled in during validation. Thus it won't show errors, then you submit and it shows errors for all empty fields. You fill in one field and all errors disappear, making the user think they can submit the form. Seems like a tricky workaround to me
- There's no method of setting custom validation messages
- No support for [error bags](https://inertiajs.com/validation#error-bags), how's this going to work with multiple forms where Inertia relies on error bags?

These are some of the things I noticed by quickly looking through the code and testing it locally. It's a great start, it could easily be published as a package. But, to me it looks too opinionated, even if the shortcomings/bugs I mentioned are fixed

HassanZahirnia:
Hey @RobertBoes 👋 Thanks for sharing your thoughts and feedback! 😄
> To share my two cents, I don't think this should be part of Inertia to be honest. The reason being, this is a very opinionated approach, which imo doesn't provide good UX, so I'd end up creating my own approach anyway. To give some context:

I understand that this can be taken as an opinionated approach, although I tried to go with an implementation that provides the most "standard/general" real-time validation results. Take Livewire's [real-time validation](https://laravel-livewire.com/docs/2.x/input-validation#real-time-validation) for instance, one could say that is also opinionated and should not be part of core, but I believe every major framework needs to start having these features.

> There are already approaches out there, like Juhlinus/kakunin

I just found out about Linus's package just yesterday, I don't think it's development is very active anymore, I had a look at it and it really took a different approach from my version. I tried to go with a minimal approach that provides most results

> Instead of watching the form object, devs might want to use a keyup or blur event on inputs, hence this won't be useable for me (and perhaps more people) because imo it's a UX nightmare

Maybe this can be tweaked a little... I thought watching the `form.data()` should be logical since any sort of events that change the form data probably means the data needs to validate.

> It does not work with FormRequests, so if you want to use this feature you'd have to ditch FormRequests

I have not yet thought about adding support for form requests tbh. Thanks for pointing that out.

> The Inertia::validate() does not validate, so you'd need an additional validator to actually validate your requests (duplicate code)

Can you explain a bit more ? I didn't want the `Inertia::validate` to validate anything directly, but have the user pass the validation rules like any other situation and leave the validation upto laravel.

> The rules are modified, the sometimes rule is appended. Likely to prevent the validation to show warnings for fields that have not yet been filled in. However, that also means it doesn't have to be filled in during validation. Thus it won't show errors, then you submit and it shows errors for all empty fields. You fill in one field and all errors disappear, making the user think they can submit the form. Seems like a tricky workaround to me

Yeah the usage of `sometimes` rule was more like a shortcut. I should work on how the rules are processed.

> There's no method of setting custom validation messages

Correct. Noted.

> No support for error bags, how's this going to work with multiple forms where Inertia relies on error bags?

I believe Inertia forms error bags are scoped ( as mentioned in the docs ). So I thought there is no need for that ?

reinink:
@HassanZahirnia Hey, thanks so much for sharing this idea! However, I tend to agree with @RobertBoes that this is a little too opinionated to be included in Inertia core. And if I'm totally honest, I kind of dislike real-time validation — I personally prefer to have validation messages shown after hitting a submit button.

Sorry, I hate closing big PRs like this that obviously took a lot of work — please understand that I appreciate your efforts even if it's not something I think should be included in the library.

Hope that makes sense 🙏

PR #1099, opened 2022-02-15
Change to preferring element.scrollTo

### LINKED ISSUE #1079: Navigation scroll position reset not working on Firefox
### Versions:

- `@inertiajs/inertia-vue3` version: 0.6.0

### Describe the problem:

I'm having an issue with the scrolling on navigation. On Firefox, clicking an InertiaLink results in the scroll-to-top behaviour not working at all (next page loads with the user still positioned at the bottom of the page), or only climbing halfway up the page.

### Steps to reproduce:

- Create page with some decent amount of vertical height.
- Use a persistent layout
- Use Firefox
- Click an Inertia link

craigrileyuk:
Scroll position can fail to update in Firefox when setting it via `element.scrollTop` and `element.scrollLeft`. Change to preferring `element.scrollTo` with a fallback for the previous method to ensure backwards compatibility.

Resolves #1079

### DIFF
--- a/packages/inertia/src/router.ts
+++ b/packages/inertia/src/router.ts
@@ -79,8 +79,9 @@ export class Router {
   protected resetScrollPositions(): void {
     window.scrollTo(0, 0)
     this.scrollRegions().forEach(region => {
-      if (canScrollTo) region.scrollTo({ top: 0, left: 0 })
-      else {
+      if (typeof region.scrollTo === 'function') {
+        region.scrollTo(0, 0)
+      } else {
         region.scrollTop = 0
         region.scrollLeft = 0
       }
--- a/packages/inertia/src/router.ts
+++ b/packages/inertia/src/router.ts
@@ -97,10 +97,13 @@ export class Router {
     if (this.page.scrollRegions) {
       this.scrollRegions().forEach((region: Element, index: number) => {
         const scrollPosition = this.page.scrollRegions[index]
-        if (scrollPosition && canScrollTo) {
-          region.scrollTo({ top: scrollPosition.top, left: scrollPosition.left })
+        if (! scrollPosition) {
+          return
         }
-        else if (scrollPosition) {
+
+        if (canScrollTo) {
+          region.scrollTo(scrollPosition.top, scrollPosition.left)
+        } else {
           region.scrollTop = scrollPosition.top
           region.scrollLeft = scrollPosition.left
         }
--- a/packages/inertia/src/router.ts
+++ b/packages/inertia/src/router.ts
@@ -77,11 +77,18 @@ export class Router {
   }
 
   protected resetScrollPositions(): void {
-    document.documentElement.scrollTop = 0
-    document.documentElement.scrollLeft = 0
+    const canScrollTo = typeof document.documentElement.scrollTo === 'function'
+    if (canScrollTo) document.documentElement.scrollTo({ top: 0, left: 0 })
+    else {
+      document.documentElement.scrollTop = 0
+      document.documentElement.scrollLeft = 0
+    }
     this.scrollRegions().forEach(region => {
-      region.scrollTop = 0
-      region.scrollLeft = 0
+      if (canScrollTo) region.scrollTo({ top: 0, left: 0 })
+      else {
+        region.scrollTop = 0
+        region.scrollLeft = 0
+      }
     })
     this.saveScrollPositions()
     if (window.location.hash) {
@@ -90,10 +97,14 @@ export class Router {
   }
 
   protected restoreScrollPositions(): void {
+    const canScrollTo = typeof document.documentElement.scrollTo === 'function'
     if (this.page.scrollRegions) {
       this.scrollRegions().forEach((region: Element, index: number) => {
         const scrollPosition = this.page.scrollRegions[index]
-        if (scrollPosition) {
+        if (scrollPosition && canScrollTo) {
+          region.scrollTo({ top: scrollPosition.top, left: scrollPosition.left })
+        }
+        else if (scrollPosition) {
           region.scrollTop = scrollPosition.top
           region.scrollLeft = scrollPosition.left
         }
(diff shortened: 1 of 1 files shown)


packages/inertia/src/router.ts
```diff
@@ -77,11 +77,18 @@ export class Router {
...
 
   protected resetScrollPositions(): void {
-    document.documentElement.scrollTop = 0
-    document.documentElement.scrollLeft = 0
+    const canScrollTo = typeof document.documentElement.scrollTo === 'function'
+    if (canScrollTo) document.documentElement.scrollTo({ top: 0, left: 0 })
+    else {
+      document.documentElement.scrollTop = 0
+      document.documentElement.scrollLeft = 0
+    }
```
RobertBoes, line 85: The window API seems to be full supported, so changing this to window would eliminate the need to check if you can use `scrollTo`, see https://caniuse.com/mdn-api_window_scrollto
The options parameter has less support, so it would probably be best to use `scrollTo(0, 0)`, see https://caniuse.com/mdn-api_window_scrollto_options_top_parameter

```suggestion
    window.scrollTo(0, 0)
```

packages/inertia/src/router.ts
```diff
@@ -77,11 +77,18 @@ export class Router {
...
     this.scrollRegions().forEach(region => {
-      region.scrollTop = 0
-      region.scrollLeft = 0
+      if (canScrollTo) region.scrollTo({ top: 0, left: 0 })
+      else {
+        region.scrollTop = 0
+        region.scrollLeft = 0
+      }
```
RobertBoes, line 91:
```suggestion
      if (typeof region.scrollTo === 'function') {
        region.scrollTo(0, 0)
      } else {
        region.scrollTop = 0
        region.scrollLeft = 0
      }
```

packages/inertia/src/router.ts
```diff
@@ -90,10 +97,14 @@ export class Router {
...
       this.scrollRegions().forEach((region: Element, index: number) => {
         const scrollPosition = this.page.scrollRegions[index]
-        if (scrollPosition) {
+        if (scrollPosition && canScrollTo) {
+          region.scrollTo({ top: scrollPosition.top, left: scrollPosition.left })
+        }
+        else if (scrollPosition) {
           region.scrollTop = scrollPosition.top
           region.scrollLeft = scrollPosition.left
         }
```
RobertBoes, line 110: This might be a bit nicer to read
```suggestion
        if (! scrollPosition) {
          return
        }

        if (canScrollTo) {
          region.scrollTo(scrollPosition.top, scrollPosition.left)
        } else {
          region.scrollTop = scrollPosition.top
          region.scrollLeft = scrollPosition.left
        }
```

RobertBoes (changes requested):
Looks good to me.

Left a few notes about browser support. All browsers support `scrollTo(0, 0)` on the window, but not the object syntax as parameter. If Inertia wants to keep the scrolling on the documentElement instead of the window I'd still recommend the usage of `scrollTop(0, 0)`

AlejandroAkbal:
I'm eagerly awaiting this fix to be merged!

RobertBoes:
@reinink is there anything preventing this PR from being merged? It fixes smooth scroll and it seems the issue is also present on Chrome 102 (without smooth scroll). This makes Inertia pretty much unusable since it causes a bad UX see #1192

oliverbj:
Experiencing these issues as well. Hopefully we can get it merged soon.

reinink:
Hey can anyone explain to me how to reproduce this? I followed the steps in #1079 using https://inertiajs.com (which uses a persistent layout) as well as another project of mine, but don't see this issue. I've tried Firefox and Chrome, but it seems to work fine. This feels like a pretty major bug, but before we fix it, I'd love to actually reproduce it.

reinink:
Okay, did some more testing and I was able to reproduce this in Chrome (103.0.5060.53) by setting `scroll-behavior: smooth` on the `<html>` element. Not able to reproduce this without smooth scrolling though.

reinink:
@craigrileyuk thanks for sharing this fix! 🙏

packages/inertia/src/router.ts
```diff
@@ -90,10 +93,17 @@ export class Router {
...
     if (this.page.scrollRegions) {
       this.scrollRegions().forEach((region: Element, index: number) => {
         const scrollPosition = this.page.scrollRegions[index]
-        if (scrollPosition) {
+        if (! scrollPosition) {
+          return
+        }
+
+        if (canScrollTo) {
+          region.scrollTo(scrollPosition.top, scrollPosition.left)
```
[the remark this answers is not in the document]
reinink: Fixed: e22f5bed0780b3c1066af5e4aa7929b3c71a0264

reinink:
Inertia working with `scroll-behavior: smooth` on a scroll region:

https://user-images.githubusercontent.com/882133/176174737-a2215de8-7814-40e1-97f7-444c702a0c91.mov

Pretty neat, but I'm not sure this is a very good idea — these kinds of page animations are super annoying IMHO 😅

reinink:
> But Inertia mimics default browser behavior as much as possible, right?

Yup, which is why I am totally happy to make this work if we can 👍

simon-roland:
Smooth scroll with Inertia broke together with the release of Chrome 102, it's been working fine up till then.

Thanks for merging the changes, really appreciate it.

Unrelated to this issue, I just wanna say fantastic work with SSR support, it's great.

raphaelbeckmann:
The latest version `@inertiajs/inertia @ 0.11.0` hasn't yet the updated `resetScrollPositions()` method.
How can I get the update in my dependencies?

reinink:
Hey folks, I just deployed this fix as `@inertiajs/inertia@0.11.0`.

raphaelbeckmann:
Hey @reinink 
Unfortunately, this bug still exists with `@inertiajs/inertia@0.11.1` + Chrome 108 + `scroll-behavior: smooth`

PR #1102, opened 2022-02-16
closes #1100

### LINKED ISSUE #1100: Link component keeps rendering `href` attribute even when it is not passed or `:href="undefined"`
### Versions:

### Describe the problem:

Link component keeps rendering `href` attribute even when it is not passed or `:href="undefined"`

apparently all Link components rely on `mergeDataIntoQueryString`

this affects at least vue3  & react, but probably others too

### Why this is a problem? 

- cannot properly disable a link, which requires removing `href` attribute completely
- browsers treat `<a href>` as `<a href=".">` (eg link to the current page)

### Reproduction:

https://stackblitz.com/edit/vitejs-vite-irxlcc - vue3

https://stackblitz.com/edit/vitejs-vite-cn2nkk - react

#### option 1

update `mergeDataIntoQueryString` to return href as `undefined` instead of `''` empty string

#### option 2 

update Link component in every lib-specific packages (inertia-vue3, inertia-react, etc)

i am willng to open a pr 🙂

rtabulov:
Closes #1100

### DIFF
--- a/packages/inertia/src/router.ts
+++ b/packages/inertia/src/router.ts
@@ -235,7 +235,7 @@ export class Router {
 
     if (!(data instanceof FormData)) {
       const [_href, _data] = mergeDataIntoQueryString(method, url, data, queryStringArrayFormat)
-      url = hrefToUrl(_href)
+      url = hrefToUrl(_href || '')
       data = _data
     }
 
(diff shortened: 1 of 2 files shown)


reinink:
Hey thanks for this, but I'm pretty certain this was resolved already in #843. You'll just need to make sure you're on the latest version of the Vue adapters 👍

PR #1110, opened 2022-02-24
Feat: configure recently successful delay

RobertBoes:
This allows users to configure the `recentlySuccessful` delay, which currently is set to 2000 with no option to change this. The delay can be set using the second (or third) parameter of creating a form, this config could in the future be used to control other aspects of the form helper

Usage:
```js
const form = useForm({
  name: 'Foo',
}, {
  recentlySuccessfulDelay: 5000,
})

// Or set it after form creation
form.recentlySuccessfulDelay(5000).post('/url')
```


boring-dragon:
This will be a nice option to have for sure!

reinink:

shealavington:
@reinink I'm wondering if it might be good to re-open this as a feature/improvement for v2.0. It's a relatively simple solution too.

Some clients I work with, complain that the default timeout is not accessible enough, and I think this is a good and simple solution to the problem. I have a temporary solution using a custom timeout, but the whole point of the form helper is to avoid the need to keep copy-pasting functionality.

PR #1117, opened 2022-03-03
Feature: new defineLayout composable for Vue 3 <script setup>

lumore:
Hey 👋🏻
First of all, thanks to @reinink and Inertia team for the great tool!

This PR introduces a new `defineLayout` composable.
Motivation is pretty obvious, right now we have to use 2 scripts to define a layout in a script setup component:
```vue
<script setup>
// ...
</script>

<script>
import ApplicationLayout from '@/Layouts/ApplicationLayout.vue'

export default {
  layout: ApplicationLayout,
}
</script>

<template>
  <!-- ... -->
</template>
```

With `defineLayout` we can drop the second script:
```vue
<script setup>
import { defineLayout } from '@inertiajs/inertia-vue3'
import ApplicationLayout from '@/Layouts/ApplicationLayout.vue'

defineLayout(ApplicationLayout)
</script>

<template>
  <!-- ... -->
</template>
```

It also works with nested layouts:
```vue
<script setup>
import { defineLayout } from '@inertiajs/inertia-vue3'
import ApplicationLayout from '@/Layouts/ApplicationLayout.vue'
import SettingsLayout from '@/Layouts/SettingsLayout.vue'

defineLayout([ApplicationLayout, SettingsLayout])
</script>

<template>
  <!-- ... -->
</template>
```

I hope you find this feature useful and would welcome any feedback about it.


RobertBoes:
I feel like this shouldn't be part of Inertia, but rather an unplugin feature so it can be a compiler macro. There is an active RFC for `defineOptions`, but I'm not sure if this also includes custom properties: https://github.com/vuejs/rfcs/discussions/430

It's also made as an unplugin: https://github.com/sxzz/unplugin-vue-define-options

reinink:
@lumore Hey thanks so much for this contribution — I really like it! I am curious where things go with the `defineOptions` API, because that would be obviously preferable, but reading Evan's comments on that RFC I'm skeptical it's going to be adopted. Let's maybe just wait for a bit and see what happens.

lumore:
Haven't seen that `defineOptions` PR, that's definitely a better approach. Alright, let's see how it goes!

lepikhinb:
@reinink, that's such a great QoL PR. I believe it doesn't need to be implemented on the compiler level, as the Inertia way works well here.

It's been 5 months since the RFC was first introduced, yet still no progress there.

reinink:
@lepikhinb Yup, agreed, I'm thinking the approach in this PR is probably the way to go. I'm going to do some testing with it 👍

lumore:
@reinink let me know if you need any help, I would love to!

reinink:
Thanks @lumore! Just gave this a try and it works great. Nice work 👏

mathieutu:
@reinink It seems it has been removed in https://github.com/inertiajs/inertia/commit/329fa115cf4e6c45557856c7ec3d11e8600715d0 without explication. Did you change your mind?

Thanks.

lumore:
@mathieutu This feature was not working with SSR. Take a look at https://github.com/lepikhinb/momentum-layout

mathieutu:
Thanks for the answer and the discovery of Momentum packages!

reinink:
@lumore Yeah, had to remove it because it doesn't work in SSR, and there unfortunately wasn't any way to make it work without a bundler level plugin. The Momentum Layout plugin for Vite is a great choice though! Another option is the [defineOptions plugin](https://vue-macros.sxzz.moe/macros/define-options.html), we mention this one in [the docs](https://inertiajs.com/pages):

image

Would be good to mention the Momentum Layout plugin there as well...

lumore:
@reinink I prefer momentum-layout, that's what I was trying to achieve in this PR, would be nice if you mentioned it in the docs to give users some options 👍🏻

PR #1165, opened 2022-05-03
Add `undefined` as a valid option FormDataConvertable

hailwood:
Doing a reload and sending across `undefined` explicitly for a key in data works, it will remove it from the url.
But the type doesn't allow it

E.g. the following works in straight js
```js
// page is /hello

Inertia.reload({data: {test: 5}});
//page is now /hello?test=5

Inertia.reload({data: {test: null}});
//page is now /hello?test=

Inertia.reload({data: {test: undefined}});
//page is now /hello
```


reinink:
@hailwood dig it — thanks!

PR #1171, opened 2022-05-05
[react] store transform callback in useRef

### LINKED ISSUE #1131: FormHelper transform not working
### Versions:

- `@inertiajs/inertia-react` version: 0.8.0

### Describe the problem:

The ```transform``` function of the form helper does not work.

### Steps to reproduce:

```js
const { data, setData, transform, post } = useForm({
        status: 'default',
        foo: 'bar',
    })

function handleDraft(event) {
        event.preventDefault()
        transform(data => ({ ...data, status: 'draft' }))
        post('/test')
 }

function handleSave(event) {
        event.preventDefault()
        transform(data => ({ ...data, status: 'open' }))
        post('/test')
}
```

```php
Route::post('/test', function (Request $request) {
    return dd($request->all());
});
```

In both cases (```handleSave``` or ```handleDraft``` called) the result from the laravel controller is:
```php
^ array:2 [[▼]()
  "status" => "default"
  "foo" => "bar"
]
```

mrmonat:
This should fix #1131

I think because the old implementation did not persist the callback between renders, the transform method of the form helper had no effect (was reset to default `(data) => data` callback) when called like shown in the example in issue #1131


packages/inertia-react/src/useForm.js
```diff
@@ -16,7 +16,7 @@ export default function useForm(...args) {
   const [progress, setProgress] = useState(null)
   const [wasSuccessful, setWasSuccessful] = useState(false)
   const [recentlySuccessful, setRecentlySuccessful] = useState(false)
-  let transform = (data) => data
+  const transform = useRef((data) => data)
```
emachiels, line 19: Isn't it clearer to use the `useCallback` function? That way you can omit the `.current` parts, which will increase readability. 
https://reactjs.org/docs/hooks-reference.html#usecallback

mxmtsk:
I ran into the same issue and can confirm that this fixes the issue.

ashlimx:
Can we please put priority on this pull request? It's not great the documented transform feature just 100% does not work at all for this long with an active fix available

mahangm:
Is there no update on this PR?

Jamesking56:
This looks like it needs rebasing as it has a bunch of other unrelated changes in it

rsdrsd:
This is a huge update!

mrmonat:
#1491 is newer.

PR #1210, opened 2022-06-30
Add support react <Head> children props

marktan93:
Add support of typescript to <Head> children props


reinink:
Hey now that the React adapter has been converted to TypeScript we no longer have a types file like this. If this is still an issue with the type definitions in the updated adapter, feel free to submit a PR to correct that. Thanks! 🙏

Issue #1211, opened 2022-07-01
Scroll regions inside page component are not restored after form submit.


ragulka:
### Versions:

- `@inertiajs/inertia` version: 0.11.0
- `@inertiajs/inertia-vue3` version: 0.6.0

### Describe the problem:

When a page, for example, an edit form contains a `scroll-region` and we use either the form helper or `Inertia.post` with `{ preserveScroll: true, preserveState: false }`, and then server redirects back to the form after a successful save, the scroll-region scroll position is reset, instead of being preserved.

When the `scroll-region` is outside the page component (in the `Layout` or the blade template, for example), it works as expected - the scroll position is maintained. This is because neither the layout nor the blade template get re-rendered.

Note that the issue only happens when using `preserveState` with either `false` or `'error'`. This is natural, because when the state is preserved, there are no DOM changes. However, when the state is refreshed from the server, the component is re-mounted, DOM changes and this resets the scroll position.

Inertia has a [`restoreScrollPositions()`](https://github.com/inertiajs/inertia/blob/e22f5bed0780b3c1066af5e4aa7929b3c71a0264/packages/inertia/src/router.ts#L95-L109) method, that seems like is built for restoring scroll positions inside scroll regions. However, this is never called after making a form submit.

What happens is that when the request is successful, [`setPage` is called with the correct `preserveScroll` flag](https://github.com/inertiajs/inertia/blob/e22f5bed0780b3c1066af5e4aa7929b3c71a0264/packages/inertia/src/router.ts#L335), but `setPage` itself only uses this flag to _reset_ scroll position to the top of the page, if `preserveScoll: false`.

This relies on the fact that if the document body is scrollable, then browsers would maintain the scroll position. However, browsers will not maintain scroll positions of the individual scroll regions inside the page component.

It seems to me that `restoreScrollPositions()` should be called somewhere in the `setPage` method, after the `swapComponent` resolves.

### Steps to reproduce:

1. Create a `scroll-region` inside a page that has a form (such as an edit page)
2. Ensure that when the form submits, it uses the `{ preserveState: 'errors', preserveScroll: true }` options. 
3. Ensure the server redirects back to the same page
4. Scroll the `scroll-region` inside the page
5. Submit the form
6. Notice how the scroll position inside the `scroll-region` is reset.

For easier reproduction, I've opened a PR on my fork with exact steps: https://github.com/ragulka/pingcrm/pull/1

[...]

PR #1212, opened 2022-07-01
Restore scroll position inside setPage

ragulka:
This is a proposed fix for #1211 

This change will ensure that whenever `setPage` is called, the `preserveScroll` option is respected. Since this option is always passed from current callers to setPage anyway, it should be safe to move the handling to `setPage`. This will centralize `preserveScroll` handling when using `setPage`. There are a few other instances where scroll needs to be restored, which does not involve calling `setPage`, and this PR does not affect them.

Alternatively, we could add a promise resolved handler [here](https://github.com/inertiajs/inertia/blob/e22f5bed0780b3c1066af5e4aa7929b3c71a0264/packages/inertia/src/router.ts#L335), and call `restoreScrollPositions()`, but it seems redundant.


reinink:

PR #1257, opened 2022-08-17
Allow any type for useForm transformation

### LINKED ISSUE #1001: Incorrent type annotation for `useForm().transform`
### Versions:

- `@inertiajs/inertia` version: 0.10.1
- `@inertiajs/inertia-react` version: 0.7.1

### Describe the problem:

The `transform` function has this signature ([src](https://github.com/inertiajs/inertia/blob/master/packages/inertia-react/index.d.ts#L89)):

```ts
export interface InertiaFormProps<TForm = Record<string, any>> {
	transform: (callback: (data: TForm) => TForm) => void
}
```

which requires the transformed data must be the same type as the form data. This is mostly unexpected. The transformed data should be allowed to be any type.

### Steps to reproduce:

```ts
const { transform } = useForm({
    email: '',
    password: '',
    remember: false,
})

transform((data) => ({
    ...data,
    remember: data.remember ? 'on' : '',
}));
```

The code results in the error:

```
Type 'string' is not assignable to type 'boolean'. ts(2322)
Login.tsx(14, 9): The expected type comes from property 'remember' which is declared here on type '{ email: string; password: string; remember: boolean; }'
```

genezys:
Fix #1001


reinink:

PR #1259, opened 2022-08-19
feat: Add `preserveUrl` option

xiCO2k:
> Same PR as #1009 but with a rebase, messed up the other PR somehow 🤦 .

Hey guys, this PR adds an option to preserve the current url, this is specially useful when there is an infinite scroll pagination.

# Example:

```js
const loadMoreIntersect = ref();

const load = () => {
        return;
    }

    Inertia.get(props.posts.next_page_url, {}, {
        preserveState: true,
        preserveScroll: true,
        preserveUrl: true,
        only: ['posts'],
    })
}

onMounted(() => {
    const observer = new IntersectionObserver(
        { rootMargin: '-150px 0px 0px 0px' }
    );

    observer.observe(loadMoreIntersect.value)
});
```

Thanks,
Francisco.


xiCO2k:
Hey @reinink, should I target the `next` branch on this PR?

Thanks.

xiCO2k:
@reinink Should I do a rework on this PR?

PR #1197, opened 2022-06-10
Bump Axios to ^0.27.0

RobertBoes:
Bumps Axios to the latest 0.27.x release. As far as I can see there's no impact for Inertia

Release notes:
- [v0.27.0](https://github.com/axios/axios/releases/tag/v0.27.0)
- [v0.27.1](https://github.com/axios/axios/releases/tag/v0.27.1)
- [v0.27.2](https://github.com/axios/axios/releases/tag/v0.27.2)

Issue #1106, opened 2022-02-22
Excessive calls to `window.history.replaceState` when using a `rememberKey` with `useForm`


WilHall:
### Versions:

- `@inertiajs/inertia` version: 0.10.0
- `@inertiajs/inertia-svelte` version: 0.7.4

### Describe the problem:

Occasionally in our Sentry reporting, we see the following error:

```
SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds: SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds
  at replaceState([native code])
  at replaceState(./node_modules/@inertiajs/inertia/dist/index.js:434:35)
  at saveScrollPositions(./node_modules/@inertiajs/inertia/dist/index.js:145:10)
  at visit(./node_modules/@inertiajs/inertia/dist/index.js:294:16)
  at method(./node_modules/@inertiajs/inertia/dist/index.js:488:69)
  at submit(./node_modules/@inertiajs/inertia-svelte/src/useForm.js:227:17)
  at patch(./node_modules/@inertiajs/inertia-svelte/src/useForm.js:240:12)
  at apply(./app/javascript/pages/setup/savings_profile/PrimaryTransportationMode.svelte:31:11)
  at o(./node_modules/@sentry/browser/esm/helpers.js:87:17)
```

This is due to throttling of calls to `window.history.replaceState` by some browsers. The failure in this situation is not graceful, since depending on when the browser starts throttling _only partial state_ may be persisted.

When using a `rememberKey` with `useForm`, the error occurs on the following line:
https://github.com/inertiajs/inertia/blob/master/packages/inertia-svelte/src/useForm.js#L203

This gets called at least seven times on a successful submission for each modification to the store, in addition to a call any time the form data changes:
https://github.com/inertiajs/inertia/blob/master/packages/inertia-svelte/src/useForm.js#L125-L129

Each call results in a call to `window.history.replaceState` _even if the form data and errors did not change_.

This issue _doesn't exist_ in `inertia-react` because `useRemember` is implemented with a `useEffect` that includes the form data/errors in the dependency array, and thus `Inertia.remember` is only called when the data changes:
https://github.com/inertiajs/inertia/blob/master/packages/inertia-react/src/useRemember.js#L11-L13

I suspect a similar guard around the `Inertia.remember` call in `inertia-svelte` would fix this issue.

### Steps to reproduce:

1. Create a form using `useForm` and pass a string `rememberKey` as the first argument
[...]

PR #1055, opened 2022-01-18
Fix default <head> title tag

claudiodekker:
This PR fixes the `<head>` title tag not defaulting to the `createInertiaApp` when the `title` callback is defined.

PR #1282, opened 2022-09-09
Next

reinink:
This PR introduces a pretty significant architectural change to Inertia from a publishing perspective.

Previously to use Inertia you needed to install the Inertia core package `@inertiajs/inertia` as well as the adapter of your choice, such as `@inertiajs/inertia-react` or `@inertiajs/inertia-vue3`. In addition to this you could also optionally install the `@inertiajs/server` and `@inertiajs/progress` libraries.

This approach was problematic because each library was versioned independently. For example, right now to use Inertia with React, you must be on `@inertiajs/inertia` at `v0.11.0` and `@inertiajs/inertia-react` at `v0.8.1`. This is confusing for end users since it's hard to know at any given point which version of the React adapter is compatible with which version of the core adapter. It's also annoying to manage from a library maintenance perspective, since we need to maintain seven different changelogs and release notes, and it's hard to announce new versions from a marketing perspective since there are so many.

We're changing all this moving forward, and end users will only ever interact with *one* Inertia library — the adapter of their choice. If you use React, you'll install the `@inertiajs/react` package, and will have access to everything you need from there. Technically there will still be a core package that is shared by all the adapters, but that will automatically be installed as a dependency of the adapter library. This will mean that some of the imports for Inertia will change. For example:

```js
// Before
import { Link, Head } from '@inertiajs/inertia-react'

// After
import { router, Link, Head } from '@inertiajs/react'
```

In addition to this we're also synchronizing the version numbers across all the packages — so whether you're using React, Vue or Svelte, everyone will be on the same version number. I'm so looking forward to having *one* changelog!

## Changes

- [x] Rename `@inertiajs/inertia` to `@inertiajs/core`, and make it a dependency of all the adapters.
- [x] Rename `@inertiajs/inertia-react` to `@inertiajs/react`
- [x] Rename `@inertiajs/inertia-vue` to `@inertiajs/vue2`
- [x] Rename `@inertiajs/inertia-vue3` to `@inertiajs/vue3`
- [x] Rename `@inertiajs/inertia-svelte` to `@inertiajs/svelte`
- [x] Merge `@inertiajs/server` to `@inertiajs/core`
- [x] Merge `@inertiajs/progress` to `@inertiajs/core`
- [x] Update `axios` to `v0.27.0` ([#1197](https://github.com/inertiajs/inertia/pull/1197))
- [x] Add new `defineLayout` hook to Vue 3 adapter ([#1117](https://github.com/inertiajs/inertia/pull/1117))
- [x] Add new `defineLayout` hook to Vue 2 adapter (332b1b65aa4cff3fd0057b194a5d25760c51c48f)
- [x] Add new `useForm` hook to Vue 2 adapter (ff591960b2e3eeb51f0813c7f4b34010e40c72c5)
- [x] Add `usePage` hook to Vue 2 adapter
- [x] Get (Vue 2) Cypress test suite passing
- [ ] ~~Add `useRemember` hook to Vue 2 adapter~~
- [ ] ~~Add throttling to fix `router.remember()` issues ([#411](https://github.com/inertiajs/inertia/pull/411), [#1106](https://github.com/inertiajs/inertia/issues/1106), [#1286](https://github.com/inertiajs/inertia/pull/1286))~~

## Fixes

- The `title` tag is now injected by default when one is defined ([#1055](https://github.com/inertiajs/inertia/pull/1055))
- Add `undefined` as a valid `FormDataConvertable` option ([#1165](https://github.com/inertiajs/inertia/pull/1165))


packages/core/src/types.ts
```diff
@@ -0,0 +1,209 @@
...
+  method: Method
+  data: RequestPayload
+  replace: boolean
+  preserveScroll: PreserveStateOption
+  preserveState: PreserveStateOption
+  only: Array<string>
+  headers: Record<string, string>
+  errorBag: string | null
+  forceFormData: boolean
+  queryStringArrayFormat: 'indices' | 'brackets'
```
sebsobseb, line 75: Maybe add `'comma'` too, since it's supported by the `qs` library?

image
https://github.com/ljharb/qs#stringifying

packages/vue2/tests/app/Pages/Links/Data/Object.vue
```diff
@@ -3,13 +3,23 @@
     <span class="text">This is the links page that demonstrates passing data through plain objects</span>
 
     <inertia-link method="GET" href="/dump/get" :data="{ foo: 'get' }" class="get">GET Link</inertia-link>
-    <inertia-link as="button" method="POST" href="/dump/post" :data="{ bar: 'post' }" class="post">POST Link</inertia-link>
+    <inertia-link as="button" method="POST" href="/dump/post" :data="{ bar: 'post' }" class="post"
+      >POST Link</inertia-link
+    >
```
sebsobseb, line 8: Formatting looks weird here; shouldn't it have to be?:

```html
<inertia-link as="button" method="POST" href="/dump/post" :data="{ bar: 'post' }" class="post">
    POST Link
</inertia-link>
```

jcs224:
Now that the core library is being merged into the individual libraries, will it still be possible to create our own client-side frameworks? Like the Mithril library, for example, which uses the core as part of its package:

https://github.com/maicol07/inertia-mithril

hailwood:
@jcs224 

> Technically there will still be a core package that is shared by all the adapters, but that will automatically be installed as a dependency of the adapter library.

packages/core/src/types.ts
reinink: Yes, not necessarily opposed to this, but going to leave this for a subsequent update 👍

packages/vue2/tests/app/Pages/Links/Data/Object.vue
reinink: This is just Prettier doing its horrendous work in Vue.

emargareten:
> Update axios to v0.27.0 (https://github.com/inertiajs/inertia/pull/1197)

what about bumping axios to [`1.x`](https://github.com/axios/axios/releases/tag/v1.0.0)?
latest release is `1.2.2`

reinink:
@emargareten yeah that's probably a good idea, I was considering that but forgot, thanks for the reminder 👍

PR #1286, opened 2022-09-12
Svelte: De-duplicate calls to Inertia.remember

christoomey:
This change attempts to solve the `useForm` `history.replaceState` `SecurityError` described in
https://github.com/inertiajs/inertia/issues/1106 by ensuring we only call `Inertia.remember` when the state has changed.

This should more closely match the React implementation which uses a `useEffect` call with the `state` in the dependencies array. My guess is the Vue3 implementation has similar logic via the `watch` wrapper in that implementation's `useForm`.

Note: There is also https://github.com/inertiajs/inertia/pull/411/files open which attempts to solve this issue by debouncing. Debouncing is likely a more complete solution, but has a tradeoff of potential race conditions that might cause incorrect state values.

### DIFF
--- a/packages/inertia-svelte/src/useForm.js
+++ b/packages/inertia-svelte/src/useForm.js
@@ -200,7 +200,12 @@ function useForm(...args) {
     }
 
     if (rememberKey) {
-      Inertia.remember({ data: form.data(), errors: form.errors }, rememberKey)
+      const currentState = Inertia.restore(rememberKey)
+      const newState = { data: form.data(), errors: form.errors }
+
+      if (!isEqual(currentState, newState)) {
+        Inertia.remember({ data: form.data(), errors: form.errors }, rememberKey)
+      }
     }
   })
 


reinink:
@christoomey This is a pretty interesting solution — but I worry how much this will actually help.

I think most of the time this error occurs when you're typing in a form that uses the remember feature in Inertia, and it updates on each keystroke. Eventually you hit this issue, as it's not too hard to do 100 keystrokes per 30 seconds — however in each case the objects will be different, so an update will be necessary, and you'll still hit this issue.

From your testing, has this implementation actually helped that kind of situation?

christoomey:
@reinink 👋 Thanks for the notes here. In testing our fix did seem to resolve things, but the forms in question for us have fewer text inputs so I was missing that. I actually didn't realize that each keystroke would first a `remember` call, so this solution does not seem like it will work for us after all.

Looking at https://github.com/inertiajs/inertia/pull/1282 I see that you're intending to add throttling at the core Inertia router level which makes sense to me. I'm going to close this now (we'll implement a local throttling for now), and look forward to moving to _✨next✨_ when that is released.

Thanks as always for the work on this fantastic framework 😄

reinink:
@christoomey Hey you're quite welcome! I actually think a combination of throttling and object comparison might ultimately be the way to go here — going to do some more testing, and definitely planning to get this into Inertia core so all the adapters can benefit from it 💪

christoomey:
> I actually think a combination of throttling and object comparison might ultimately be the way to go here
We were actually _just_ working on our vendored fix and came to the same conclusion 🙂 

Excited for this change all the others in the next PR 🚀

PR #1315, opened 2022-10-12
Explicitly add children to InertiaHeadProps type definition

aviemet:
The current type definition for the Head component relies on React.FC, which as of React version 18 does not implicitly include the `children` prop. This results in typescript errors when using the Head component with children as in the example on the Inertia docs site.


reinink:
Hey now that the React adapter has been converted to TypeScript we no longer have a types file like this. If this is still an issue with the type definitions in the updated adapter, feel free to submit a PR to correct that. Thanks! 🙏

PR #1337, opened 2022-11-11
Properly emit events in Vue 3 links

joelvh:
Inertia link events don't work in Vue 3 the way this was implemented. Instead, we need to emit events to trigger handlers.

This faulty behavior caused a regression in our application when we upgraded to Vue 3 and using `emit` solved the problem.


joelvh:
Turns out this is irrelevant and a side effect of using `@vue/compat`.

PR #1344, opened 2022-11-24
Improve Svelte use:inertia and <Link />

pedroborges:
This PR brings a bunch of improvements to the `use:inertia` action and `<Link />` component.

### Stop dispatching the click event ([#640f690](https://github.com/inertiajs/inertia/commit/640f69055c06c34102297cf9ce7165e8de231a6b))

It was a mistake to dispatch the `click` event with `createEventDispatcher` from the action in first place. Its use is [intended for components](https://svelte.dev/docs#template-syntax-component-directives). We ran into an edge-case bug because of this in our own app.

The element where `use:inertia` is used on can still have an `on:click` handler or an event forwarder.

### Dispatch Inertia events on the HTML node ([#00a4fa2](https://github.com/inertiajs/inertia/commit/00a4fa213756c108f8727795f89fd59f2c9b559e))

This allows you to listen to Inertia event callbacks the Svelte way when using `use:inertia`.
  
```svelte
<button
  use:inertia={{ method: 'PUT', href: '/posts' }}
  type="button"
  on:start={({ detail }) => console.log(detail.visit)}
  on:finish={({ detail }) => console.log(detail.visit)}
>
  Create post
</button>
```

All [event callbacks](https://inertiajs.com/manual-visits#event-callbacks) are supported: `on:cancel-token`, `on:before`, `on:start`, `on:progress`, `on:finish`, `on:cancel`, `on:success`, and `on:error`. They also receive the same `detail` parameters as the Inertia event callbacks.
  
### Add `as` prop ([#7ff652a](https://github.com/inertiajs/inertia/commit/7ff652a377d1a1879f745bd40289ed0831a4f02b))

You can now specify which HTML element the `<Link />` component should render.

```svelte
<Link as="a" href="/home">Home</Link>
<Link as="button" href="/home">Home</Link>
```

The `as` prop defaults to the `a` HTML element.

### Remove duplicated logic ([#7ff652a](https://github.com/inertiajs/inertia/commit/7ff652a377d1a1879f745bd40289ed0831a4f02b))

Previously the `<Link />` component and `use:inertia` action had a bunch of duplicated code. Now all this logic only exists on the action. The `<Link />` component uses the action under the hood. This will make it easy to maintain both.

### Forward all Inertia callback events ([#85d6704](https://github.com/inertiajs/inertia/commit/85d67044981932a60061a61c1b5ace6b80c5ac57))

```svelte
<Link
  as="button"
  type="button"
  href="/posts"
  method="PUT"
  on:start={({ detail }) => console.log(detail.visit)}
  on:finish={({ detail }) => console.log(detail.visit)}
>
  Create post
</Link>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
     "packages/*"
   ],
   "dependencies": {
-    "prettier": "^2.7.1"
+    "prettier": "^2.7.1",
+    "prettier-plugin-svelte": "^2.8.1"
   }
 }
--- a/packages/svelte/src/App.svelte
+++ b/packages/svelte/src/App.svelte
@@ -12,9 +12,9 @@
       store.update((current) => ({
         component,
         page,
-        key: preserveState ? current.key : Date.now()
+        key: preserveState ? current.key : Date.now(),
       }))
-    }
+    },
   })
 
   $: child = $store.component && h($store.component.default, $store.page.props)
--- a/packages/svelte/src/Link.svelte
+++ b/packages/svelte/src/Link.svelte
@@ -46,6 +46,14 @@
   on:mouseout
   on:mouseover
   on:mouseup
+  on:cancel-token
+  on:before
+  on:start
+  on:progress
+  on:finish
+  on:cancel
+  on:success
+  on:error
 >
   <slot />
 </svelte:element>
--- a/packages/svelte/src/Link.svelte
+++ b/packages/svelte/src/Link.svelte
@@ -30,7 +30,7 @@
     method,
     replace,
     preserveScroll,
-    preserveState,
+    preserveState: preserveState ?? method !== 'get',
     only,
     headers,
     queryStringArrayFormat,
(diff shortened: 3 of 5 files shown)


FluffyDiscord:
Really nice Svelte-y PR

reinink:
@pedroborges Thanks so much for this. Are there any breaking changes here?

reinink:
@pedroborges Hey I tested this and it seems to work, so I merged in it lol. Would be good to know if there are any breaking changes here. Love that you were able to use `<svelte:element>`.

Also, for the Link event callbacks, do these still won't work with return values, like we discussed in #237? I just tested this out, and returning `false` doesn't seem to cancel the event when using `on:start`:

```svelte
<Link href="/users" on:start={() => false}>Users</Link>
```

Or am I doing this wrong?

pedroborges:
> Thanks so much for this. Are there any breaking changes here?

None that I can think of.

> Love that you were able to use <svelte:element>.

This Svelte feature was only released a while ago. I'm glad the Svelte adapter can now have the `as` prop just as the other adapters 😄 

> Also, for the Link event callbacks, do these still won't work with return values, like we discussed in https://github.com/inertiajs/inertia/pull/237? I just tested this out, and returning false doesn't seem to cancel the event when using `on:start`:

I think this is one downside of this approach, I'll play with it and see if there's any work-around for it.

Issue #1260, opened 2022-08-21
Nested object in useForm must be cloned before assigning them as default.


ravibpatel:
### Versions:

- `@inertiajs/inertia` version: 0.11.0
- `@inertiajs/inertia-vue3` version: 0.6.0

### Describe the problem:

Nested object in useForm must be cloned before assigning them as default.

### Steps to reproduce:

I have a form object like the following.

```js
const props = defineProps({
    fields: {
        type: Object,
        required: true,
    },
});

const myForm = useForm({
    field1: null,
    field2: true,
    field3: {},
});
```

Now I have dynamic fields that I have to get from passed prop object, so I fill it in field3 object.

```js
onBeforeMount(() => {
    props.fields.forEach((field) => {
	myForm.field3[field.label] = field.value;
    });

    // Assiging the object without cloning causes issues.
    myForm.defaults('field3', myForm.field3);
    // This doesn't work too.
    myForm.defaults();

    // So I have to do clone before assigning it.
    myForm.defaults('field3', cloneDeep(myForm.field3));
});

```

### Potential Fix:

If this is an intended behavior, then it is a good idea to show it in the example provided in the docs. I had to source dive to find out what the issue was.

If this is not an intended behavior, then it is a good idea to clone at [this line](https://github.com/inertiajs/inertia/blob/master/packages/vue3/src/useForm.js#L35-L43).

```js
defaults(key, value) {
  if (typeof key === 'undefined') {
	defaults = cloneDeep(this.data())
  } else {
	defaults = Object.assign(
	  {},
	  cloneDeep(defaults),
	  value ? ({ [key]: cloneDeep(value) }) : cloneDeep(key),
	)
  }

  return this
},
```

ravibpatel:
@reinink If you consider adding this, then I like to open another pull request with this change.

reinink:
Hey! Thanks so much for your interest in Inertia.js and for sharing this issue/suggestion.

In an attempt to get on top of the issues and pull requests on this project I am going through all the older issues and PRs and closing them, as there's a decent chance that they have since been resolved or are simply not relevant any longer. My hope is that with a "clean slate" me and the other project maintainers will be able to better keep on top of issues and PRs moving forward.

Of course there's a chance that this issue is still relevant, and if that's the case feel free to simply submit a new issue. The only thing I ask is that you please include a super minimal reproduction of the issue as a Git repo. This makes it much easier for us to reproduce things on our end and ultimately fix it.

[...]

PR #1350, opened 2022-11-27
Fixed an issue that causes nested value in defaults to be overridden when data changes.

ravibpatel:
Please check out #1260 for detailed explanation of this issue.


pedroborges:
I tried the steps from #1260 and couldn't reproduce the issue. Closing this for now 👍

PR #1360, opened 2022-12-14
[1.x] preserveScroll should be true on initial page visit

addgod:
Preserve scroll should be true on initial page visit to mimic normal browser behavior. If a user has scrolled down the page, and then reloads, the page should stay where it is, so we cannot scroll to top, ergo we must preserve scroll.


saas786:
Upon the initial page load, if my URL has #some-element-id, it scrolls down to that specific element. After a short delay, it scrolls back to the top (or saved scroll position), which is an unusual behavior.

Using preserveScroll: true stops the scroll down to the element and back to the top.

Ideally, it should either not scroll down to #some-element-id, or if it does, it should not scroll back to the top or the previously saved scroll position. Since it is the first page load, on subsequent requests using InertiaJS Link, it can continue to behave the same way.

pedroborges:
Related: #1782

pedroborges (approved):
Thanks @addgod!

joetannenbaum (approved):
Looks good, thank you!

PR #1373, opened 2023-01-07
Simplify `usePage()` hook in Vue 3 adapter

reinink:
This PR simplifies the `usePage()` hook in the Vue 3 adapter. We use to run `computed()` on each value in the `page` object (in the root `<App>` component):

```js
export function usePage() {
  return {
    props: computed(() => page.value.props),
    url: computed(() => page.value.url),
    component: computed(() => page.value.component),
    version: computed(() => page.value.version),
  }
```

In hindsight I honestly don't know why we did this. As far as I can tell you still always need to use `usePage()` in conjunction with a `computed()` call within your own components anyway, so this is redundant. Here's what you need to do today:

```js
import { computed } from 'vue'

const appName = computed(() => usePage().props.value.appName)
                                          //  ^
                                          //  |- I hate that this .value is required
```

This would be worth it if you didn't need to still use `computed()` in your own components. But since you do, we might as well simplify `usePage` to this:

```js
export function usePage() {
  return page.value
}
```

Doing that makes this hook easier to consume in your own components, because you no longer need to know that `.value` is required after each top level property:

```js
import { computed } from 'vue'

const appName = computed(() => usePage().props.appName)
                                          //  ^
                                          //  |- No more .value required!
```

To me this is *much* better, but it's a breaking change. Obviously we're on the brink of an official 1.0 release, so now is the time to do it. It just means a slightly more difficult upgrade process for folks.

The upgrade process would just involve removing `.value` from their `usePage()` calls:

```diff
- const appName = computed(() => usePage().props.value.appName)
+ const appName = computed(() => usePage().props.appName)
```

Worth it? 🤔

### DIFF
--- a/packages/vue3/src/app.ts
+++ b/packages/vue3/src/app.ts
@@ -1,16 +1,5 @@
-import { createHeadManager, Page, PageProps, router } from '@inertiajs/core'
-import {
-  computed,
-  ComputedRef,
-  DefineComponent,
-  defineComponent,
-  h,
-  markRaw,
-  Plugin,
-  PropType,
-  ref,
-  shallowRef,
-} from 'vue'
+import { createHeadManager, Page, router } from '@inertiajs/core'
+import { DefineComponent, defineComponent, h, markRaw, Plugin, PropType, ref, shallowRef } from 'vue'
 import remember from './remember'
 import { VuePageHandlerArgs } from './types'
 import useForm from './useForm'
@@ -126,16 +115,6 @@ export const plugin: Plugin = {
   },
 }
 
-export function usePage(): {
-  props: ComputedRef<PageProps>
-  url: ComputedRef<string>
-  component: ComputedRef<string>
-  version: ComputedRef<string | null>
-} {
-  return {
-    props: computed(() => page.value.props),
-    url: computed(() => page.value.url),
-    component: computed(() => page.value.component),
-    version: computed(() => page.value.version),
-  }
+export function usePage() {
+  return page.value
 }
--- a/playgrounds/vue3/resources/js/Components/Layout.vue
+++ b/playgrounds/vue3/resources/js/Components/Layout.vue
@@ -2,7 +2,7 @@
 import { Link, usePage } from '@inertiajs/vue3'
 import { computed } from 'vue'
 
-const appName = computed(() => usePage().props.value.appName)
+const appName = computed(() => usePage().props.appName)
 </script>
 
 <template>


JoseLuisRNP:
@reinink I think this may have some problems, since we are returning only the value at the moment of executing the composable usePage, we lose the possibility of having computeds/watchers that react to changes in the data provided by usePage.

Anyway, this is just a warning since I don't know how the navigation between components works in Inertia and if the data is refreshed by reloading all the data and not mutating the ref this would not be a problem.

Here a simple demo (only recreating the problem, without Inertia) [demo](https://stackblitz.com/edit/vue-n9k9av?file=src%2FApp.vue,src%2FusePage.js)

Thanks for your work, Inertia is great!

timvanuum:
In 99.9% of the cases this would be fine. 
But I agree with @JoseLuisRNP, it feels like going against vue.

On the other hand, I think you can simply do:
`watch(() => page, doSomeThing)`

And the use case to watch page is extremely rare I think.

reinink:
@JoseLuisRNP So from my testing, making the `usePage()` properties computed ahead of time doesn't actually have any benefit — you still always need to use `computed()` in your own components when consuming it.

I've tested this new implementation with Inertia's persistent layout in Vue 3 ([see here](https://github.com/inertiajs/inertia/blob/master/playgrounds/vue3/resources/js/Components/Layout.vue#L5)), and it updated exactly as expected when navigating between pages. That was my main concern as well, but it seems to work perfectly.

That said, I am hoping to get another 1.0 beta out this week, so hopefully folks can battle test this change for me, just in case I missed anything.

Tofandel:
@reinink 

> I've tested this new implementation with Inertia's persistent layout in Vue 3 ([see here](https://github.com/inertiajs/inertia/blob/master/playgrounds/vue3/resources/js/Components/Layout.vue#L5)), and it updated exactly as expected when navigating between pages. That was my main concern as well, but it seems to work perfectly.

The problem with your test is that you're using the `usePage` mixin directly inside a computed which goes against the vue ideology, the mixins should always be called at the top level like this

Old version
```js
const page = usePage()

const pageTitle = computed(() => page.props.value.title)
```

New version
```js
const page = usePage()

const pageTitle = computed(() => page.props.title) // This is not reactive because page has already been dereferenced
```

This effectively creates reactivity problems

nielsquirynen:
Is there any chance this is getting reverted?
The previous usePage implementation seemed more in line with how composables are supposed to be implemented in Vue. Having to use the usePage hook inside a computed seems to go against the best practices.

mathieutu:
I've submitted a [fix](https://github.com/inertiajs/inertia/pull/1469).

mathieutu:
For those who would like to fix this behaviour while waiting for #1469 to be merged, you can do:

```ts
import { Page, PageProps } from '@inertiajs/core'
import { usePage as buggyUsePage } from '@inertiajs/vue3'
import { computed, reactive } from 'vue'

/** @see https://github.com/inertiajs/inertia/pull/1469 */
const usePage = <SharedProps extends PageProps>(): Page<SharedProps> => (
  reactive({
    props: computed(() => buggyUsePage<SharedProps>().props),
    url: computed(() => buggyUsePage<SharedProps>().url),
    component: computed(() => buggyUsePage<SharedProps>().component),
    version: computed(() => buggyUsePage<SharedProps>().version),
    scrollRegions: computed(() => buggyUsePage<SharedProps>().scrollRegions),
    rememberedState: computed(() => buggyUsePage<SharedProps>().rememberedState),
    resolvedErrors: computed(() => buggyUsePage<SharedProps>().resolvedErrors),
  })
)
```

nielsquirynen:
My current workaround:


```
import { usePage } from '@inertiajs/vue3';
import { toRefs } from '@vueuse/core';
import { InertiaPageProps } from '@/utils/types'; // Shared page props containing auth state etc

export const usePageProps = <
  Props extends Record<string, any>,
  Globals = InertiaPageProps,
>() => {
  return toRefs(computed(() => usePage<Globals & Props>().props));
};

// Page.vue
const { count } = usePageProps<{
   count: number;
}>()
 
watch(count, () => { ... })

```

reinink:
Hey folks, these issues have been fixed (see #1469), and have been released in [v1.0.4](https://github.com/inertiajs/inertia/releases/tag/v1.0.4). Thanks for all your input/help! 🙏

PR #1385, opened 2023-01-16
Add InertiaHead and InertiaLink to Vue 2/3 exports

RobertBoes:
The 1.0 release removed the `InertiaLink` and `InertiaHead` components from the exports. Using multi-word component names is considered a good practice by Vue, see https://vuejs.org/style-guide/rules-essential.html#use-multi-word-component-names


sebsobseb (approved):

reinink:
Yeah personally not a fan of `InertiaHead` and `InertiaLink` at all, always regretted that naming, which is why I deprecated it and removed it in v1.0. In Next.js, it's not called a `NextLink`, it's just a `Link`.

The main reason for that Vue recommendation is to prevent collisions with HTML elements, but that's not an issue with Inertia, since you're never using Inertia with in-DOM templates.

And if this really bugs you, you can always do this:

```js
import { Head as InertiaHead } from '@inertiajs/vue'
```
Or go a step further and create your own components that extends these Inertia components.

That's my gut reaction at least :man_shrugging:

reinink:
@RobertBoes Hey, I've thought about this for a couple days and while I totally understand where you're coming from — I too generally prefer to follow style guides whenever possible — in the case of an Inertia app, which will always use SFC, I think it's totally fine to use a title cased single word component, as there won't ever be conflicts. I also think it's weird to have multiple exports for the same components — which one do we document? I really appreciate you bringing this to my attention either way. If it bugs you for your own project, I'd recommend aliasing it 👍

PR #1387, opened 2023-01-16
Adds `clearChangedErrors` method to clear changed fields errors

rodrigopedra:
This PR adds a new method called `clearChangedErrors`.

It will clear errors from any changed fields. 

This is useful when resubmitting a form where a user tried to correct its errors, and validation is performed server-side.

Take this snippet as an example:

```js
function handleSubmit() {
    if (! this.form.isDirty) {
        return alert('No changes made');
    }
    
    // as .clearChangedErrors() returns 
    // the form instance (on Vue 2, Vue 3 and Svelte) 
    // we can check if there are still uncorrected fields
    // while cleaning any changed fields
    if (this.form.clearChangedErrors().hasErrors) {
        return alert('Please correct all fields with errors');
    } else {
        this.form.defaults();
    }

    this.form.post('https://...');
}
```

This PR:
- adds the method `clearChangedErrors` to all current adapters (Vue 2, Vue 3, Svelte, and React) 
- adds a test case to the Vue 2 test suite (the only adapter with a test suite as of the present time)

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -23,6 +23,7 @@ export interface InertiaFormProps<TForm extends Record<string, unknown>> {
   setDefaults(fields: Record<keyof TForm, string>): void
   reset: (...fields: (keyof TForm)[]) => void
   clearErrors: (...fields: (keyof TForm)[]) => void
+  clearChangedErrors: () => void
   setError(field: keyof TForm, value: string): void
   setError(errors: Record<keyof TForm, string>): void
   submit: (method: Method, url: string, options?: VisitOptions) => void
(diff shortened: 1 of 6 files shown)


rodrigopedra:
As a more generic alternative that would allow this check to be performed on user land, we could consider exposing a `fieldIsDirty(field)` (naming is not my *forte*) that would allow checking if a single field is dirty.

If the approach above is preferred, let me know, and I will try sending a PR.

rodrigopedra:
I thought about it, and it doesn't seem to be a good idea, as a user could restore a previously errored out field's value, and the error message would not be restored.

Keeping track of it looks too much of a hassle.

But I still think having a helper to check if a field is dirty would be a nice addition. Let me know if you would be interested in this.

PR #1392, opened 2023-01-17
Prevent needing to use `Method` enum with the `Link` component

### LINKED ISSUE #1384: Typescript error with Link component together with "method"
### Version:

- `typescript` version: 4.9.4

### Describe the problem:

If you use a `<Link` like so:

```js
<Link
    as="button"
    method="post"
```

It gives the following error:
`Type '"post"' is not assignable to type 'Method | undefined'.`

What does work, is using the Method enum like so:

```js
import { Method } from '@inertiajs/core';

<Link
    as="button"
    :method="Method.POST"
```

So if the plain `method="post"` is no longer supported, it should be [reflected in the docs](https://inertiajs.com/links#method).

### Steps to reproduce:

See the description.

jessarcher:
This PR fixes an issue where TypeScript users now need to use the `Method` enum with the `Link` component.

Prior to 1.0, the `Method` enum was only needed when calling `Inertia.visit()` in TypeScript, but in most cases, people were probably using the shortcut methods that bypass the need to specify the `method` argument.

When the adapters were converted to TypeScript, their existing type definitions specified the `method` prop as a `string`, but this was not compatible with the stricter definition that already existed in core. Rather than loosening the definition in core, I opted to make the adapter definitions compatible. In hindsight, this creates an unnecessary burden on TypeScript users.

As suggested in https://github.com/inertiajs/inertia/issues/1384#issuecomment-1383824602, this PR introduces a union type that offers the same type safety but without needing to import the type.

Removing the existing enum would be a breaking change, so this PR just deprecates it for now.

Fixes #1384

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -21,6 +21,7 @@ export type FormDataConvertible =
 
 export type HttpMethod = 'get' | 'post' | 'put' | 'patch' | 'delete'
 
+/** @deprecated */
 export enum Method {
   GET = 'get',
   POST = 'post',
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -19,6 +19,8 @@ export type FormDataConvertible =
   | null
   | undefined
 
+export type HttpMethod = 'get' | 'post' | 'put' | 'patch' | 'delete'
+
 export enum Method {
   GET = 'get',
   POST = 'post',
@@ -69,7 +71,7 @@ export type LocationVisit = {
 }
 
 export type Visit = {
-  method: Method
+  method: HttpMethod | Method
   data: RequestPayload
   replace: boolean
   preserveScroll: PreserveStateOption
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -19,7 +19,7 @@ export type FormDataConvertible =
   | null
   | undefined
 
-export type HttpMethod = 'get' | 'post' | 'put' | 'patch' | 'delete'
+export type VisitMethod = 'get' | 'post' | 'put' | 'patch' | 'delete'
 
 /** @deprecated */
 export enum Method {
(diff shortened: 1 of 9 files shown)


packages/core/src/types.ts
```diff
@@ -19,6 +19,9 @@ export type FormDataConvertible =
   | null
   | undefined
 
+export type HttpMethod = 'get' | 'post' | 'put' | 'patch' | 'delete'
```
jessarcher, line 22: I wonder whether `VisitMethod` would be better? HTTP methods are usually all caps, and this isn't an exhaustive list of HTTP methods either.

innocenzi: Suggestion: you could allow both upper and lowercase: 

```ts
export type HttpVerb = 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE'
export type Method = HttpVerb | Lowercase<HttpVerb>
```

And just use `.toLowerCase()` when you work with a user-given method.

jessarcher: TIL `Lowercase`!

But wouldn't that break the existing usage of `Method` as an enum in userland code?

reinink:
As discussed, decided to just change the existing `Method` enum to a new string based union type. Technically this type was never exported from any of the adapters, so this is considered private API, and not a breaking change.

packages/core/src/types.ts
innocenzi: A bit late, but yeah if users use `Method` themselves for something else than Inertia (eg. they do something like `if (method === 'get')` in their own code, it could break. I'd say that it's unlikely though!

palypster:
Well, except, it was exported (in [@inertiajs/core](https://github.com/inertiajs/inertia/blob/v1.0.0/packages/core/src/index.ts)) and thus it did break our apps :) Do you really find it as a private API? No worries, we easily fixed that, I'm just curious about your view on the "private API" part. Awesome work guys, thank you! <3

PR #1394, opened 2023-01-17
Restore Vue 3 `usePage` generic type

### LINKED ISSUE #1381: usePage types not inheriting @inertiajs/core Page type
### Version:

### Describe the problem:

`@inertiajs/vue3/types/app.d.ts@usePage` types are declared as
```ts
export declare function usePage(): {
    component?: string;
    props?: {
        [x: string]: unknown;
        errors: import("@inertiajs/core").Errors & import("@inertiajs/core").ErrorBag;
    };
    url?: string;
    version?: string;
    scrollRegions?: {
        top: number;
        left: number;
    }[];
    rememberedState?: Record<string, unknown>;
    resolvedErrors?: import("@inertiajs/core").Errors;
};
```
it does not inherit from the `Page` type from `@inertiajs/core` and it causing all instances of `usePage()` to ignore the `PageProps` interface in `@inertiajs/core/types/types.ts`

if extending the `PageProps` interface via 
```ts
  interface PageProps {
    myProp: string
  }
}
```

it does not reflect in `usePage()` typings

jessarcher:
When converting the adapters to TypeScript, I forgot to copy over the existing type definition for `usePage` which used a generic.

This PR attempts to restore this, however, I can't currently get it to build.

Fixes #1381

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -27,7 +27,7 @@ export interface PageProps {
   [key: string]: unknown
 }
 
-export interface Page<SharedProps = PageProps> {
+export interface Page<SharedProps extends PageProps = PageProps> {
   component: string
   props: PageProps &
     SharedProps & {
--- a/packages/react/src/createInertiaApp.ts
+++ b/packages/react/src/createInertiaApp.ts
@@ -9,13 +9,13 @@ type ReactComponent = ReactNode
 type HeadManagerOnUpdate = (elements: string[]) => void // TODO: When shipped, replace with: Inertia.HeadManagerOnUpdate
 type HeadManagerTitleCallback = (title: string) => string // TODO: When shipped, replace with: Inertia.HeadManagerTitleCallback
 
-type AppType<SharedProps = PageProps> = FunctionComponent<
+type AppType<SharedProps extends PageProps = PageProps> = FunctionComponent<
   {
     children?: (props: { Component: ComponentType; key: Key; props: Page<SharedProps>['props'] }) => ReactNode
   } & SetupOptions<unknown, SharedProps>['props']
 >
 
-export type SetupOptions<ElementType, SharedProps> = {
+export type SetupOptions<ElementType, SharedProps extends PageProps> = {
   el: ElementType
   App: AppType
   props: {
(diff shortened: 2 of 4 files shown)


hailwood:
@jessarcher just a note to please check the other adapters too.
I can confirm that the same issue applies to the react adapter.

jessarcher:
> @jessarcher just a note to please check the other adapters too. I can confirm that the same issue applies to the react adapter.

Good call! Looks like https://github.com/inertiajs/inertia/pull/1396 beat me too it though :)

reinink:
@innocenzi Hey! We've got this particular issue fixed in #1396, but we're struggling to know exactly how to do this in the Vue adapters. We're getting this error:

```
src/app.ts:119:3 - error TS2322: Type '{ component: string; props: { [x: string]: unknown; errors: Errors & ErrorBag; }; url: string; version: string; scrollRegions: { top: number; left: number; }[]; rememberedState: Record<string, unknown>; resolvedErrors: Errors; }' is not assignable to type 'Page<T>'.
  Types of property 'props' are incompatible.
    Type '{ [x: string]: unknown; errors: Errors & ErrorBag; }' is not assignable to type 'PageProps & T & { errors: Errors & ErrorBag; }'.
      Type '{ [x: string]: unknown; errors: Errors & ErrorBag; }' is not assignable to type 'T'.
        'T' could be instantiated with an arbitrary type which could be unrelated to '{ [x: string]: unknown; errors: Errors & ErrorBag; }'.

119   return page.value
      ~~~~~~~~~~~~~~~~~


Found 1 error in src/app.ts:119
```

Do you have any recommendations on how to fix this?

innocenzi:
@reinink Looks like the `page` variable is the culprit; it has a generic type of `Page` but the generics are not constrained. I think using `const page = ref<Page<any>>()` would fix it. But I also recommend constraining the generics of `usePage`:

```ts
// in Vue
export function usePage<T extends PageProps>(): Page<T> {
  return page.value
}

// and in core
export interface Page<SharedProps extends PageProps = PageProps> {
  // ...
}
```

Libero-Software:
Any news on this?

JustInToCoding (changes requested):

JustInToCoding (approved):

PR #1397, opened 2023-01-17
Fix: export types correctly

christian98:
The "exports" field of the package.json was misused for the type-declarations. Also, the type-declarations of secondary entrypoint (server.ts) of the platform packages was not generated.

Docs:
- https://2ality.com/2021/06/typescript-esm-nodejs.html#mapping-a-directory-to-a-file
- https://dev.to/binjospookie/exports-in-package-json-1fl2

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -38,7 +38,9 @@
   },
   "typesVersions": {
     "*": {
-      "server": ["types/server.d.ts"]
+      "server": [
+        "types/server.d.ts"
+      ]
     }
   },
   "scripts": {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -26,10 +26,12 @@
   "types": "types/index.d.ts",
   "exports": {
     ".": {
+      "types": "./types/index.d.ts",
       "import": "./dist/index.esm.js",
       "require": "./dist/index.js"
     },
     "./server": {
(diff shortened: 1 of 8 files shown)


jessarcher:
I've just rebased this with the latest changes from `master` to confirm everything is working and to get the new CI checks to run.

I've confirmed it's an improvement and fixes the issue with the server type exports (evidenced by needing to remove two `@ts-expect-error` directives in the Vue 3 playground :tada:)

jessarcher (approved):

jessarcher:
It has been brought to my attention that the package.json `exports` _is_ a valid way to provide type declarations, and it has priority over `typesVersions`.

https://github.com/microsoft/TypeScript/pull/50890

I'm going to submit an alternative version of this PR that should fix the server export issues.

jessarcher:
It appears I was too hasty in closing this PR. The `exports.*.types` feature is only for projects with `moduleResolution` set to `node16` or `nodenext`, and while it would be nice to move towards that, I don't want it to delay this fix any longer.

jessarcher:
I've restored the `exports.*.types` fields, but fixed the file extension on them, and also moved them to the top of their respective sections per https://nodejs.org/api/packages.html#community-conditions-definitions

Issue #1082, opened 2022-02-04
Custom component for InertiaLink component


bobbypiper:
### Discussed in https://github.com/inertiajs/inertia/discussions/998

<div type='discussions-op-text'>

<sup>Originally posted by **ZeoKnight** December  8, 2021</sup>
Can this line be removed from InertiaLink in React? With the `as.toLowerCase()` included, this blocks the ability to render as custom components (perhaps using a UI framework etc).

With the line removed I can easily do;
```
<InertiaLink as={Button} >Foo</InertiaLink>
```

Thoughts?

https://github.com/inertiajs/inertia/blob/7fcab0e8411e67d93942a4f44cf426670cc7939f/packages/inertia-react/src/Link.js#L75

</div>

ibnu-ja:
+1
not just react, but Vue can benefit from this. UI lib like Vuetify does not support InertiaLink atm

BuyOwnEx:
+1 
add this feature, please

reinink:
Hey! Thanks so much for your interest in Inertia.js and for sharing this issue/suggestion.

In an attempt to get on top of the issues and pull requests on this project I am going through all the older issues and PRs and closing them, as there's a decent chance that they have since been resolved or are simply not relevant any longer. My hope is that with a "clean slate" me and the other project maintainers will be able to better keep on top of issues and PRs moving forward.

Of course there's a chance that this issue is still relevant, and if that's the case feel free to simply submit a new issue. The only thing I ask is that you please include a super minimal reproduction of the issue as a Git repo. This makes it much easier for us to reproduce things on our end and ultimately fix it.

Really not trying to be dismissive here, I just need to find a way to get this project back into a state that I am able to maintain it. Hope that makes sense! ❤️

PR #1398, opened 2023-01-17
Add ability to use a custom component with <Link>

glafarge:
Adding the option to use either a string **or a React component** as a rendering option for `<Link>` component using `as` attribute.
This allows for greater customization, as seen in the examples below:

**Use HTML tag as a wrapper**
_(current implementation)_
```jsx
import { Link } from '@inertiajs/react'

<Link href="/logout" method="post" as="button" type="button">Logout</Link>

// Renders as... 
<button type="button">Logout</button>
```

**Use a custom component**
```jsx
import { Link } from '@inertiajs/react'
import { MyCustomButton } from "./buttons/my-custom-button";

<Link href="/logout" method="post" as={MyCustomButton} foo="bar">Logout</Link>

// Renders as... 
<MyCustomButton foo="bar">Logout</MyCustomButton>
```

This addresses issue #1082 and improves overall functionality.
Any feedback or suggestions for improvement are welcomed!

### DIFF
--- a/packages/react/src/Link.ts
+++ b/packages/react/src/Link.ts
@@ -7,12 +7,12 @@ import {
   router,
   shouldIntercept,
 } from '@inertiajs/core'
-import { createElement, forwardRef, useCallback } from 'react'
+import { createElement, forwardRef, useCallback, ComponentType } from 'react'
 
 const noop = () => undefined
 
 interface BaseInertiaLinkProps {
-  as?: string
+  as?: string | ComponentType,
   data?: Record<string, FormDataConvertible>
   href: string
   method?: Method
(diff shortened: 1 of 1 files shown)


jessarcher:
Wouldn't it be cleaner to make `MyCustomButton` wrap the `Link` component so that you don't need two imports when using it? The button component could conditionally render a `Link` or `button` depending on whether the `href` prop has been provided.

mathieutu:
@jessarcher Same issue here.
I have a basic `PrimaryButton` component which has its own "as".
I use it for different usages, and sometimes I would like to use it as a Inertia Link. It make sense to me to be able to pass Link in its `as` props. It's also the pretty standard behavior of such `as` props. (example with EmotionCSS or HeadlessUI).

pedroborges:
Thank you for your contribution, @glafarge! We really appreciate your effort in adding this functionality. We've decided to proceed with #1940, which provides the same feature.

Issue #1402, opened 2023-01-19
Typescript definitions missing


TeaDrinkingProgrammer:
### Version:

- `@inertiajs/vue3` version: 1.0.0

### Describe the problem:

I have upgraded from Inertia 0.x to 1.0 and have found a difference that isn't mentioned in the docs: a lot of type definitions are not exported when they previously were. I have issues with these ones, but there might be more that are missing:
- InertiaFormProps
- InertiaForm (useful for typing a form passed as a prop)

I can see the type definitions in the types folder, they are just not exported.

A sidenote regarding this, is that it isn't clear that there is now a core and a platform (e.g vue3) submodule in the new Inertia dependency, not just the platform-specific module when upgrading to 1.0 in the docs.

saibotk:
As a workaround you can try to access some of those typings from the return types of eg. `useForm`.

A quick example to access the InertiaForm type:
```ts
import type { useForm } from '@inertiajs/vue3'

type InertiaForm<T extends Record<string, unknown>> = ReturnType<typeof useForm<T>>
```

BR0kEN-:
Another option is to add the following to, for instance, project-level `types.d.ts`.

```typescript
declare module '@inertiajs/vue3' {
  import type { InertiaForm } from '@inertiajs/vue3/types/useForm';

  export { InertiaForm };
}
```

reinink:
Hey! Thanks so much for your interest in Inertia.js and for sharing this issue/suggestion.

In an attempt to get on top of the issues and pull requests on this project I am going through all the older issues and PRs and closing them, as there's a decent chance that they have since been resolved or are simply not relevant any longer. My hope is that with a "clean slate" me and the other project maintainers will be able to better keep on top of issues and PRs moving forward.

Of course there's a chance that this issue is still relevant, and if that's the case feel free to simply submit a new issue. The only thing I ask is that you please include a super minimal reproduction of the issue as a Git repo. This makes it much easier for us to reproduce things on our end and ultimately fix it.

Really not trying to be dismissive here, I just need to find a way to get this project back into a state that I am able to maintain it. Hope that makes sense! ❤️

PR #1413, opened 2023-01-26
Feat(vue): Allowing stronger type support of useForm objects.

Nesuarg:
By exporting InertiaForm type, we can have stronger typing of useForm.

in my case I would like to do:
`type TwoWheeler = InertiaForm<MyBike> | InertiaForm<MyScooter>;`

This allow me to exclude objects like "MyCar"


BR0kEN-:
Fixes https://github.com/inertiajs/inertia/issues/1402

timoransky:
I would also love to see this merged since it kinda broke all our types during the upgrade to 1.0 😞

jessarcher:
Thanks! I've made the same change to the Vue 2 adapter as well to help keep them consistent.

PR #1427, opened 2023-02-03
Convert React Playground to TypeScript

jessarcher:
This PR converts the React playground to TypeScript in order to make it easier to test Inertia's TypeScript support and replicate any issues.

This was previously done for Vue 3 in #1399.


jessarcher:
This PR is currently a draft because there is a type issue in `ssr.tsx` that I'm having trouble solving.

PR #1441, opened 2023-02-09
Export `axios` as httpClient for general use

zcuric:
### This PR
- [x] Exports `axios` from inertia build to users for general use

We found ourselves installing `axios` even though it is a core library used in Inertia so exposing it to users would remove unnecessary installing and users would just import httpClient from Inertia core.
Example: 

```js
import { httpClient } from '@intertiajs/core'

httpClient.get('/api')
```

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -6,5 +6,6 @@ export { default as shouldIntercept } from './shouldIntercept'
 export * from './types'
 export { hrefToUrl, mergeDataIntoQueryString, urlWithoutHash } from './url'
 export { type Router }
+export { default as httpClient } from 'axios'
 
 export const router = new Router()


woosade:
That would be so helpful

jessarcher:
Hey! I'm not sure about this one - I think if your project has a direct dependency on using axios then it should be reflected in your `package.json`.

zcuric:
@jessarcher Thanks for the response! My reasoning behind this is that most apps built with Inertia usually have some API calls that need regular requests and that it would be neat that Intertia would provide direct export for httpClient that it uses anyway without the user needing to think about other libraries. Interia requests already wrap up axios so why not give also barebone axios methods to users anyway. This could be considered a utility provided by Intertia. 

But also, I understand your viewpoint.

Ifriqiya:
Doesn't Laravel already come with axios and is available globally-see bootstrap file? I use axios in my vue SFCs without having to first import it. Or is this different?

taozuhong:
@zcuric finally, what is the solution for this? could you share the final solution?

PR #1396, opened 2023-01-17
Make usePage generic again

Bloemendaal:

PR #1451, opened 2023-02-15
Improve React `usePage` generic type

jessarcher:
In #1396, the `usePage` generic type was (re)introduced for the React adapter.

When trying to use this, I've found it to be quite inconvenient because it's expecting us to pass in an entire `Page` type, which has things like the `url`, `component`, etc. with the props being nested under a `props` key.

This PR aims to solve that by changing the generic type to just accept the `PageProps` type.

This would also make the usage consistent with the Vue version.

### DIFF
--- a/packages/react/src/usePage.ts
+++ b/packages/react/src/usePage.ts
@@ -1,8 +1,8 @@
-import { Page } from '@inertiajs/core'
+import { Page, PageProps } from '@inertiajs/core'
 import { useContext } from 'react'
 import PageContext from './PageContext'
 
-export default function usePage<TPage extends Page = Page>(): TPage {
+export default function usePage<TPageProps extends PageProps = PageProps>(): Page<TPageProps> {
   const page = useContext(PageContext)
 
   if (!page) {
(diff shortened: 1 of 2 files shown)


jessarcher:
@Bloemendaal, @hailwood - I'd appreciate your thoughts on this one :slightly_smiling_face:

zzzet:
with [release 1.0.1](https://github.com/inertiajs/inertia/pull/1396) in React
```typescript
import type { Page } from "@inertiajs/core";
const { appName } = usePage<Page<{ appName: string }>>().props
```

vs with https://github.com/inertiajs/inertia/pull/1451

```typescript
const { appName } = usePage<{ appName: string }>().props
```

this is convenient and less code 👍🏻

Bloemendaal:
Looks like an improvement to me! 👍 

Just wondering if there is any way (or reason) to extend the `Page` interface itself, since this is a property in the `createInertiaApp` object...

hailwood:
Yep that's exactly how I'd expect to use it 👍
Very nice 🙂

PR #1464, opened 2023-02-25
Add `preserveUrl` option.

xiCO2k:
This PR adds an option to preserve the current url. 

This is specially useful when there is an **infinite scroll pagination**.

## Before

To the infinite scrolls using Inertia we had to do something similar to this:

```js
const nextPageUrl = ref(props.posts.next_page_url || null);
const loading = ref(false);

const load = () => {
    if (nextPageUrl.value === null) {
        return;
    }

    loading.value = true;
    axios.get(nextPageUrl.value, {
        headers: {
            'X-Inertia': true,
            'X-Inertia-Partial-Component': 'Posts/Index',
            'X-Inertia-Partial-Data': 'posts',
            'X-Inertia-Version': usePage().version,
        },
    }).then(({ data }) => {
        nextPageUrl.value = data.props.posts.next_page_url;
        allPosts.value.push(...data.props.posts.data);
        loading.value = false;
    }).catch(() => {
        loading.value = false;
        router.get(nextPageUrl.value);
    })
}

let observer;

onMounted(() => {
    observer = new IntersectionObserver(
    );

    observer.observe(document.querySelector('footer'));
});
```

## After

This this addiction, we can take advantage of the `preserveUrl` option, and pass it on the `router.get` method, like this:

```js

const load = () => {
        return;
    }

    router.get(props.posts.next_page_url, {}, {
        preserveState: true,
        preserveScroll: true,
        preserveUrl: true,
        only: ['posts'],
    })
}

onMounted(() => {
    observer = new IntersectionObserver(
    );

    observer.observe(document.querySelector('footer'));
});
```

Thanks,
Francisco.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3841,7 +3841,7 @@
       "requires": {
         "@inertiajs/react": "^1.0.0-beta.2",
         "@types/react": "^16.14.35",
-        "@types/react-dom": "^16.9.17",
+        "@types/react-dom": "^16.9.18",
         "@vitejs/plugin-react": "^3.0.0",
         "autoprefixer": "^10.4.13",
         "axios": "^1.1.2",
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -162,9 +162,9 @@ export class Router {
     })
   }
 
-  protected locationVisit(url: URL, preserveScroll: LocationVisit['preserveScroll']): boolean | void {
+  protected locationVisit(url: URL, preserveScroll: LocationVisit['preserveScroll'], preserveUrl: LocationVisit['preserveUrl']): boolean | void {
     try {
-      const locationVisit: LocationVisit = { preserveScroll }
+      const locationVisit: LocationVisit = { preserveScroll, preserveUrl }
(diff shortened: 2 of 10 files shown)


maxflex:
I am looking exactly for that feature for my infinite scroll functionality

maxflex:
Just dug deeper into the source code. Though the solution offered in this commit might work as a workaround, it is not comprehensive enough to be merged. 

Overall – just too much unwanted code is executed to perform this simple task. Here are just some bugs:

Image

- Route active state breaks (_see bottom nav_) probably due to [link format change](https://github.com/inertiajs/inertia/pull/1464/commits/e3d9a4bb26b61bbd7764d24c29de8e180b9558c3#diff-243439b8403c7b4ea2340ee5cfdba2ef9851f1a7e54d66bbc5805583f21140ecR468) (_see before-after in console_) 
- Buggy browser back behaviour. This can be solved by adding `replace: false` option. Those should be mutually exclusive
options

But then again all those default options all the time still resulting in a lot of unnecessary code being executed under the hood

xiCO2k:
Thanks for reviewing @maxflex.

If you have any suggestion, I can update my PR.

reinink:

PR #1469, opened 2023-02-27
Fix `usePage()` reactivity in Vue 3 adapter

### LINKED ISSUE #1432: vue3 usePage state not updating on props changing
### Version:

### Describe the problem:

when using `usePage()` it does not return a ref and there for does not update correctly when the page changes 

because `usePage` is returning `page.value` and not the `ref` once you reassign `page.value` the old value will not get updated. 

Example:
```php
use Inertia\Inertia;

Route::get('/test', function () {
  return Inertia::render('Test', [
    'time' => time(),
  ]);
});
```

```vue
<script setup>
  const page = usePage()
  const time = computed(() => page.props.time)
  const newTime = ref(0)

  let timeout
  function reload() {
    timeout = setTimeout(() => {
      console.log(page.props) // { time: "1675567188" }
      console.log(time.value) // "1675567188"

      const page2 = usePage() // call it again now that the props have changed
      console.log(page2.props) // shows true value

      newTime.value = page2.props.time // update newTime ref to show new time

      router.reload()
      clearTimeout(timeout)
      reload()
    }, 5000)
  }

  onMounted(() => {
    reload()
  })
</script>

<template>{{ time }} : {{ newTime }}</template>
```

~~this would be solved by returning the ref not the value, but would be a breaking change due to you needing to access page values via `page.value`~~ I forgot about reactive objects

mathieutu:
Hi,
Thanks for your great work on this package.

## Why
This PR fixes the lack of reactivity of `usePage` function return in Vue3 adapter.

Indeed, by returning `page.value`, we're losing the link with the page ref. 
When the page value is fully replaced, the props don't change in the component.

You can find a full simplified example [here](https://codesandbox.io/p/sandbox/nameless-http-78bf43?selection=%5B%7B%22endColumn%22%3A12%2C%22endLineNumber%22%3A16%2C%22startColumn%22%3A12%2C%22startLineNumber%22%3A16%7D%5D&file=%2Fsrc%2Fcomponents%2FNew.vue) with a comparison before/after.

References: 
- Discussion https://github.com/inertiajs/inertia/pull/1373#issuecomment-1400488273,
- Fix #1432
- [Another demo of the issue](https://stackblitz.com/edit/vue-n9k9av?file=src/App.vue)

## How
- Transform the page object to a reactive
- Set each other page key independently (with `Object.assign`)
- Return the reactive object

Ideally (cf [vue docs](https://vuejs.org/guide/reusability/composables.html#conventions-and-best-practices)), it would have been better to return refs, as it was previously (the [previous behaviour](https://github.com/inertiajs/inertia/blob/41b3c76337bce4ea1a02354af95ab39feec74ac1/packages/vue3/src/app.ts#L136-L139) was the recommended one, unfortunately).
However making this would involve a breaking change. To destructure people can use `toRefs(page)` (see [demo](https://codesandbox.io/p/sandbox/nameless-http-78bf43?selection=%5B%7B%22endColumn%22%3A31%2C%22endLineNumber%22%3A15%2C%22startColumn%22%3A31%2C%22startLineNumber%22%3A15%7D%5D&file=%2Fsrc%2Fcomponents%2FNew.vue))

## Result

```js
const page = usePage()

const { props, component } = toRefs(page)

const posts = computed(() => page.props.posts)
const user = computed(() => props.value.user)
```

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -37,8 +37,9 @@ export interface Page<SharedProps extends PageProps = PageProps> {
   url: string
   version: string | null
 
-  // Refactor away
+  /** @internal */
   scrollRegions: Array<{ top: number; left: number }>
+  /** @internal */
   rememberedState: Record<string, unknown>
 }
 
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -40,7 +40,6 @@ export interface Page<SharedProps extends PageProps = PageProps> {
   // Refactor away
   scrollRegions: Array<{ top: number; left: number }>
   rememberedState: Record<string, unknown>
-  resolvedErrors: Errors
 }
 
 export type PageResolver = (name: string) => Component
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -19,7 +19,7 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "preserveConstEnums": true,
-    "removeComments": true,
+    "removeComments": false,
     "typeRoots": ["./node_modules/@types"],
     "strict": true
   }
--- a/packages/vue3/src/app.ts
+++ b/packages/vue3/src/app.ts
@@ -1,5 +1,16 @@
 import { createHeadManager, Page, PageProps, router } from '@inertiajs/core'
-import { computed, DefineComponent, defineComponent, h, markRaw, Plugin, PropType, reactive, ref, shallowRef } from 'vue'
+import {
+  computed,
+  DefineComponent,
+  defineComponent,
+  h,
+  markRaw,
+  Plugin,
+  PropType,
+  reactive,
+  ref,
+  shallowRef,
+} from 'vue'
 import remember from './remember'
 import { VuePageHandlerArgs } from './types'
 import useForm from './useForm'
--- a/packages/vue3/src/app.ts
+++ b/packages/vue3/src/app.ts
@@ -1,5 +1,5 @@
 import { createHeadManager, Page, PageProps, router } from '@inertiajs/core'
-import { DefineComponent, defineComponent, h, markRaw, Plugin, PropType, reactive, ref, shallowRef } from 'vue'
+import { computed, DefineComponent, defineComponent, h, markRaw, Plugin, PropType, reactive, ref, shallowRef } from 'vue'
 import remember from './remember'
 import { VuePageHandlerArgs } from './types'
 import useForm from './useForm'
@@ -15,7 +15,7 @@ export interface InertiaAppProps {
 export type InertiaApp = DefineComponent<InertiaAppProps>
 
 const component = ref(null)
-const page = reactive({} as Page<any>)
+const page = ref<Page<any>>(null)
 const layout = shallowRef(null)
 const key = ref(null)
 let headManager = null
@@ -48,7 +48,7 @@ const App: InertiaApp = defineComponent({
   },
   setup({ initialPage, initialComponent, resolveComponent, titleCallback, onHeadUpdate }) {
     component.value = initialComponent ? markRaw(initialComponent) : null
-    Object.assign(page, initialPage)
+    page.value = initialPage
     key.value = null
 
     const isServer = typeof window === 'undefined'
@@ -60,7 +60,7 @@ const App: InertiaApp = defineComponent({
         resolveComponent,
         swapComponent: async (args: VuePageHandlerArgs) => {
           component.value = markRaw(args.component)
-          Object.assign(page, args.page)
+          page.value = args.page
           key.value = args.preserveState ? key.value : Date.now()
         },
       })
@@ -73,7 +73,7 @@ const App: InertiaApp = defineComponent({
         component.value.inheritAttrs = !!component.value.inheritAttrs
 
         const child = h(component.value, {
-          ...page.props,
+          ...page.value.props,
           key: key.value,
         })
 
@@ -92,7 +92,7 @@ const App: InertiaApp = defineComponent({
             .reverse()
             .reduce((child, layout) => {
               layout.inheritAttrs = !!layout.inheritAttrs
-              return h(layout, { ...page.props }, () => child)
+              return h(layout, { ...page.value.props }, () => child)
             })
         }
 
@@ -108,7 +108,7 @@ export const plugin: Plugin = {
     router.form = useForm
 
     Object.defineProperty(app.config.globalProperties, '$inertia', { get: () => router })
-    Object.defineProperty(app.config.globalProperties, '$page', { get: () => page })
+    Object.defineProperty(app.config.globalProperties, '$page', { get: () => page.value })
     Object.defineProperty(app.config.globalProperties, '$headManager', { get: () => headManager })
 
     app.mixin(remember)
@@ -116,5 +116,13 @@ export const plugin: Plugin = {
 }
 
 export function usePage<SharedProps extends PageProps>(): Page<SharedProps> {
-  return page
+  return reactive({
+    props: computed(() => page.value.props),
+    url: computed(() => page.value.url),
+    component: computed(() => page.value.component),
+    version: computed(() => page.value.version),
+    scrollRegions: computed(() => page.value.scrollRegions),
+    rememberedState: computed(() => page.value.rememberedState),
+    resolvedErrors: computed(() => page.value.resolvedErrors),
+  })
 }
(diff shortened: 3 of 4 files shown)


Tofandel:
You can actually revert the previous change and wrap the return of usePage in `reactive` to achieve the same behavior

It's a neat trick because if you use reactive in vue it will automatically unref values for you

mathieutu:
Indeed, I didn't think about it, it's a very cool trick to convert a ref object to a reactive (maybe the one I've been looking for for days?!)
.

BTW. I am looking for a way to have direct access to the props as objects of refs (in pages children components).

```ts
const { user, posts } = usePageProps()
```

Any idea how we could achieve that? (I initially pushed a commit that allowed to do that by `Object.assign(page.props, args.props)` but it doesn't remove the props that is not needed anymore, so it doesn't work).

The issue is that we 
I tried many different ways, but I haven't done any vuejs for a couple of years and my mental model is polluted by React's. 😅

Tofandel:
Oh, will you look at that, another useless force push.. And no reply on this but no problem merging his own PRs to make a release... 
For me it's the last drop, I'm jumping ship on this and starting a fork as soon as I got some time

reinink:
Hey folks, sorry for the force push on master — bad habit of mine while working on my personal projects, always forget it messes up PRs. I've rebased this PR against master to fix this.

@Tofandel I get that your frustrated, but keep in mind that Inertia is a project of mine that I created for myself initially, and only opened sourced for the benefit of others. Sometimes my responses will be fast, but often they will be slow. That's how open source works. Please don't be rude — I don't deserve it.

@mathieutu Thanks so much for taking a crack at fixing this. Really wish I had caught this issue before releasing 1.0, but that's how it goes sometimes. Going to do some testing on this. Is your latest iteration still a breaking change?

reinink:
@mathieutu @Tofandel Hey thanks so much for your help here — I've just tested this out and it works perfect, plus it's a non-breaking change, so that's a huge win 🥳

FYI, I made a few small tweaks around the internal `page` object properties.

I'm going to get this merged in and released right now.

reinink:
Released: https://github.com/inertiajs/inertia/releases/tag/v1.0.4

Tofandel:
> I get that your frustrated, but keep in mind that Inertia is a project of mine that I created for myself initially, and only opened sourced for the benefit of others. Sometimes my responses will be fast, but often they will be slow. That's how open source works. Please don't be rude — I don't deserve it.

Sorry about that outburst, I already made several comments over the course of the year, I posted constructive feedback and proposed my help. None of which got any reply until now. And I'm not the only person sharing this experience. So yes, frustration got the best of me. 

Given that this is a well sponsored project and advertised by laravel I would expect more attention to the community and less unilateralism (peer reviewed releases and maintainers). For the past year most releases introduced more bugs than they solved and then it was radio silence on those issues but I still saw new features and versions released and never any comment on the outlined issues and PR's aimed at solving them.

As a best practice a master branch should be protected on an open source project so that nobody can force push to it not even you. Altering the history of the project is not a good look and messing up PR's is even worse for your community which I was a part of.

For me this has only proven I cannot rely on this project and I've already decided on switching away.

It was fun to play with it and it is a great idea, I got the chance to use it on one project and was aware it was still a work in progress and was hoping to see it grow out of it's flaws, but my feeling is that it went the other way and I just cannot justify using this project in a professional setting for all the given reasons. I bet and it didn't play out

I do wish you all the best with it and hope you'll consider the feedback for a bright future on your project, cheers🙏❤️

fragkp:
@reinink Shouldn't the project be supported by the laravel team since version 1.0 or did I misunderstand something? I remember Taylor had said something about that once. That would be very good for the project.

emargareten:
Seems like there was some breaking change with this PR, I just upgraded to v1.0.5 and I get `caught TypeError: Cannot read properties of null (reading 'props')`.
It failed at this code `const auth = computed(() => usePage()?.props?.auth)` inside a plugin

onlime:
> Seems like there was some breaking change with this PR, I just upgraded to v1.0.5 and I get `caught TypeError: Cannot read properties of null (reading 'props')`.
> It failed at this code `const auth = computed(() => usePage()?.props?.auth)` inside a plugin

Same here. It already broke in v1.0.4

reinink:
@emargareten @onlime Hmm, shoot — tested it last night in a TypeScript project, and didn't run into this issue — and I tested it using the same code that you're showing.

Are you able to provide me with a minimal reproduction of this? 🙏

reinink:
Note for myself later when I take a closer look at this — I wonder if this is the solution:

```diff
export function usePage<SharedProps extends PageProps>(): Page<SharedProps> {
   return reactive({
-    props: computed(() => page.value.props),
-    url: computed(() => page.value.url),
-    component: computed(() => page.value.component),
-    version: computed(() => page.value.version),
-    scrollRegions: computed(() => page.value.scrollRegions),
-    rememberedState: computed(() => page.value.rememberedState),
+    props: computed(() => page.value?.props)
+    url: computed(() => page.value?.url),
+    component: computed(() => page.value?.component),
+    version: computed(() => page.value.version),
+    scrollRegions: computed(() => page.value?.scrollRegions),
+    rememberedState: computed(() => page.value?.rememberedState),
   })
 }
```

onlime:
> Note for myself later when I take a closer look at this — I wonder if this is the solution:

pretty sure yes. But please investigate what has changed from v1.0.0 -> v1.0.1 as the problem (`usePage()` initially returning `null`) exists since then, see my updated comment above.

onlime:
oh, maybe https://github.com/inertiajs/inertia/compare/v1.0.0...v1.0.1

```diff
- const page = ref<Partial<Page>>({})
+ const page = ref<Page<any>>(null)
```

Why suddenly initializing it with `null` instead of `{}`? I'm not really a frontend guy and not into TS, so can't help you any further on this. Hope you'll figure it out.

onlime:
tested #1530 with a local build in my project and all works as expected again. This also fixes the problem since v1.0.1, so now I can use `usePage().props` again instead of `usePage()?.props`.
@reinink Thanks for your efforts!

DJafari:
@reinink In order for you to find the bug better, I will give an explanation that I just realized

if we use `usePage` composable in external npm package, output of usePage always is null! (why ? i don't know ) 

i don't have any problem in `1.0.2 version`

reinink:
@DJafari Hey thanks for that info — what 3rd party library are you using that uses `usePage()`?

DJafari:
@reinink you're welcome

for mine this is a private package for my custom components for my inertia projects

but you can test with `momentum-modal` and etc

reinink:
Pretty sure I know what’s going wrong here. Looking at the Momentum Modal library, it's using the `usePage()` hook before Inertia is even initialized — meaning the internal `page` ref is still `null`, and so we get this error. I think the solution I suggested earlier (https://github.com/inertiajs/inertia/pull/1469#issuecomment-1516611185) will fix this.

reinink:
Fixed in #1530, and released as [v1.0.6](https://github.com/inertiajs/inertia/releases/tag/v1.0.6).

DJafari:
@reinink That error was removed for me, but there is a problem

in 3rd library, `page.value` always is null and never filled!

i defined this composable in my package :

```javascript
export default function useBackendUser() {
  const page = usePage()
  return computed(() => page?.props?.auth?.user)
}
```

and in my inertia app :

```javascript
const user = useBackendUser()
console.log(user.value) // undefined
```

but when i use this in my inertia app :

```javascript
const page = usePage()
const user = computed(() => page?.props?.auth?.user)
console.log(user.value) // my user is here
```

Tofandel:
FYI, this PR was assuming `page` was always an object because it reverted the old behavior, if page is initialised as null and you use `usePage` before initialisation it will not be reactive, it needs to be an object to be reactive

```js
const page = usePage(); // page = null
return computed(() => page?.props?.auth?.user) // Page will never update it's null which doesn't hold a reference
```

```js
const page = usePage(); // page = {}
return computed(() => page?.props?.auth?.user) // Page will update because it holds a reference to a reactive proxy
```

reinink:
@DJafari @Tofandel Hey I'm still not totally convinced that our current implementation is wrong. The `usePage()` hook now *always* returns a reactive object. Yes, the values within it are nullable, but the returned `page` is always a reactive object and is never `null`.

The `page` object that's referenced in the `usePage` hook is a `ref` — meaning it can have an initial value of `null` and still be reactive — that's how refs work in Vue 3 (as far as I'm aware).

I've tested this new implementation in the Vue 3 adapter with the Momentum Modal library, and it works exactly as expected, so I'm wondering if there is something else going on here with your particular implementation @DJafari. Are you able to please provide a minimal reproduction of this problem as a Git repo for me to test with?

Tofandel:
I take that back, @reinink is right
https://github.com/inertiajs/inertia/blob/d8a181beb6512307cf5cc5bca4c6b708faa47868/packages/vue3/src/app.ts#L129-L137

usePage never returns `null`

PR #1491, opened 2023-03-14
fix: Stores transform method in useRef

aviemet:
I'm resubmitting these changes in a new PR since the original, which was submitted nearly a year ago, now has conflicts due to the typescript rewrite. This is a duplicate of  #1171 in response to #1131.

Please, please, please, let's get this fixed. `useForm` is just simply broken without the ability to reliably transform data before submitting.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -58,7 +58,7 @@ export default function useForm<TForm extends Record<string, unknown>>(
   const [progress, setProgress] = useState(null)
   const [wasSuccessful, setWasSuccessful] = useState(false)
   const [recentlySuccessful, setRecentlySuccessful] = useState(false)
-  let transform = (data) => data
+  const transformRef = useRef<(data: TForm) => Record<string, any>>(data => data)
 
   useEffect(() => {
     isMounted.current = true
@@ -157,9 +157,9 @@ export default function useForm<TForm extends Record<string, unknown>>(
       }
 
       if (method === 'delete') {
-        router.delete(url, { ..._options, data: transform(data) })
+        router.delete(url, { ..._options, data: transformRef.current(data) })
       } else {
-        router[method](url, transform(data), _options)
+        router[method](url, transformRef.current(data), _options)
       }
     },
     [data, setErrors],
@@ -184,7 +184,7 @@ export default function useForm<TForm extends Record<string, unknown>>(
     wasSuccessful,
     recentlySuccessful,
     transform(callback) {
-      transform = callback
+      transformRef.current = callback
     },
     setDefaults(fieldOrFields?: keyof TForm | Record<keyof TForm, string>, maybeValue?: string) {
       if (typeof fieldOrFields === 'undefined') {


rsdrsd:
Without this the react version is unusable. What is the best method to install this PR for the time being?

reinink:

aviemet:
Just noticed this was closed again without being addressed. This bug has existed for years, and it's really, really simple to fix. I can't believe that 3 PRs in, this still hasn't been merged. I understand that you have a lot of issues to go through, but it seems like this has been closed without even being read, just like the previous one. There are 4 lines of code changed in this PR, they fix a long-standing bug which renders the documentation _wrong_ and fix a bug that's difficult to notice. Please, let's finally get this fixed.

aviemet:
@reinink People are bumping the discussion in the issue #1631, so I figured I'd try bumping the PR. 
This is still a bug and still an easy fix, I think we'd all appreciate this finally being addressed.

derrickreimer:
@aviemet trying to understand the unexpected behavior around `transform` here. To clarify, is the main issue here **attempting to call `transform` inside of a handler function (rather than in the body of the component definition)**? 

To give a concrete example:

```javascript
// A: This does work

const Page = () => {
  const { post, transform } = useForm({});

  transform((data) => ({ ...data, foo: "bar" }));
  
  const handleSubmit = (e) => {
    e.preventDefault();
    post("/route");
  }

  // ...
}

// B: This doesn't work

const Page = () => {
  const { post, transform } = useForm({});
  
  const handleSubmit = (e) => {
    e.preventDefault();
    transform((data) => ({ ...data, foo: "bar" }));

    // A re-render gets triggered somewhere in `post`, 
    // which wipes out the `transform` call just made
    post("/route");
  }

  // ...
}
```

I'm seeing scenario A work consistently, which squares with the implementation I'm seeing currently in Inertia. Since `transform()` is just a proxy for setting a local variable, as long as the `transform` call is placed in the function body, the internal state will get set to our custom definition on every render.

Does this all sound correct? Just hoping to crisply define the underlying issue so that we move the ball forward on either clarification in the docs or a change to the implementation.

aviemet:
Hello @derrickreimer, thanks for taking a look at this. To be honest, I wrote my own workaround which I've been using for a long time, so it's been a while since I've had to experience the issue first hand. I remember discovering it when I first started using Inertia, I would call `transform` when submitting a form and nothing would happen. Seeing your examples, I think I now understand how the method was intended to be used, and how that differs from how I and others have tried to use it. I'm pretty sure I only ever called it from a callback as in example B, I've never called it in the body of a component.

Part of the reason for doing so is that I also don't use the hook directly in pages. I created a Form component which takes care of a bunch of boilerplate and exposes the return values of `useForm` through callbacks and React context. I also created input components which interact with the form data through that context, so it's all abstracted away. 

I do think it's confusing to have a method which is only relevant when submitting a form, which can't be called from an `onSubmit` handler. It's also the only function returned by the hook which needs to be called in the body of the same component where the hook is called. If docs are updated to state that it should only be called directly after the hook, I wonder if there might also be a way to throw an error or a warning if it's not used as intended. Currently it just fails silently which makes it very difficult to track down the problem.

The only downside I can think of to memoizing the method is that someone may call it in multiple places, and then it's anyone's guess as to which transform gets called. I think it's still more intuitive than having it just get completely re-written if not called in the right place, though ultimately it's a sign that there's a better implementation possible.

I think the better choice would be to accept `transform` as an option to the hook, that way it isn't part of the return value and there's no confusion about where it should be called. Moving in this direction could also address the inconsistent interface for the hook which optionally accepts a string as the first argument. Accepting an object as the only parameter would also make it easy to accept more options parameters in the future without causing breaking changes, though it itself would be a breaking change.

```typescript
useForm<TForm extends Record<string, unknown>>({
  rememberKey?: string,
  initialData?: TForm,
  transform?: (cb: (data: TForm) => Record<string, unknown>) => void)
})
```

I hope that was helpful, thanks again for giving this some attention.

PR #1504, opened 2023-03-28
Fix form helper types for `setDefaults()` method (React) and `defaults()` method (Vue)

### LINKED ISSUE #1502: Form Defaults: Typescript forced to use strings.
### Versions:

- `@inertiajs/core` version: 1.0.2

### Describe the problem:

Currently, the type hinting when setting form defaults is requiring a string, whereas I would like my default to be of type `number` and possibly my own custom object type for a select list. I don't understand why this is forced to a string, not all defaults have to be strings.

```js
// @inertiajs/vue3/types/useForm.d.ts
...
    defaults(field: keyof TForm, value: string): this;
    defaults(fields: Record<keyof TForm, string>): this;
...
```

### Steps to reproduce:

Here's some sample code that shows the issue:

```js
interface User {
  name: string
  age: number
}

const form: InertiaForm<User> = useForm({})

onMounted(() => {
  getUser().then((data: User) => {
    form.defaults({ ...data }) // *TS-Error
    form.reset()
  })
})
```

> *TS-Error: Argument of type '{ name: string; age: number; }' is not assignable to parameter of type 'Record<keyof User, string>'.
  Types of property 'age' are incompatible.ts(2345)

shealavington:
Allows types than strings, based on the type passed in by the implementer.
fixes #1502 

Successfully tested locally in my project.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -19,8 +19,8 @@ export interface InertiaFormProps<TForm extends Record<string, unknown>> {
   setData: setDataByObject<TForm> & setDataByMethod<TForm> & setDataByKeyValuePair<TForm>
   transform: (callback: (data: TForm) => TForm) => void
   setDefaults(): void
-  setDefaults(field: keyof TForm, value: string): void
-  setDefaults(fields: Record<keyof TForm, string>): void
+  setDefaults(field: keyof TForm, value: typeof TForm): void
+  setDefaults(fields: TForm): void
   reset: (...fields: (keyof TForm)[]) => void
   clearErrors: (...fields: (keyof TForm)[]) => void
   setError(field: keyof TForm, value: string): void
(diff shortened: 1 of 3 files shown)


weavdale:
I'm running into this issue also. I have typed objects in my useForm and get a typescript error when trying to set defaults because it's expecting a string.

```js
const editForm = useForm({
  workOrder: props.workOrder,
  jobCosts: props.jobCosts,
  deletedJobCosts: new Array<number>(),
});

// this causes a type error because it's expecting strings
editForm.defaults({ workOrder: props.workOrder, jobCosts: props.jobCosts, deletedJobCosts: [] });
```

Another problem is the typing with the `errors` property. It doesn't allow for validation errors returned from Laravel for array values being validated. When I submit an array value to Laravel and the validation returns an error for an array field, I can't access the error for the array value in my template using `form.errors['jobCosts.1.amount']` or `form.errors.jobCosts.1.amount` without getting a type error.

reinink:
@shealavington Hey thanks so much for submitting this fix! I've made some further improvements to these types by taking advantage of the `FormDataConvertible` type that exists in Inertia core.

reinink:
These changes have been published in `v1.0.11` 👍

PR #1507, opened 2023-04-07
Update link to use props to retrieve event callbacks not attrs

likeadeckofcards:
When using Vue Attrs over the component props we lose the capability to define the prop as camelcase or kebab-case. In most cases, this is fine. If you are using Eslint, one of the rules from the vuejs core team that is recommended/strongly-recommended is using [kebab-case for component props in the template](https://eslint.vuejs.org/rules/attribute-hyphenation.html). By using the attrs to retrieve this data the end user cannot conform to eslint rules defined in their project. The following PR updates this functionality to use the props to retrieve the callbacks. The defaults for the props have been updated to match the intended definitions specified in the router class. This update removes the need to use the `// @ts-expect-error` comments in the render function.

### DIFF
--- a/packages/vue3/src/link.ts
+++ b/packages/vue3/src/link.ts
@@ -14,6 +14,7 @@ export interface InertiaLinkProps {
   only?: string[]
   onCancelToken?: (cancelToken: import('axios').CancelTokenSource) => void
   onBefore?: () => void
+  onError?: () => void
   onStart?: () => void
   onProgress?: (progress: Progress) => void
   onFinish?: () => void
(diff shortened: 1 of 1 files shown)


likeadeckofcards:
@reinink or @claudiodekker would love to get your review of this update.

likeadeckofcards:
@reinink can you please take a look at this PR?

ricardogobbosouza:
I just had this problem
PR more than necessary

pedroborges:
@likeadeckofcards Could you provide a reproducible repository or example that I can use to test the issue? It would really help in identifying the problem. Appreciate it!

joetannenbaum:
Closing this for now as we're not adding any new features or non-emergency fixes for 1.x. If This is still an issue in 2.x please submit a new PR.

PR #1509, opened 2023-04-09
Change Vue3 Head fragment detection method

craigrileyuk:
In development builds, a `v-for` array is received by the `renderNodes` function as type "Symbol(Fragment)" which is correctly detected. However, in production builds, this same array is received as "Symbol()". This results in the `v-for` array of Head elements being discarded.

### DIFF
--- a/packages/vue3/src/head.ts
+++ b/packages/vue3/src/head.ts
@@ -87,7 +87,7 @@ const Head: InertiaHead = defineComponent({
     renderNodes(nodes) {
       return this.addTitleElement(
         nodes
-          .flatMap((node) => (node.type.toString() === 'Symbol(Fragment)' ? node.children : node))
+          .flatMap((node) => (node.type.toString().startsWith('Symbol') ? node.children : node))
           .map((node) => this.renderTag(node))
           .filter((node) => node),
       )


reinink:
@craigrileyuk Thanks for catching and fixing this! 🙏

reinink:
Released: https://github.com/inertiajs/inertia/releases/tag/v1.0.7

SergkeiM:
@reinink @craigrileyuk after update I get this error: 

```js
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'toString')
    at Proxy.renderTag (head.ts:65:21)
```

SergkeiM:
After digging into the issue, it works fine prior vue <= v3.2.47 and breaks in v3.3.0

craigrileyuk:
Yeah. Vue seems to have made a major change to its vNode type property. 

Currently the Head component can’t handle comment tags or other Vue components (I.e if you tried to use <FacebookMeta /> inside Head. 

If no one gets to it before the weekend, I’ll give it a proper refactor. 

Until then, a workaround is to avoid anything that can cause a comment (e.g v-if)

SergkeiM:
Hi @craigrileyuk thank you,

I think simple filter will work here:

```js
.filter((node) => node && !node.text)
```

SergkeiM:
Hi @craigrileyuk I think this is related: https://github.com/vuejs/core/issues/4733, will check it more.

cjholowatyj:
I'm running `Vue@3.3.4` & `InertiaJS@1.0.7` and I can confirm that this issue is still not fixed.
Screenshot 2023-05-23 at 10 47 46 PM
Screenshot 2023-05-23 at 10 55 11 PM

SergkeiM:
I guess this #1570 PR by @craigrileyuk fixes this issues

olieady:
Having the same issue when trying to get my Breadcrumbs schema rendering in the head. 

<template>
    <Breadcrumbs
        :items="breadcrumbs"
        class="my-1 px-2"
        link-mode="inertia"
        v-if="breadcrumbs.length > 1"
    />
    <Head>
        {{ breadcrumbsSchema }}
    </Head>
</template>
```
<script setup>
import { Breadcrumbs } from "../";

import { usePage, Head } from "@inertiajs/vue3";
import { computed } from "vue";

const breadcrumbs = computed(() => usePage().props.breadcrumbs);

const breadcrumbsSchema = computed(() => {
    return usePage().props.breadcrumbsSchema;
});
</script>
```

reinink:
This fixed has been released as part of [v1.0.8](https://github.com/inertiajs/inertia/releases/tag/v1.0.8) 👍

olieady:
Hey, this doesn't fix the issue when trying to add schema to <head>. 

```vue
<template>
    <Head>
        {{ breadcrumbsSchema }}
    </Head>
</template>

<script setup>
import { usePage, Head } from '@inertiajs/vue3'
import { computed } from 'vue'

const breadcrumbsSchema = computed(() => usePage().props.breadcrumbsSchema)
</script>
```

```
<script type="application/ld+json">{"@context":"https:\/\/schema.org","@type":"BreadcrumbList","itemListElement":[{"@type":"ListItem","position":1,"item":{"@id":"https:\/\/my-site.com\/","name":"Home","image":null}}]}</script>
```

This works up with the most recent inertia version. However, Vue needs to be pinned below 3.3. 

Perhaps there's a better way to add JSON Schema to the head?

craigrileyuk:
> Perhaps there's a better way to add JSON Schema to the head?

Vue doesn't like the use of `<script>` tags or `<style>` tags inside the template so this probably shouldn't work.

A far better option is to [VueUse's useScriptTag](https://vueuse.org/core/useScriptTag/#usescripttag) and use it inside your layout component.

olieady:
Hey, I did think of this. However this wouldn’t be server side rendered. And since the purpose of using JSON schema here is to help search engines by supplying meta data about breadcrumbs, I feel it should probably be server side rendered.

iambryansanders:
Did either of you figure this out? @olieady @craigrileyuk @SergkeiM 

Need to have ld+json rendered via SSR.

olieady:
This issue was fixed in a later version. Here's what my Breadcrumbs component looks like: 

```vue
<template>
    <div>
        <Breadcrumbs
            :items="breadcrumbs"
            class="my-1 px-2"
            link-mode="inertia"
            v-if="breadcrumbs.length > 1"
        />
        <Head>
            {{ breadcrumbsSchema }}
        </Head>
    </div>
</template>

<script setup>
import { Breadcrumbs } from '@babestationui/vue'

import { usePage, Head } from '@inertiajs/vue3'
import { computed } from 'vue'

const breadcrumbs = computed(() => usePage().props.breadcrumbs)

const breadcrumbsSchema = computed(() => {
    return usePage().props.breadcrumbsSchema
})
</script>
```

PR #1516, opened 2023-04-13
Add initialization callback to form helper in Vue adapters

### LINKED ISSUE #1485: NProgress CSS included in the production JS bundle with includeCSS option set to false
### Version:

- `@inertiajs/vue3` version: 1.0.2

### Describe the problem:

I created Inertia app with the following configuration:

```javascript
createInertiaApp({
  ...
  progress: {
    includeCSS: false,
    showSpinner: false,
  },
});
```

When I run `vite build` I can see the NProgress CSS being added to the final JavaScript bundle:

Screenshot 2023-03-09 at 12 52 13 PM

reinink:
Right now there is no easy way to reinitialize a form using the form helper without providing all the data a second time. This creates unnecessary boilerplate in situations where a simple `form.reset()` isn't enough because you want to update the values based on props that may have changed. For example:

**Before (option 1):**

```js
// MyDialog.vue

const props = defineProps({
  company_id: Number,
})

const open = ref(false)
const form = useForm({
  first_name: '',
  last_name: '',
  email: '',
  role: 'User',
  company_id: props.company_id,
})

function open() {
  form.first_name = ''
  form.last_name = ''
  form.email = ''
  form.role = 'User'
  form.company_id = props.company_id
  open.value = true
}
```

**Before (option 2):**

```js
// MyDialog.vue

const props = defineProps({
  company_id: Number,
})

const open = ref(false)
const form = useForm(getFormData())

function getFormData() {
   return {
    first_name: '',
    last_name: '',
    email: '',
    role: 'User',
    company_id: props.company_id,
  }
}

function open() {
  Object.assign(form, getFormData())
  open.value = true
}
```

This PR solves this problem by adding an new optional initialization callback to the form helper. This function is automatically called when resetting the form, ensuring that the latest component state (props) is used.

```js
// MyDialog.vue

const props = defineProps({
  company_id: Number,
})

const open = ref(false)

const form = useForm(() => ({
  first_name: '',
  last_name: '',
  email: '',
  role: 'User',
  company_id: props.company_id,
}))

function open() {
  form.reset()
  open.value = true
}
```

### DIFF
--- a/packages/vue2/src/useForm.ts
+++ b/packages/vue2/src/useForm.ts
@@ -40,7 +40,7 @@ export default function useForm<TForm>(data: TForm | (() => TForm)): InertiaForm
 export default function useForm<TForm>(rememberKey: string, data: TForm | (() => TForm)): InertiaForm<TForm>
 export default function useForm<TForm>(...args): InertiaForm<TForm> {
   const rememberKey = typeof args[0] === 'string' ? args[0] : null
-  const data = (typeof args[0] !== 'string' ? args[0] : args[1]) || {}
+  const data = (typeof args[0] === 'string' ? args[1] : args[0]) || {}
   const restored = rememberKey ? (router.restore(rememberKey) as { data: any; errors: any }) : null
   let defaults = typeof data === 'object' ? cloneDeep(data) : cloneDeep(data())
   let cancelToken = null
--- a/packages/vue2/src/useForm.ts
+++ b/packages/vue2/src/useForm.ts
@@ -61,6 +61,8 @@ export default function useForm<TForm>(...args): InertiaForm<TForm> {
       Object.keys(this.data()).forEach((key) => delete this[key])
       defaults = cloneDeep(data)
       Object.keys(data).forEach((key) => Vue.set(this, key, data[key]))
+
+      return this
     },
     data() {
       return Object.keys(defaults).reduce((carry, key) => {
--- a/packages/vue2/src/useForm.ts
+++ b/packages/vue2/src/useForm.ts
@@ -11,6 +11,7 @@ interface InertiaFormProps<TForm> {
   progress: Progress | null
   wasSuccessful: boolean
   recentlySuccessful: boolean
+  init(data: TForm): this
   data(): TForm
   transform(callback: (data: TForm) => object): this
   defaults(): this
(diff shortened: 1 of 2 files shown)


packages/vue2/src/useForm.ts
```diff
@@ -36,19 +36,19 @@ export interface InertiaFormTrait {
...
-export default function useForm<TForm>(data: TForm): InertiaForm<TForm>
-export default function useForm<TForm>(rememberKey: string, data: TForm): InertiaForm<TForm>
+export default function useForm<TForm>(data: TForm | (() => TForm)): InertiaForm<TForm>
+export default function useForm<TForm>(rememberKey: string, data: TForm | (() => TForm)): InertiaForm<TForm>
 export default function useForm<TForm>(...args): InertiaForm<TForm> {
   const rememberKey = typeof args[0] === 'string' ? args[0] : null
   const data = (typeof args[0] === 'string' ? args[1] : args[0]) || {}
   const restored = rememberKey ? (router.restore(rememberKey) as { data: any; errors: any }) : null
-  let defaults = cloneDeep(data)
+  let defaults = typeof data === 'object' ? cloneDeep(data) : cloneDeep(data())
```
Tofandel, line 45: `useForm()` without any argument now throws an error, but it was a valid call before (in fact it's used widely in laravel)


It should be like this
```js
typeof data === 'function' ? cloneDeep(data()) :  cloneDeep(data)`
```

packages/vue3/src/useForm.ts
```diff
@@ -31,27 +31,27 @@ interface InertiaFormProps<TForm extends Record<string, unknown>> {
...
     ? (router.restore(rememberKey) as { data: TForm; errors: Record<keyof TForm, string> })
     : null
-  let defaults = cloneDeep(data)
+  let defaults = typeof data === 'object' ? cloneDeep(data) : cloneDeep(data())
   let cancelToken = null
   let recentlySuccessfulTimeoutId = null
   let transform = (data) => data
 
   let form = reactive({
-    ...(restored ? restored.data : data),
```
ragulka, line 54: Note that this is a breaking change: previously, it was possible to pass in a reactive object like this:

```js
const product = reactive({ name: '' });

const form = useForm({ product });
```

This way, `product.name` and `form.product.name` would always be in sync.

However, this no longer works, as `data` is now deep-cloned twice before spreading into the form object.

packages/vue3/src/useForm.ts
```diff
@@ -31,27 +31,27 @@ interface InertiaFormProps<TForm extends Record<string, unknown>> {
...
+    ...(restored ? restored.data : cloneDeep(defaults)),
```
ragulka, line 54: Note that this is a breaking change: previously, it was possible to pass in a reactive object like this:

```js
const product = reactive({ name: '' });

const form = useForm({ product });
```

This way, `product.name` and `form.product.name` would always be in sync.

However, this no longer works, as `data` is now deep-cloned twice before spreading into the form object.

thomasmokuaomweri2011: @reinink I recently ran into this breaking change while trying to keep a reactive object in sync with the form object. Has there been a proper graceful fix or we may have to find alternative approaches?

PR #1530, opened 2023-04-20
Fix `usePage()` null object error in Vue 3 adapter

onlime:
follow-up fix for #1469, as proposed by @reinink 

I have tested this with a local build and all is working as expected again.


reinink:
Hey, thanks for putting that together!

onlime:
@reinink Hey, thanks for the explanation and fast response!

PR #1531, opened 2023-04-21
Add beforeLeave hook on router

DJafari:
in some case we want doing a async work before inertia changed page

for example :
i have a form in modal, i want after form submited and result is success, before inertia navigate me to another page, hide my modal dialog

with this pr i can doing it by passing `onBeforeLeave` to `VisitOptions`

thanks

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -269,6 +269,7 @@ export class Router {
       onProgress = () => {},
       onFinish = () => {},
       onCancel = () => {},
+      onBeforeLeave = () => {},
       onSuccess = () => {},
       onError = () => {},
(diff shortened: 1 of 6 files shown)


jesusantguerrero:
Isn't onBefore hook covering that use case?

Adesin-fr:
Not sure if this is the same that onBefore, but I'd like to use async on the onBefore event, and it is currently not awaiting it !

joetannenbaum:
I believe this can be accomplished by hooking into the `onFinish` of the form, but regardless closing this as we're not adding any new features or non-emergency fixes for 1.x. If This is still an issue in 2.x please submit a new PR.

PR #1537, opened 2023-04-28
Improve type checking of request data

7nohe:
I implemented the feature of #1536.

```
import { router } from '@inertiajs/vue3'

const onClick = () => {
  const data = { message: 'Hi' }
  router.post<{ title: string }>(url, data, options) // <- NG
  router.post<{ message: string }>(url, data, options) // <- OK
}
```

In this PR, I have added improved type checking for request data.
The code snippet above demonstrates how the new feature works.
When posting data with an incorrect property, such as `{ title: 'hello' }`, it will result in a TypeScript error.

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -251,11 +251,11 @@ export class Router {
     }
   }
 
-  public visit(
+  public visit<TData extends RequestPayload = RequestPayload>(
     href: string | URL,
     {
       method = 'get',
-      data = {},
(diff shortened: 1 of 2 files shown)


joetannenbaum:
Closing this as we're not adding any new features or non-emergency fixes for 1.x. But I do like this a lot, so I ported it over for 2.x here with credit to you, thank you!

https://github.com/inertiajs/inertia/pull/2304

PR #1554, opened 2023-05-17
Fix: Add the transform to the callback's dependencies in submit funct…

shouwda:
The transform is not working before the submit.
So, put it into the submit dependencies for updates.


aviemet:
I just wanted to point to the other PR currently open relating to this bug (#1718)

pedroborges:
Thank you for your contribution, @shouwda! We've decided to proceed with #1607, which provides the same fix and some other improvements.

PR #1565, opened 2023-05-24
New feature: Frames

buhrmi:
My hope for this PR is to serve as a proof-of-concept to implement modals via a new feature I'd like to call Inertia Frames. An Inertia Frame is similar to standard browser `<iframe>`: By default, all user navigation and form submissions are encapsulated within a frame.

An example usage looks like this:
```html
<script>
import { Frame } from '@inertiajs/svelte`
</script>

<div class="modal">
  <Frame src="/session/new" id="login">
    Loading...
  </Frame>
</div>
```

When a `Frame` component is mounted into the page, it makes an Inertia Visit to the path specified in the `src` attribute. The Visit is made with a new attribute introduced in this PR: `router.visit(src, {frameId: "login"}`. When the router makes a visit with a specified frameId, it does not update the "global" page component, but the component within the Frame. The `id` attribute on the <Frame> is optional. If not specified, a random frameId is used.

I just started work on the implementation for the Svelte adapter and would like to keep working on this until an "official" or better implementation/concept for modals emerges. I don't really expect this PR to get merged any time soon, but I do hope to ignite some discussion about modals again.


jefhar:
I would be interested in seeing how this concept works inside the [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element.

buhrmi:
This is just an experimental proof of concept. Sorry I didn't have any expectations of it actually getting considered.

PR #1568, opened 2023-05-25
Clear errors on form reset

reinink:
Right now when using the form helper, calling `form.reset()` does not clear the form errors. This feels like a bug, since if you're resetting the form to its initial state you likely also want the errors to be cleared out.

This PR updates the form helper to clear form errors when resetting the form:

```js
// will reset all form errors
form.reset()
```

If you're only resetting certain fields, only those corresponding errors will be cleared:

```js
// will only reset "name" and "email" errors
form.reset('name', 'email')
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ For changes prior to v1.0.0, see the [legacy releases](https://legacy.inertiajs.
 
 ## [Unreleased](https://github.com/inertiajs/inertia/compare/v1.0.12...HEAD)
 
-- Nothing!
+- Clear errors on form reset ([#1568](https://github.com/inertiajs/inertia/pull/1568))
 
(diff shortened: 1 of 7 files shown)


weavdale:
The Svelte version of the useForm helper actually has a few bugs in it. I believe it boils down to trying to directly read values in a Svelte store using `this.`. 

Look on line 68 of the setError function. It is directly referencing this.errors variable to get a list of the current errors to add back into the store along with any new errors. This is not the correct way to get current values from the Svelte store. You need to subscribe to the store or use the store's update function which passes in the stores current values to the callback function. In this case when you call clearErrors and then immediately setError the internal variable this.errors still contains the old errors that were cleared in the clearErrors function.

weavdale:
@reinink I just PR'd a fix https://github.com/inertiajs/inertia/pull/1574

PR #1570, opened 2023-05-28
[vue3]: Backwards compatible vNode detection for head

craigrileyuk:
Improved node detection process for Head component which now supports fragments, conditionals and functional components. Comments nodes are now automatically discarded.

### DIFF
--- a/packages/vue3/src/head.ts
+++ b/packages/vue3/src/head.ts
@@ -87,10 +87,25 @@ const Head: InertiaHead = defineComponent({
     renderNodes(nodes) {
       return this.addTitleElement(
         nodes
-          .flatMap((node) => (node.type.toString().startsWith('Symbol') && node.children ? node.children : node))
+          .flatMap((node) => this.resolveNode(node))
           .map((node) => this.renderTag(node))
           .filter((node) => node),
       )
+    },
+    resolveNode(node) {
+      const nodeType = node.type && node.type.toString()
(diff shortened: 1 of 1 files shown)


reinink:
@craigrileyuk Hey I really appreciate you tackling this! This looks good, although a plain `<title>` tags no longer work:

```vue
<Head>
  <title>Home</title>
</Head>
```

Is there a reason you're excluding nodes with children, here?

```js
else if (!isComment && !node.children) return node
```

Could this just be this?

```js
else if (!isComment) return node
```

craigrileyuk:
> Could this just be this?
> `else if (!isComment) return node`

Yep. That fixes the `<title>` issue. I was trying to get the head node resolution to work with full SFCs and I think that one line was a holdover I forgot from that. Unfortunately, I couldn't get it to play nice with SSR, so I dropped it.

reinink:
Okay great, I'll get this merged in and released. I'll probably tweak the code formatting a little to match what we normally do in this project, but really appreciate you doing the hard work here! 🙏

> Unfortunately, I couldn't get it to play nice with SSR, so I dropped it.

Well, thanks for trying either way 😊

reinink:
@craigrileyuk I've tweaked the formatting here to try and simplify things — I *think* everything still functionally works the same, but feel free to review 😊

reinink:
This fixed has been released as part of [v1.0.8](https://github.com/inertiajs/inertia/releases/tag/v1.0.8) 👍

Issue #1521, opened 2023-04-19
useForm helper on 422 validation errors in the helper won't update


stephenkhoo:
### Version:

- `@inertiajs/svelte` version: 1.0.2, 1.0.3

### Describe the problem:

When a form is submitted, errors from backend passed to the helper error bag with setError
However, when we fill in a few fields, and resubmit, the error supposedly clear and set with new errors.
But it didn't updated.

### Steps to reproduce:

useForm to prepare data and submit to a validation protected request
Submit without data, and let the $form.errors shows up.
Add in a few data (supposed to remove some of the errors), and submit again
The errors remain.

### What we've tried

We tried to use options.onError
To set the error again

`$form.clearErrors().setError(errors)`

The problem remains

Then we attempt to break the chain

```
$form.clearErrors()
$form.setError(errors)
```

Then it works.
Not sure if it is cause by some bug within svelte when chain in this way.
However, when we do not chain it, it will work as expected

weavdale:
I ran into this problem also. I thinking about submitting a PR for the fix.

The problem is the useForm helper is improperly using the Svelte store. On line 75 the `clearErrors` function is calling the `setStore` helper function which then calls the `update` function of the Svelte store. This is the correct way to update a value in the Svelte store. The store will then notify all subscribers of the change.

The problem comes on line 68 of the `setError` function. It is directly referencing `this.errors` variable to get a list of the current errors to add back into the store along with any new errors. This is not the correct way to get current values from the Svelte store. You need to subscribe to the store or use the store's `update` function which passes in the stores current values to the callback function. In this case when you call `clearErrors` and then immediately `setError` the internal variable `this.errors` hasn't been updated yet and still contains the old errors that were cleared in the `clearErrors` function.

There is more than one way to fix this, but my recommendation is to eliminate the helper function `setStore` to update the store and to explicitly call the store's `update` function every time you want to update the store. This way you always have access to the store's current values.

The new `setError` and `clearErrors` would look like this. Note, all other functions using `setStore` would also need updated.

```
function setError(key, value) {
  store.update((currentState) => ({
    ...currentState,
    errors: Object.assign(
      {},
      currentState.errors,
      value ? { [key]: value } : key
    ),
  }));

  return this;
}

function clearErrors(...fields) {
  store.update((currentState) => ({
    ...currentState,
    errors: Object.keys(currentState.errors).reduce(
      (carry, field) => ({
        ...carry,
[...]

PR #1574, opened 2023-05-30
Update useForm to correctly update/read values in Svelte store

weavdale:
The Svelte version of the useForm helper is incorrectly reading values from the Svelte store by using `this`.

On line 68 of the `setError` function is directly referencing `this.errors` variable to get a list of the current errors to add back into the store along with any new errors. This is not the correct way to get current values from a Svelte store. You need to subscribe to the store or use the store's update function which passes in the stores current values to the callback. In this case when you call `clearErrors` and then immediately `setError` the internal variable `this.errors` still contains the old errors that were cleared in the `clearErrors` function. The store hasn't updated them yet.

With this PR I updated the `setStore` function to allow the `value` parameter to be a callback (in addition to string and object) that passes in the current store values. This allows the `setError` and `clearErrors` function to have access to the current store values by using the callback.

This should also fix https://github.com/inertiajs/inertia/issues/1521

### DIFF
--- a/packages/svelte/src/useForm.js
+++ b/packages/svelte/src/useForm.js
@@ -22,7 +22,8 @@ function useForm(...args) {
     processing: false,
     setStore(key, value) {
       store.update((store) => {
-        return Object.assign({}, store, typeof key === 'string' ? { [key]: value } : key)
+        const _value = typeof value === 'function' ? value(store) : value
+        return Object.assign({}, store, typeof key === 'string' ? { [key]: _value } : key)
       })
     },
     data() {
@@ -57,30 +58,29 @@ function useForm(...args) {
             .reduce((carry, key) => {
               carry[key] = defaults[key]
               return carry
-            }, {}),
+            }, {})
         )
       }
 
       return this
     },
     setError(key, value) {
-      this.setStore('errors', {
-        ...this.errors,
+      this.setStore('errors', (store) => ({
+        ...store.errors,
         ...(value ? { [key]: value } : key),
-      })
+      }))
 
       return this
     },
     clearErrors(...fields) {
-      this.setStore(
-        'errors',
-        Object.keys(this.errors).reduce(
+      this.setStore('errors', (store) =>
+        Object.keys(store.errors).reduce(
           (carry, field) => ({
             ...carry,
-            ...(fields.length > 0 && !fields.includes(field) ? { [field]: this.errors[field] } : {}),
+            ...(fields.length > 0 && !fields.includes(field) ? { [field]: store.errors[field] } : {}),
           }),
-          {},
-        ),
+          {}
+        )
       )
 
       return this


reinink:
@weavdale Awesome — thanks for this! I'm going to do some testing, and get #1568 updated after merging this in.

weavdale:
Awesome! Also Svelte 4 will be released in the next week or two. Not supposed to be much breaking changes but you can start testing with the beta. https://github.com/sveltejs/svelte/releases/tag/svelte%404.0.0-next.1

Rich just did a video recently with the changes coming. 
https://youtu.be/AOXq89h8saI?t=195

There will be a new website with a migration guide but Svelte 4 is mostly a house cleaning release to prepare for Svelte 5.
https://svelte-dev-2.vercel.app/

reinink:
@weavdale sweet! going to give it a try! 🤞

weavdale:
Glad you got Svelte 4 support pushed out. Great job!

reinink:
> With this PR I updated the `setStore` function to allow the `value` parameter to be a callback (in addition to string and object) that passes in the current store values. This allows the `setError` and `clearErrors` function to have access to the current store values by using the callback.

@weavdale can you explain how you would use this callback?

weavdale:
@reinink sure! On line 66 of the current `userForm.js` you are trying to access a Svelte store's value from inside the Svelte store by calling the variable directly `this.errors`. 

```js
setError(key, value) {
    this.setStore('errors', {
        ...this.errors,     // Can't access a reactive store value this way because it may not be updated yet
        ...(value ? { [key]: value } : key),
    })

    return this
},
```
You can't access reactive store values in this way because Svelte may have not updated the internal value yet. You can read a store value using the `subscribe` method, set the value using the `set` method, or update the value using the `update` method (https://svelte.dev/docs/svelte-store). `update` is a method that takes one argument which is a callback. The callback takes the existing store value as its argument and returns the new value to be set to the store.

So in the code above you are trying to spread the existing errors back into the `errors` variable using `...this.errors`. This will not always get the current value of `errors`. Especially if you update `errors` and then immediately try to access the value of `errors` like you are doing on line 137. You clear errors and then assign new ones using the `setError` function above. You would think that when `setErrors` function calls `...this.error` that it would be empty. But it's not. It still has all the old errors that you just tried to clear. That's because `this.error` wasn't updated yet by Svelte. So all the cleared errors get added back.
```js
onError: (errors) => {
    this.setStore('processing', false)
    this.setStore('progress', null)
    this.clearErrors().setError(errors) // This will add all cleared errors back into errors variable

    if (options.onError) {
        return options.onError(errors)
    }
},
```
So I proposed to modify the `setStore` function to also accept a callback that gets the store's current values passed into it.
```js
setStore(key, value) {
    store.update((store) => {   // current store values gets passed in by the update method
    // If "value" is a callback, pass the current store's value into it so it has access to existing values.
    // Then the callbacks return value is assigned to _value which is then used to update the store 
    // with the new value.
    const _value = typeof value === 'function' ? value(store) : value 
        return Object.assign({}, store, typeof key === 'string' ? { [key]: _value } : key)
    })
},
```
With this updated `setStore` function, we can modify the `setError` function to pass a callback to `setStore`. That way `setStore` will pass the current value of the store to `setError`'s callback so it can spread existing errors and add any new ones and then return the value back to `setStore` to update the store's value.
```js
setError(key, value) {
    this.setStore('errors', (store) => ({
        ...store.errors,
        ...(value ? { [key]: value } : key),
    })) 

    return this
```

weavdale:
Any updates on the progress of  fixing `useForm` for Svelte?

weavdale:
I'm am closing this PR because I like `setStore` implementation for `useForm` that PR https://github.com/inertiajs/inertia/pull/1614 and PR https://github.com/inertiajs/inertia/pull/1610 are proposing. Much cleaner and doesn't require the `setStore` function to receive a callback to update specific store values.

PR #1586, opened 2023-06-06
Preserve Default Meta in Inertia.js Core Package

emrancu:
I have made a simple update to the Inertia.js core package that addresses the issue of default meta deletion when the inertia attribute is added to a meta tag but not explicitly passed to the ```Head``` component.

In the current system, if the inertia attribute is present on a meta tag but not passed to the ```Head``` component, the default meta is deleted from the head section. This behavior has caused inconvenience for certain use cases.

To overcome this issue, I have removed the line ```targetElement?.parentNode?.removeChild(targetElement)``` from the core package's 'head.ts``` file . This change ensures that the default meta remains intact even when the meta not passed to the ```Head``` component.
 
I believe that this modification will greatly enhance the usability and flexibility of the Inertia.js core package. It resolves the problem of unintended meta deletion and ensures a more consistent and intuitive experience.

Thank you for considering this pull request. I am excited to contribute to the Inertia.js project and look forward to your feedback and suggestions.

### DIFF
--- a/packages/core/src/head.ts
+++ b/packages/core/src/head.ts
@@ -41,7 +41,6 @@ const Renderer = {
     targetElements.forEach((targetElement) => {
       const index = this.findMatchingElementIndex(targetElement as Element, sourceElements)
       if (index === -1) {
-        targetElement?.parentNode?.removeChild(targetElement)
         return
       }
 


emrancu:
### Senario:

Suppose: I added title & meta on app.blade.php file. 
```html
 <title inertia>Awesome Inertia</title>
 <meta  name="title" content="Love Inertia JS" inertia>
```

and I add Head component on Home.vue page 

```html
<Head>
  <meta  name="title" content="You have to love Inertia JS">
</Head>
```

Result will be on html preview:

```html
<head>
   <meta  name="title" content="You have to love Inertia JS" inertia>
</head>
```
In Current System title meta will be deleted.

But after this update result will be:

```html
<head>
   <title inertia>Awesome Inertia</title>
   <meta  name="title" content="You have to love Inertia JS" inertia>
</head>
```

reinink:
Hey thanks for this suggestion, but from my testing this doesn't actually seem to solve the `<title>` problem, plus it also introduces a bug where head elements that exist on one page and not on another are not properly removed when navigating between pages.

The whole purpose of the `inertia` attribute on the `<title>` tag is to tell Inertia.js to track it, and remove if it doesn't exist on subsequent pages.

My recommendation would be to just have a default `<title>` client-side, either in a layout file, or even via a `title` callback: https://inertiajs.com/title-and-meta#title-callback

Hope that helps 👍

emrancu:
Hello @reinink ,
It was my purpose to keep default title if I not pass on Head tag.

In certain cases, we may want to define a default title for our pages, such as using the Laravel framework's ```@yield('title', 'Site')``` syntax.
However, if a specific page title is set, we want it to override the default title. If no specific title is provided, we should retain the default title. This behavior ensures flexibility while maintaining a consistent title structure across the site.

PR #1590, opened 2023-06-14
Refinements to node detection

craigrileyuk:
Added support for JSON schema script strings that the internal engine changes introduced in Vue 3.3 broke.

Last one (hopefully).


reinink:
@craigrileyuk Thanks again! 🙏

reinink:
Released: https://github.com/inertiajs/inertia/releases/tag/v1.0.9

PR #1596, opened 2023-06-22
[1.x] Export `InertiaFormProps` in the React adapter

olragon:
This is helpful when defining a custom Form wrapper component that will provide custom methods. E.g. `getError(fieldPath: string)`, `getData(fieldPath: string)`.


kendepelchin:
I also need this, but fixed it using the following types:
```form: ReturnType<typeof useForm<{your own type of form}>>;````

rrmesquita (approved):

pedroborges (approved):
The other adapters already export this type, so I think we should do the same for React as well 👍

Thanks @olragon!

joetannenbaum (approved):
Looks good, thanks!

PR #1607, opened 2023-07-10
[1.x] Fix useForm re-renders by memoizing functions in React

Hasan-Mir:
When using the `useForm()` in a React component that has some heavy pure components(components using `React.memo`) as children and passing some methods returned by `useForm()` like `setData`, `clearErrors`, etc to them, on every re-render, these methods returned by `useForm()` are getting **new references** so new props are passed to these `React.memo()`ed children and they will be re-rendered unnecessarily. 
This pull request fixes this issue by using `useCallback` for those methods in the `useForm` hook.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -165,9 +165,8 @@ export default function useForm<TForm extends Record<string, unknown>>(
     [data, setErrors],
   )
 
-  return {
-    data,
-    setData(keyOrData: keyof TForm | Function | TForm, maybeValue?: TForm[keyof TForm]) {
+  const setDataFunction = useCallback(
+    (keyOrData: keyof TForm | Function | TForm, maybeValue?: TForm[keyof TForm]) => {
       if (typeof keyOrData === 'string') {
         setData({ ...data, [keyOrData]: maybeValue })
       } else if (typeof keyOrData === 'function') {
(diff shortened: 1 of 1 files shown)


jameshulse:
I just want to flag that this PR (mostly) fixed some issues that we were running into. I have ended up taking a copy of this `useForm` implementation and referencing it directly in our application. So just giving a +1 that this PR could be prioritised to bring in.

I also needed to add `data` as a dependency of the `transformFunction` callback to ensure that it worked correctly for how we are using the hook:

```
  const transformFunction = useCallback(
    (callback) => {
      transform = callback;
    },
    // Falcon: We need to refresh the callback if the data changes
    [data]
  );
```

Hasan-Mir:
Hi @jameshulse, thank you for the hint, I fixed it in the last commit by using the `useRef` instead of declaring the `transform` as `let transform = (data) => data`. This way the transform function always has the same reference between re-renders.

kendepelchin:
Anyone has an idea on what this PR is waiting? Or could I patch package this somehow?

derrickreimer (approved):
These changes seem good & safe. It requires a bit of diligence to keep the dependency arrays up-to-date on the `useCallback` calls, but I think worthwhile for performance.

pedroborges (approved):
👍

joetannenbaum (approved):
This looks good, thank you! As @derrickreimer mentioned, we just have to take care to maintain the dependency arrays.

Hasan-Mir:
@joetannenbaum
Adding Eslint with [eslint-plugin-react-hooks](https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks) simplifies maintaining the dependency arrays.

Issue #1276, opened 2022-09-02
Form Previous Error Not Getting Cleared


ramonmalcolm10:
### Versions:

- `@inertiajs/inertia` version: 0.11.0
- `@inertiajs/inertia-svelte` version: 0.8.0

### Describe the problem:

When submit a form with multiple input fields and those fields failed validation, if correct anyone of those field and resubmit, the previous error message will still displayed for the now valid field.

### Steps to reproduce:

- Create a multiple input forms using  `useForm({ first_name: '', last_name: ''  })`
- Add required server side validations for those fields
- Display error `{#if $form.errors.first_name}
                            <p>{$form.errors.first_name}</p>
                        {/if}
{#if $form.errors.last_name}
                            <p>{$form.errors.last_name}</p>
                        {/if}`
- Submit form without populating the fields `<form on:submit|preventDefault={$form.post('/test'))}>`
- After getting back the validations failure, input data in one of the fields and resubmit

ainesophaur:
FWIW for anyone who comes across this.. Ive noticed `$form.errors` will only retain the initial errors returned by the server and doesnt update. However, `$page.props.errors` does contain the updated validation results.

edgars-vasiljevs:
Looks like updating store from within a store doesn't update properties internally.

Here is a quick way to reproduce the issue:

    let form = useForm({ test: '' });
    $form.setError('test', 'some error');
    let context = $form.clearErrors(); //clearErrors() returns "this" object
    console.log(JSON.stringify(context.errors));
    console.log(JSON.stringify($form.errors));

Output:

    {"test":"some error"}
    {}

Internally `useForm` calls `.clearErrors` and then `.setError` to set new errors received from failed request ([src code](https://github.com/inertiajs/inertia/blob/master/packages/inertia-svelte/src/useForm.js#L139)). 
Since it doesn't clear errors internally, `useForm` just appends new errors to existing ones (basically merging them).

talovicnedim:
Thanks @ainesophaur 

@edgars-vasiljevs what is your approach? Using `$page.props.errors` or? Thanks!

I wonder whether should I go with Vue or Svelte for a new project.

ramonmalcolm10:
@talovicnedim I would recommend going with Vue, much better support is available.

talovicnedim:
That's so fast! @ramonmalcolm10 

I'm thinking that too. Svelte looks cool, but seems it's too early for that.

reinink:
[...]

PR #1610, opened 2023-07-13
fixes for svelte's useForm helper

edgars-vasiljevs:
This fixes two issues in useForm helper that I found.

- Form fields with objects as values (arrays/objects) has broken reactivity. This is because input `data` is shallow copied to `defaults`. Updating nested objects in form updates `defaults` object as well. Since the current form data is compared to `defaults`, no changes are detected.
- Broken `setStore`. Svelte's stores replaces store value with anything it receives from `update`'s callback function. In this case `setStore` creates completely new object instance every time it's called by returning `Object.assign({}, store, ...)`. Later when form submit fails, `this.clearErrors().setError(errors)` is called on object instance that request was made from which is NOT the object currently stored in form store hence nothing is cleared.

This should fix form error issue in #1521 #1276

Edit:

Added two more fixes:

- reset function also shallow copies but this time defaults to data. This again breaks nested array/object reactivity.
~~- updating defaults to current data after successful request. It's debatable but IMO form defaults should be updated because previous form state is irrelevant after successful submit.~~

I see that Vue adapters work very similarly to the changes I made.

### DIFF
--- a/packages/svelte/package.json
+++ b/packages/svelte/package.json
@@ -29,6 +29,7 @@
   },
   "dependencies": {
     "@inertiajs/core": "1.0.9",
+    "lodash.clonedeep": "^4.5.0",
     "lodash.isequal": "^4.5.0"
   }
 }
--- a/packages/svelte/src/useForm.js
+++ b/packages/svelte/src/useForm.js
@@ -7,7 +7,7 @@ function useForm(...args) {
   const rememberKey = typeof args[0] === 'string' ? args[0] : null
   const data = (typeof args[0] === 'string' ? args[1] : args[0]) || {}
   const restored = rememberKey ? router.restore(rememberKey) : null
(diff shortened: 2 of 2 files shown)


edgars-vasiljevs:
Added fixes for two more issues:

- `reset` function also shallow copies but this time `defaults` to `data`. This again breaks nested array/object reactivity.
- updating `defaults` to current data after successful request. It's debatable but IMO form defaults should be updated because previous form state is irrelevant after successful submit.

I see that Vue adapters work very similarly to the changes I made.

taylorotwell:
Hey there - this is on our radar. Hoping to get it merged next week. 👍

jessarcher (approved):
I've tested this locally with and without this branch and can confirm the issue and the fix.

Thank you!

PR #1617, opened 2023-07-23
Add ability to handle deferred props

adrum:
This works in tandem with https://github.com/inertiajs/inertia-laravel/pull/531. This is the client-side implementation of this feature. These deferred properties will only ever be loaded automatically on page loads/navigations. Otherwise, they behave just the same as Lazy props.

See the previously mentioned PR for more details.


NickSdot:
@reinink Cheers to you for giving such thoughtful, reasoned answers to people. ❤️

PR #1626, opened 2023-07-28
Incorrect types: Form should extend `RequestPayload`

Bloemendaal:
Resubmission of https://github.com/inertiajs/inertia/pull/1479 since this is still a problem for us.

I'm getting the following errors:

```
TS2344: Type 'MyFormData' does not satisfy the constraint 'Record<string, unknown>'.
Index signature for type 'string' is missing in type 'MyFormData'.
```

This PR uses the `RequestPayload` type which is also used by the router to validate the data.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -1,4 +1,4 @@
-import { Method, Progress, router, VisitOptions } from '@inertiajs/core'
+import { Method, Progress, RequestPayload, router, VisitOptions } from '@inertiajs/core'
 import isEqual from 'lodash.isequal'
 import { useCallback, useEffect, useRef, useState } from 'react'
 import useRemember from './useRemember'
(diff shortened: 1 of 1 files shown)


jessarcher:
Hey @Bloemendaal,

Where are you seeing this error? I can't figure out how to replicate it to confirm the issue and solution.

Can you please provide a minimal reproduction?

Bloemendaal:
Hi! Thanks voor looking into this. Here's an snippet from our codebase:

```tsx
interface FormData {
    fullName: string;
    phone: string;
    email: string;
}

export interface BasicInfoProps extends SharedProps {
    fullName: '';
    phone: '';
    email: '';
}

const BasicInfo = ({
    nextRouteUri,
    errors,
    fullName,
    phone,
    email,
}: BasicInfoProps) => {
    const { data, setData, post, processing } = useForm<FormData>('basicInfo', {
        fullName: fullName || '',
        phone: phone || '',
        email: email || '',
    });

   // ...

    return <div>Hello {data.fullName}!</div>
};
```

This gives the following error:
Scherm­afbeelding 2023-08-09 om 08 39 03

reinink:
Hey @Bloemendaal thanks so much for submitting this fix, but I think we're going to go with the solution in #1649 instead — which should also solve this issue for you 👍

Issue #78, opened 2019-09-26
Add the ability to mutate pageProps


sebastiandedeyne:
Sometimes I wish there was a way to mutate the page props. 

For example: you have a list of posts that have a publish button that uses Axios. After clicking publish, the post's status needs to be updated.

The current way of doing thigs:

1. Inertia sends a `posts` object in the response
2. You need to copy the `posts` object to local component state, which may be mutated
3. Whenever an `Inertia.visit` happens with `preserveState`, you need to manually override your local component state with the fresh `posts` data

This is really annoying to maintain, and would be simpler if we could just mutate the prop in Inertia.

This is a potentially dangerous feature, so I'm not sure about it yet. Would like some external input.

reinink:
So, in your example, why wouldn't you just do an `Inertia.put()` to update the post status, and then redirect back to the post index? What's the benefit of doing this an an Axios request?

sebastiandedeyne:
This probably doesn't matter in the posts example, however, if you're dealing with data that changes often, there's a chance that your list has completely changed by the time you do the `put` request.

If you're on page 2 of a list, modify an item, and receive a new set of data for page 2, there's a chance that the item you modified isn't part of the list anymore, which would be super confusing for the user.

We do this in Flare, since errors can come in pretty fast is an application is on fire. We use Axios call to resolving or snoozing errors to ensure the context doesn't completely change after a request.

reinink:
Ahh yes, real interesting. That makes sense.

Something that needs to be considered here is that if we change the props somehow, we'll also need update the cached history state data as well.

I also wonder if, with a feature like this, you quickly get into modifying the URLs and history state as well. I feel like this is a dangerous road to go down, but will be very tempting.

reinink:
So, I'm seeing some interesting use cases for a feature like this, where you can modify the root level props (page object), which then reactively updates the page component props.

One really neat use case for this is infinite scrolling. Imagine you have a tweets page, that shows the 15 most recent tweets. Then, when you scroll and load additional tweets, you could use something like `Inertia.updateProps()` to push these additional tweets into the `page` object. Something like this:

```js
loadMoreTweets() {
  fetch(`/tweets?page=${nextPage}`)
    .then(response => response.json())
    .then(tweets => Inertia.updateProps(props => {
      props.tweets.concat(tweets)
      return props
    }))
}
```

A different approach would be to allow per visit prop transformations:

```js
loadMoreTweets() {
  let currentTweets = this.tweets
[...]

PR #1633, opened 2023-08-01
Add ability to mutate page props

vinnygambiny:
This PR allows mutating the page props with the usePage function.

This is for these situations https://github.com/inertiajs/inertia/discussions/78

You have only to do `usePage().props.valueName = 'newValue'`

### DIFF
--- a/packages/vue3/src/app.ts
+++ b/packages/vue3/src/app.ts
@@ -128,7 +128,12 @@ export const plugin: Plugin = {
 
 export function usePage<SharedProps extends PageProps>(): Page<SharedProps> {
   return reactive({
-    props: computed(() => page.value?.props),
+    props: computed({
+      get: () => page.value?.props,
+      set: (value) => {
+        page.value.props = value
+      }
+    }),
     url: computed(() => page.value?.url),
     component: computed(() => page.value?.component),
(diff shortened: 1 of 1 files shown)


punyflash:
I don't think that should be a thing. Data which loaded the page should be accessible all the time, no matter what. In case you need it to be reactive and changeable - just create a custom reference that was initialized with the prop value.

innocenzi:
@punyflash there are definitely situations where this would be handy (eg. infinite loading), and Inertia should be more flexible about this.

pedroborges:
I'm closing this PR since the main use case discussed here and on #78 is supporting for infinite scrolling. That's already coming to v2.0 soon, as you can see in [Taylor's Laracon keynote](https://www.youtube.com/watch?v=AwWepVU5uWM&t=2845s) 🚀

Fuzzyma:
@pedroborges the usecase for updating data that comes with laravel echo events is still totally valid.
Not being able to change the page props on an event degrades the usafe of events do "hey, i got something new. Please reload".
It is really not a great ux. I feel like since events are a laravel core feature it would be nice to have a better integration path into inertia here.

There are not really any alternatives to solve this in a nice way and they all come with overhead.

Issue #1627, opened 2023-07-28
Svelte 4 warning during build


punyflash:
### Version:

- `@inertiajs/svelte` version: 1.0.9
- `svelte` version: 4

### Describe the problem:

```sh
[vite-plugin-svelte] /node_modules/.pnpm/@inertiajs+svelte@1.0.9_svelte@4.1.1/node_modules/@inertiajs/svelte/src/Link.svelte:25:0 A11y: <svelte:element> with click, dblclick, mousedown, mousemove, mouseout, mouseover, mouseup handlers must have an ARIA role
23: </script>
24: 
25: <svelte:element
    ^
26:   this={as}
27:   use:inertia={{
```
### Steps to reproduce:

```sh
vite build
```

reinink:
Hey can you provide a minimal reproduction of this as a git repo? This makes it much easier for us to reproduce things on our end and ultimately fix it if there is an issue.

Also, for what it's worth our Svelte demo app uses version 4, and I don't see this issue there. You can find that demo app here:

https://github.com/inertiajs/inertia/tree/master/playgrounds/svelte

punyflash:
> Hey can you provide a minimal reproduction of this as a git repo? This makes it much easier for us to reproduce things on our end and ultimately fix it if there is an issue.
> 
> Also, for what it's worth our Svelte demo app uses version 4, and I don't see this issue there. You can find that demo app here:
> 
> https://github.com/inertiajs/inertia/tree/master/playgrounds/svelte


This issue is quite small, I don't feel a need to create whole repository to show it. It occurred in case when `Link.svelte` is used. This is just Svelte's A11y warning

jdion84:
@punyflash  did you ever figure this out?

punyflash:
> @punyflash did you ever figure this out?
Fix in #1635. Also, I'd rather not recommend you to use inertia with Svelte 4, as #1621 is even worse

ingLomeland:
anyone figured out this problem? I've got a fresh project and not even using the Link component, but I still get this warning

driesvints:
A PR was sent in but was stalled. It's best that someone picks up the work in a new PR. Thanks

reinink:
Decided to fix this using a `svelte-ignore` (see #1858), and that has been released as part of [v1.0.16](https://github.com/inertiajs/inertia/releases/tag/v1.0.16) 👍

PR #1635, opened 2023-08-02
Fix svelte 4 warning

punyflash:
#1627

### DIFF
--- a/packages/svelte/src/Link.svelte
+++ b/packages/svelte/src/Link.svelte
@@ -12,6 +12,8 @@
   export let only = []
   export let headers = {}
   export let queryStringArrayFormat = 'brackets'
+  export let role = 'link'
+  export let tabindex = 0
 
   beforeUpdate(() => {
     if (as === 'a' && method.toLowerCase() !== 'get') {
@@ -36,6 +38,8 @@
     queryStringArrayFormat,
   }}
   {...as === 'a' ? { href } : {}}
+  {role}
+  {tabindex}
   {...$$restProps}
   on:focus
   on:blur


jessarcher (commented):
I think the warning is a false positive, and I don't think this is the right way to solve it.

Ideally, the `Link` component should always be used to render an `<a>` or `<button>`, which already comes with the correct role by default.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/link_role#best_practices

Adding the role unnecessarily seems like it's against the best practices, and interestingly it goes against the [`a11y-no-redundant-roles`](https://svelte.dev/docs/accessibility-warnings#a11y-no-redundant-roles) rule.

Ideally, the warning would be smart enough to only appear if someone chooses to render something other than an `a` or `button`; in that case, it should be on them to add the correct role for their situation. I'm not sure whether this is possible, though.

Assuming that isn't possible, this might be an appropriate case for an ignore per https://svelte.dev/docs/accessibility-warnings. E.g:

```
<!-- svelte-ignore a11y-a11y-no-static-element-interactions -->
```

pedroborges:
I agree with @jessarcher, just using the `svelte-ignore` rule should get rid of the warning. Please update and test your PR and I'll gladly approve it.

ingLomeland:
Whats the status on this PR? it fixes the issue with the ARIA warning on build in #1627

driesvints:
This PR seems to have stalled as the OP isn't answering anymore. It's best that someone else picks up the work in a new PR.

Isaac-alencar:
> This PR seems to have stalled as the OP isn't answering anymore. It's best that someone else picks up the work in a new PR.

I can take this one. I am struggling right now with this issue, so I can make the changes proposed by the original author.

reinink:
Decided to fix this using a `svelte-ignore` as @jessarcher suggested (see #1858), and that has been released as part of [v1.0.16](https://github.com/inertiajs/inertia/releases/tag/v1.0.16) 👍

PR #1651, opened 2023-08-18
Remove empty payload from GET requests

edgars-vasiljevs:
Currently all Inertia GET requests have `Content-Type: application/json` header applied because empty object is passed as a payload. Axios sees an object and sets content type to application/json even though request body is empty.

This is an issue because Laravel interprets these requests differently than regular GET request. See [here](https://github.com/laravel/framework/blob/b8557e4a708a1bd2bc8229bd53feecfa2ac1c6fb/src/Illuminate/Http/Concerns/InteractsWithContentTypes.php#L16)

One example where this is a problem - for example I have GET route and I want to add or update some query parameters in current request using `request()->merge(['foo' => 'bar'])`

When running `request()->query()` I expect to get `foo` parameter back since it is a GET request. Instead I get nothing. This is because `merge` runs on json input source instead of query. See [here](https://github.com/laravel/framework/blob/b8557e4a708a1bd2bc8229bd53feecfa2ac1c6fb/src/Illuminate/Http/Request.php#L349) and [here](https://github.com/laravel/framework/blob/b8557e4a708a1bd2bc8229bd53feecfa2ac1c6fb/src/Illuminate/Http/Request.php#L421) 

Setting data to `null` is enough for Axios to ignore payload.

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -342,7 +342,7 @@ export class Router {
     Axios({
       method,
       url: urlWithoutHash(url).href,
-      data: method === 'get' ? {} : data,
+      data: method === 'get' ? null : data,
       params: method === 'get' ? data : {},
       signal: this.activeVisit.cancelToken.signal,
       headers: {
--- a/packages/vue2/tests/cypress/integration/links.test.js
+++ b/packages/vue2/tests/cypress/integration/links.test.js
@@ -166,8 +166,6 @@ describe('Links', () => {
 
           cy.wait('@spy').then(({ request, response }) => {
             expect(request.url).to.eq(Cypress.config().baseUrl + '/dump/get?foo=get')
-            expect(request.headers).to.contain.key('content-type')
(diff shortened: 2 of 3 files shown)


edgars-vasiljevs:
Here is another case

`->get(..)` method is broken on any request instance derived from original request.

```php
$request = Request::createFrom(request());

request()->merge(['foo1' => 'bar']);
$request->merge(['foo2' => 'bar']);

var_export(request()->get('foo1'));
var_export($request->get('foo2'));

// GET request without content-type header
'bar'
'bar'

// GET request with content-type header
'bar'
NULL
```

For example this breaks `spatie/laravel-query-builder` package since it uses `->get()` method internally to get filter/sorting params from query.

joetannenbaum:
Closing this as we're not adding any new features or non-emergency fixes for 1.x. 

I understand the issue though, and ported it to 2.x with credit to you, thank you!

https://github.com/inertiajs/inertia/pull/2305

PR #1659, opened 2023-08-28
Fix `Link` type in React

### LINKED ISSUE #1511: Return type of the `Link` component is incorrect
### Version:

- `@inertiajs/react` version: 1.0.2

### Describe the problem:

`Link` component forward ref to `a` (or `as` prop) element. However, the typing of the `Link` component does not allow the use of the `ref` property.

Wrong type is here: https://github.com/inertiajs/inertia/blob/5c9c2707dc5f2f1f36d8eff99d731b5e42dab021/packages/react/src/Link.ts#L40

This tells TypeScript that `Link` is just a `FunctionComponent` when in fact it returns a component that can take a ref.

### Steps to reproduce:

```typescript
const someRef = useRef<HTMLLinkElement>(null);
const Test = <Link href={'/'} ref={someRef} />;
```

This code makes TypeScript angry with TS2769 alert.

### Possible solution

`InertiaLink` return type could be removed to just let TypeScript infer correct return type using `forwardRef`. If this solution suits you, I can create the PR.

Thanks!

guihigashi:
this fixes #1511 

it's a more explicit version of @Gregory-Gerard recommended solution

### DIFF
--- a/packages/react/src/Link.ts
+++ b/packages/react/src/Link.ts
@@ -37,7 +37,7 @@ export type InertiaLinkProps = BaseInertiaLinkProps &
   Omit<React.HTMLAttributes<HTMLElement>, keyof BaseInertiaLinkProps> &
   Omit<React.AllHTMLAttributes<HTMLElement>, keyof BaseInertiaLinkProps>
 
-type InertiaLink = React.FunctionComponent<InertiaLinkProps>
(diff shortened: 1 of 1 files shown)


reinink:
Hey thanks for sharing this fix! I've just tweaked it with an even similar solution but appreciate your help either way! 💪

PR #1705, opened 2023-10-18
Fix svelte adapter layouts bug

raghed-kahil:


reinink:
This has been fixed in #1763 has been released as part of [v1.0.16](https://github.com/inertiajs/inertia/releases/tag/v1.0.16) 👍

Thanks either way for this contribution!!

PR #1710, opened 2023-10-23
Export page context

samrobbins85:
This adds page context as an export, making it easier to mock the context provider for use in storybook or packages where components are used without the whole inertia application.


reinink:
Hey! Thanks for this PR, but I don't think we want to export the `PageContext` at this time, as this is very much an implementation detail and the way that you're supposed to get the page information in an Inertia app is by using the `usePage()` hook. By exposing this context we'd be making this "public API", which isn't something we want to commit to at this time.

Hope that makes sense!

Issue #1631, opened 2023-07-31
transform method in useForm hook doesn't work


aviemet:
### Version:

- `@inertiajs/react` version: 1.0.9

### Describe the problem:

I'm re-reporting an issue which was closed in the recent reset of issues, the original is #1131 

When using `useForm` from the React package, the `transform` method doesn't work. [The method is defined in the body of the `useForm` component](https://github.com/aviemet/inertia/blob/5c9c2707dc5f2f1f36d8eff99d731b5e42dab021/packages/react/src/useForm.ts#L61), and thus is re-initialized upon re-renders to be the original "placeholder" method. 

The transform method needs to be memoized in some way. I submitted #1491 which I've tested and it addresses this issue. It stores the `transform` method in a `ref` so that when it's value is updated it persists through re-renders.

saeidalidadi:
It works when I pass the callback with the `function` signature instead of the `arrow` signature.

saeidalidadi:
I fixed this by using a ref hook.
```javascript
const transformRef = useRef({} as any);
useEffect(() => {
       transformRef.current = (data: any) => {};
       transform(transformRef.current)
}, []);
```

qMaDMaNp:
v1.0.11 - the problem persists. 
If no changes have been made to the form fields, the transform() method is ignored thus not being called. =(

Zagoman:
This problem still persists

aviemet:
Nearly 3 years and this very basic error still persists. It genuinely makes me fear for the longevity of this amazing project. Inertia is the glue that binds the frontend and backend together and is exactly how I want to build projects, I really hope this isn't a foretelling of an eventual death of the project.

For anyone who needs a quick drop-in,[ I've built out a replacement for `useForm` which fixes the `transform` bug, and also adds a few other features](https://github.com/aviemet/useInertiaForm). Please feel free to use it, improve it, steal it, etc. Hopefully one day it won't be necessary.

rossity:
@reinink @jessarcher can we get a confirmation you are aware of this bug and that there is a PR open for it? I'm not sure why this bug has been ignored for years now. It's core functionality to one of the best features of this library.

https://github.com/inertiajs/inertia/pull/1718

FWIW this bug exists in Vue3 as well.

ThePlatinum:
Surprisingly, it was working fine for me until I made an update and it stopped working, I undid all the changes and it still not working.

connerbw:
> I fixed this using a ref hook.
[...]

PR #1718, opened 2023-11-06
Memoize `transform` to fix long standing `useForm` bug

aviemet:
I'm opening another PR to fix this bug (#1171, #1131, #1491, #1631)

In the React `useForm` package, the `transform` function is set as a local variable in the component. This means that if the page triggers _any_ re-render, the `transform` variable is re-initialized to the placeholder function (`(data) => data`). For this reason, it's more likely than not that the transform method does nothing, rendering [the docs](https://inertiajs.com/forms) inaccurate.

The previous PRs were closed without being read. I think the whole React community here would really appreciate this bug finally being fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,9 @@ For changes prior to v1.0.0, see the [legacy releases](https://legacy.inertiajs.
 
 ### Fixed
 
-- Add explicit children to `InertiaHeadProps` ([#1448](https://github.com/inertiajs/inertia/pull/1448))
-- Export `InertiaLinkProps` type ([#1450](https://github.com/inertiajs/inertia/pull/1450))
-- Improve React `usePage` generic type ([#1451](https://github.com/inertiajs/inertia/pull/1451))
+- Added explicit children to `InertiaHeadProps` ([#1448](https://github.com/inertiajs/inertia/pull/1448))
+- Exported `InertiaLinkProps` type ([#1450](https://github.com/inertiajs/inertia/pull/1450))
+- Improved React `usePage` generic type ([#1451](https://github.com/inertiajs/inertia/pull/1451))
 
 ## [v1.0.1](https://github.com/inertiajs/inertia/compare/v1.0.0...v1.0.1)
 
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -58,7 +58,7 @@ export default function useForm<TForm extends Record<string, unknown>>(
   const [progress, setProgress] = useState(null)
   const [wasSuccessful, setWasSuccessful] = useState(false)
   const [recentlySuccessful, setRecentlySuccessful] = useState(false)
-  let transform = (data) => data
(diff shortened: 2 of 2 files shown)


mattsims:
@reinink Any idea when this might be merged? 🙏

mochetts:
I need this fix as well! The `transform` callback is being overridden by the default `(data) => data` after doing something basic like `form.setData`.

Repro steps:
```rb
const myCallback = (data) => { 
 console.log("myCallback")
 return data
}
form.transform(myCallback)
form.post('some/url', { onSuccess: ({ props }) => {
  form.setData(props.formData)
  form.post('some/url') // "myCallback" is not printed out
})
```

mochetts:
I added a package containing this fix in the npm registry for now:
https://www.npmjs.com/package/@moraki/inertia-react

connerbw:
Yay, thanks! Now what?

dac-humi:
Please forgive me for tagging... @claudiodekker @jessarcher @thecrypticace
Can this PR get some feedback? Thanks for your time and hard work.

aviemet:
It would sure be great to get this long standing bug fixed. Hoping this is the bump that gets attention.
@reinink

pedroborges:
I just tried to reproduce the issue described here but couldn't 🤷‍♂️ Am I missing something?

I tested with the code below on the React Playground's `js/Pages/Form.tsx` file:

```js
  const form = useForm('NewUser', {
    name: '',
    company: '',
    role: '',
  })

  function submit(e) {
    e.preventDefault()
    form.transform((data) => ({ ...data, name: 'Xyz' }))
    form.post('/user')
  }
```

1. Fill out the form
    Screen Shot 2024-09-13 at 12 58 19

2. Submit it
    Screen Shot 2024-09-13 at 12 58 29

3. Click on the reset button, and fill out the form again
    Screen Shot 2024-09-13 at 13 06 16

4. Submit it
    Screen Shot 2024-09-13 at 13 06 21

pedroborges:
By the way, it seems #1607 fixes this and some other issues.

derrickreimer:
Hey @pedroborges, I've done some work on this one. Fundamentally there's a race condition at play here: if a rerender triggers _before_ the custom transform function gets called, then the custom transform will get overwritten. (More explanation here: https://github.com/inertiajs/inertia/pull/1491#issuecomment-2181524223). This is because `transform` is stored in a local variable in the component body, which is not a safe way to store a stable reference to something:

```javascript
// This will get reset every time the component renders
let transform = (data) => data

// ...

return {
  transform(callback) {
    transform = callback
  },
```

My recommendation is that we merge #1607 (with thanks to @aviemet for the work here as well), since it includes this fix and some additional memoization to mitigate unnecessary re-renders.

pedroborges:
Thanks @derrickreimer, I'll take your advice and close this on in favor of #1607.

Thanks for your work and patience @aviemet, I'll let you know when the other PR is merged.

PR #1724, opened 2023-11-11
Allow components and global identifiers to be passed

craigrileyuk:
Previously, only a string representation of an HTMLElement could be passed as the `as` prop to an Inertia Link instance.

This change allows both a Vue Component or a globally registered component reference to be used in addition to the default behaviour.


pedroborges:
We will add this feature but I'm closing this in favor of #1940 since it changes all adapters. Thanks @craigrileyuk!

PR #1728, opened 2023-11-14
[1.x] Fix doubling hash in React StrictMode

TomKalina:
Hello!

We are having the following problem in our application.

We are using `<React.StrictMode>` in the application, this causes the application to redirect twice in dev mode.
This forces us to double the previous hash during development because it calls the same router logic twice.
For example it makes `exanple.com/#hi` become `exanple.com/#hi#hi` after the update

This fix would prevent that from happening. If so, would you have handled this differently? 

Thank you very much for Inertia! It's a cool project 😉

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -1,33 +1,33 @@
-import { AxiosResponse, default as Axios } from 'axios'
+import { default as Axios, AxiosResponse } from 'axios'
 import debounce from './debounce'
 import {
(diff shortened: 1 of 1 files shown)


pedroborges (approved):
I ran the React playground project with `<React.StrictMode>` and got the double hash issue. This looks good to me 👍 

Thanks @TomKalina!

joetannenbaum (approved):
Verified! Looks good, thank you.

PR #1734, opened 2023-11-22
[1.x] Export InertiaFormProps in Vue and Svelte Adapters

Spice-King:
This is needed to solve type issues for a laravel/precognition PR that I'm creating. I blame Typescript for this, as for some reason, TS needs the type exported to be used as part of another packages exports, even indirectly, otherwise you can get TS error 4023. Precognition only does Vue 3, but also added it to Vue 2 for constancy reasons.


pedroborges (approved):
The React adapter already exports `InertiaFormProps`, I added the same change to the Svelte adapter.

pedroborges:
@Spice-King Could you please confirm whether we need to export InertiaFormProps from the adapter's index files, as suggested in #1596?

PR #1748, opened 2023-12-13
Revalidating isDirty when using the defaults() function.

felipedossantosoliveira:
When the form.defaults() function is used, it doesn't update the form.isDirty(), as it is revalidated by a watcher, which is only triggered after a change in the form. I mistakenly called the same validation trigger used in the watcher to revalidate the isDirty() within the defaults() function. This was an error because isDirty() returns false if a field is cleared and then rewritten with the same value after using defaults().

### DIFF
--- a/packages/vue3/src/useForm.ts
+++ b/packages/vue3/src/useForm.ts
@@ -86,6 +86,7 @@ export default function useForm<TForm extends FormDataType>(
           typeof fieldOrFields === 'string' ? { [fieldOrFields]: maybeValue } : fieldOrFields,
         )
       }
+      form.isDirty = !isEqual(form.data(), defaults)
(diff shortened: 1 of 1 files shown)


felipedossantosoliveira:
@jessarcher help me! 🥺

james-em:
@reinink 

This can make sense if you change your defaults to something that is equal to the current form data.

The reset() method already causes to revalidate isDirty so it really is specific to defaults(...).

Would need to be applied to all adapters though. I could help but PRs are already stacking so much so ...

pedroborges:
Thank you for your contribution, @felipedossantosoliveira! We've decided to proceed with #1985, which fixes the same bug.

PR #1751, opened 2023-12-16
Replace `lodash.isequal` with `fast-deep-equal` to save 50KB

osbre:
(needs testing)


SergkeiM:
Hi @osbre I've created a PR to replace both lodash.isequal`` and `lodash.clonedeep` https://github.com/inertiajs/inertia/pull/1791

reinink:
Hey thanks for this, but I think we're going to stick with `lodash.isequal` for now 👍

PR #1762, opened 2024-01-02
Add aria-role for non anchor tags.

danmatthews:


pedroborges:
The proposed changes to add `aria-role="link"` to non-anchor tag elements are not necessary for `<button>` elements. Buttons already have the appropriate semantic meaning and accessibility features built-in, so adding `aria-role="link"` would be redundant and potentially confusing for assistive technology users.

PR #1782, opened 2024-01-31
[1.x] Fix [scroll-region] on html element with overflow-scroll

axel-paillaud:
add `restoreScrollPositions()` in `setPage()` method if we have scroll-region. Use `this.page` instead of `page`.

This is related to [#1211](https://github.com/inertiajs/inertia/issues/1211)

In line 454 : `console.log(page.scrollRegions)` is always undefined, and I don't know why.
If we do only `console.log(page)`, we can see the scrollRegions field inside browser inspector, but I think it was populated asynchronously ?

With this version, `scroll-region` work well, and scrolling items are not reset when we submit the form (for example), but `this.page` is kind of a cheat ?


pedroborges (approved):
Tested this PR on the React and Vue3 playground following the steps to reproduce from #1211. The proposed solution fixes that issue 👍 

There's another issue that prevents your solution from working in React projects but #1803 already takes care of that. I tested the code from both PRs to confirm they fully fix the issue 🙌 

My only suggestion is move `this.restoreScrollPositions()` to outside of the `else` block, between `if` statements is fine 😉 

Thanks @axel-paillaud!

joetannenbaum (approved):
Thank you!

I tested this out on the playground and was able to replicate both the problem and the fix. Just flagging that it feels a bit odd that `page` and `this.page` are out of sync in this way, but the fix works and I don't see any side effects, so we'll just keep an eye on it.

axel-paillaud:
You're welcome ! Glad to help.
Let me know if I should reinvestigate this PR, I admit that since it's been a while, I would have to dive back into it, but it's possible if you think it's necessary.

Issue #1766, opened 2024-01-08
Application Crash with useRemember Hook During Browser History Navigation


Hasan-Mir:
### Version:

- `@inertiajs/react` version: 1.0.14

### Describe the problem:
The application crashes when navigating through the browser history using the **go back** & **go forward** buttons if the `useRemember` hook is used on a page. If we are on a page that uses the `useRemember` hook, if we navigate to another page by entering the page’s URL in the address bar (i.e., a full page visit, not using Inertia’s `Link` component) and then click on the **go back** button next to the address bar, the application throws an error and displays a white screen:
![image](https://github.com/inertiajs/inertia/assets/62181905/e0ec1e7c-713f-4179-89ec-4869677811e6)

This issue occurs with both Vite and Laravel-Mix. The only difference is that with Vite, the application doesn’t break and display a white screen.

I’ve created a [repo](https://github.com/Hasan-Mir/inertia-useRemember-bug) to demonstrate this bug on the `/login` page. The repo has two branches: `main` for the Laravel-Mix version and `vite-version` for the Vite version.

```js
export default function Login({ status, canResetPassword }) {
    const { data, setData, post, processing, errors, reset } = useForm({
        email: "",
        password: "",
        remember: false,
    });

    const [count, setCount] = useRemember(0, "count");

    // ...
}
```
There is no difference between providing a key (the second argument, in this case `"count"`) to `useRemember` and not providing one at all.

#### Laravel-Mix version:

https://github.com/inertiajs/inertia/assets/62181905/2ac761c7-ee0f-429c-9e01-f87f11a04cf3

When debugging the `app.jsx` file, it can be observed that during the **first render** after navigating back to `/login` from `/register`, the `name` argument of the `resolve` function is `Auth/Login`. However, during the second render (which I don't know why Inertia renders the components twice!!!), the `name` argument is `undefined`, which causes the error.

https://github.com/inertiajs/inertia/assets/62181905/f5fb01db-c58f-46d4-98bb-5a2af4849185

#### Vite version:

https://github.com/inertiajs/inertia/assets/62181905/e38029de-7d09-4922-a4cb-8557d6b01b3c

The same issue occurs when debugging the `app.jsx` file in the Vite version.

https://github.com/inertiajs/inertia/assets/62181905/a950677a-d7cc-4cb5-9aa6-d50b21fbde7e

By the way, this bug doesn’t show up when you’re using the Vue version of Inertia.

### Steps to reproduce:

[...]

Issue #247, opened 2020-10-06
Back Button Security Concerns


drfraker:
I believe there may be some issues around the way Inertia handles it's ajax calls in the presence of Cache-Control headers that are put in place for security. See the link below for the closed issue where this is discussed. 

The basic premise of cache control headers is that if you add 'no-cache, no-store' to cache control the app will require a full page reload even when a user clicks on the back button in the browser. This is a great way to prevent sensitive data from being displayed after a user logs out of an application and the back button is clicked. Since the app now has to make a full server request due to no cache history, the information will not be displayed and the user will be redirected back to the login page by the server. 

When using Inertia this behavior is not present. If you log into an app with Inertia where the Cache-control headers are present to protect back button clicks, logout and press the back button you still see the sensitive pages.

Is there a way to check for this header and redirect to login if the session is no longer valid?

_Originally posted by @drfraker in https://github.com/inertiajs/inertia/issues/102#issuecomment-704306367_

drfraker:
For reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

Specifically this directive:
**no-store**
The response may not be stored in any cache. Although other directives may be set, this alone is the only directive you need in preventing cached responses on modern browsers. max-age=0 is already implied. Setting must-revalidate does not make sense because in order to go through revalidation you need the response to be stored in a cache, which no-store prevents.

claudiodekker:
Hi,

Unless I'm mistaken, this really is not a bug, and `no-store` is **not** consider to be a security feature, and [the HTTP 1.1 specification](https://tools.ietf.org/html/rfc2616#section-13.13) even describes that browsers should not be expected to respect http caching rules when clicking the back button: 

> User agents often have history mechanisms, such as “Back” buttons and history lists, which can be used to redisplay an entity retrieved earlier in a session.
>
>History mechanisms and caches are different. In particular **history mechanisms SHOULD NOT try to show a semantically transparent view of the current state of a resource. Rather, a history mechanism is meant to show exactly what the user saw at the time when the resource was retrieved**.
>
>By default, an expiration time does not apply to history mechanisms. **If the entity is still in storage, a history mechanism SHOULD display it even if the entity has expired**, unless the user has specifically configured the agent to refresh expired history documents.
>
>This is not to be construed to prohibit the history mechanism from telling the user that a view might be stale.

[...]

PR #1784, opened 2024-02-01
Address history back button security issue (full page loads)

nicholaspufal:
_**Disclaimer:** I'm open to ideas on how to address this in a better way here. This isn't the same as the `popstate` custom handler people have proposed in other issues/PRs._ 

### Summary of the problem

From a security standpoint, there are two situations where Inertia restores `props` from its cache (i.e. `window.history.state`) that are concerning at the moment:

1. **During client-side navigation (e.g. after clicking on an `InertiaLink` component) a user goes back in the browser's history**
  - This relies on the `popstate` API to restore the app's state, has been discussed in a few issues in this project, and it's not what this PR is about. Still warrants attention though.

2. **After full page loads a user goes back in the browser's history**

  - This is what this PR is about

👨‍💻 This is the code affecting behavior related to number 2 from above: https://github.com/inertiajs/inertia/blob/master/packages/core/src/router.ts#L157-L163
- `page` here represents fresh Inertia props. When a backend request is performed, and Inertia props are provided via the rendered HTML, this is what `page` here includes
- `window.history.state` on the other hand represents stale props. This is the state of the Inertia props prior to what `page` contains. 

The issue is that the code is always giving preference to those stale props. Because Inertia props are the bread and butter of Inertia, they often contain sensitive information from the users, information which is cleared when they log out, but because Inertia pulls those from its cache (without any sort of configuration around this behaviour, or without checking if `page`'s props may be fresher) sensitive information may be leaked to third-parties in Inertia apps. 

### Reproducing this problem

_from an Inertia app_
1. User A is logged in and at the path `/foobar`
4. User A signs out and this causes a full `/foobar` page load 
5. User A leaves the shared computer
6. User B goes back in the browser's history
7. User B can see all sensitive information from User A

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -155,8 +155,21 @@ export class Router {
   }
 
   protected handleBackForwardVisit(page: Page): void {
-    window.history.state.version = page.version
-    this.setPage(window.history.state, { preserveScroll: true, preserveState: true }).then(() => {
+    let currentState
+
+    /* This ensures we give precedence to a fresh state.
+    'page' here holds the props from the latest backend request.
+    This prevents user A logging out and user B seeing sensitive data
+    from user A by going back in the history (shared computer)
+    */
+    if (page.props) {
+      currentState = page
+    } else {
+      window.history.state.version = page.version
+      currentState = window.history.state
+    }
+
+    this.setPage(currentState, { preserveScroll: true, preserveState: true }).then(() => {
       this.restoreScrollPositions()
       fireNavigateEvent(page)
     })


zcuric:
@nicholaspufal Hi, does this also fixes the `undefined` issue that I've address here #1786?

zcuric:
@nicholaspufal It is fixed, I just tested it in my project! 

So to  reproduce it, on the master branch: 
1. Go to a homepage, for example
2. Go to another page 
3. Click back and you should see something like this: 
image

This is caused by the same issue you fixed. My PR #1786 addresses just above mentioned issue. Your PR needs to be merged, especially if it's a security issue. I hope someone will look at this. 🙏🏻

Hasan-Mir:
Just wanted to mention that this PR fixes the `useRemember`'s bug https://github.com/inertiajs/inertia/issues/1766 as well.

nicholaspufal:
@jessarcher @reinink given comments above, please let me know if there is anything you need me to do to expedite this — and if you are ok with the approach here. 

Thank you!

alucic:
@jessarcher Please let me know if I can assist somehow to get this PR out, this would help us a lot. Every single back button is throwing an error and spamming our Sentry channel. The error is generic `Error` so it is hard to archive it or ignore it. Thanks

zcuric:
Just so you know, we used https://github.com/ds300/patch-package to apply patches to @inertiajs/core to fix these two issues. We applied, this PR and #1786.

stefanocurnis:
Please guys, this is serious. Any chance to bring attention to this via @lepikhinb ?

reinink:
Hey folks, just want to let you know that this is on our radar and we've been having discussions internally on how to best handle signing out of an Inertia.js app and clearing all the client-side history.

Unfortunately the browser's history APIs don't make this easy — there is no way to clear history state data. From the research I've been doing I think our best bet is to simply track a unique visit ID in the history state and then save the actual page data in `localStorage`, and then provide some type of mechanism within the Inertia router to clear all this history when logging out of your app. Possibly something like this:

```js
import { router } from `@inertiajs/react`

router.clearHistory()
```

Then when a user goes back or forward in history Inertia can check if that unique visit ID exists in the history state (in local storage), and if it doesn't it can perform a full inertia visit — and if the user has been logged out they'll then be redirected to the login page.

We might even be able to add some sort of server-side header that does this client-side history clearing automatically doing something like this:

```php
public function logout(Request $request): RedirectResponse
{
    Auth::logout();

    // ...

    Inertia::clearHistory();

    return redirect('/');
}
```

That's my working idea right now at least. Need to actually prototype this and see how it works. It's a little scary switching from history state to local storage for page data storage, but maybe it will just work 🤞

As for this particular PR, I'm not 100% sure it's the right change. Yes, it is a partial fix to the security concerns being addressed in #102 and #247, but it's not the right solution for non-auth related situations where you do want to use the data in the history state when going back or forward in your history. Just because it's a "full page" history visit doesn't necessarily mean it should read fresh data from the server. Anyway, going to leave this PR open either way as I may change my mind on this while working on these changes.

svengt:
@reinink Sounds like a great solution! `sessionStorage` might also be a great fit for this implementation as it automatically clears after closing a tab?

deleteme:
Would it be beneficial to encrypt page props with a key associated with the session? Upon log out, without the key, stale page props would be inaccessible.

Having a method to clear history state would still be valuable, to prevent the accumulation of useless state and avoid a `QUOTA_EXCEEDED_ERR` exception.

nicholaspufal:
Closing this PR now that we have an encrypted history with Inertia :) 

Thanks to everyone who worked on that feature

PR #1791, opened 2024-02-09
feat: tiny alternatives for `lodash.isequal` and `lodash.clonedeep` -> `just`

SergkeiM:
Hi @reinink 

The purpose of this PR is to reduce the size of `inertiajs`.

If we check the size of [@inertiajs/inertia-vue3](https://bundlephobia.com/package/@inertiajs/inertia-vue3@0.6.0)

**75%**  - belongs to `lodash.isequal` (**3.8kB** min + gzip) and `lodash.clonedeep` (**3.3kB** min + gzip)

image

In this **PR**:

`lodash.isequal` is replaced by [just-compare](https://github.com/angus-c/just?tab=readme-ov-file#just-compare) (**419B** min + gzip)
`lodash.clonedeep` is replaced by [just-clone](https://github.com/angus-c/just?tab=readme-ov-file#just-clone) (**410B** min + gzip)

**Note**: I have a custom build of `inertia/vue3` with these 2 libraries and didn't had any issues so far.

**Differences with Lodash**:
`lodash.clonedeep` - `Lodash` can merge circular references, `just-clone` only merges plain objects, regular arrays, functions and primitives `Lodash` merges additional non-plain object types, `Lodash` treats sparse arrays as dense

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2146,6 +2146,16 @@
         "node": ">=6"
       }
     },
+    "node_modules/just-clone": {
+      "version": "6.2.0",
+      "resolved": "https://registry.npmjs.org/just-clone/-/just-clone-6.2.0.tgz",
+      "integrity": "sha512-1IynUYEc/HAwxhi3WDpIpxJbZpMCvvrrmZVqvj9EhpvbH8lls7HhdhiByjL7DkAaWlLIzpC0Xc/VPvy/UxLNjA=="
+    },
+    "node_modules/just-compare": {
+      "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/just-compare/-/just-compare-2.3.0.tgz",
+      "integrity": "sha512-6shoR7HDT+fzfL3gBahx1jZG3hWLrhPAf+l7nCwahDdT9XDtosB9kIF0ZrzUp5QY8dJWfQVr5rnsPqsbvflDzg=="
+    },
     "node_modules/kleur": {
       "version": "4.1.5",
       "resolved": "https://registry.npmjs.org/kleur/-/kleur-4.1.5.tgz",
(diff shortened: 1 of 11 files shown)


SergkeiM:
@jessarcher @claudiodekker FYI

jamesh-purr:
@jessarcher @reinink 
When do we think this could be looked at and potentially merged? I need this issue resolved asap and if this does indeed resolve the issue without any problems then I am +1 on getting this merge asap.

reinink:
Hey! As noted in #1792 as well, thanks for this suggestion, but I think we're going to stick with `lodash ` for now because this is just too risky of a change as there could be unexpected side effects for not a lot of gain. Might revisit this in the future, but for now going to leave it as is 👍

RomainLanz:
Hey @reinink! 👋🏻 

To avoid any unexpected side effects, you could release a beta version with those changes and see if anyone come with an issue.
Since it is an easy change, it is also easily revertable.

Looking at the source code and test suite, it seems to be 100% compatible. Also, looking at the usage of `lodash.isequal`, the native [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) function may be enougth.

SergkeiM:
Hi @RomainLanz, @reinink 

Agree, having `lodash.isequal` and `lodash.clonedeep` removed, saves 7.1kB min + gzip combined, this is more then 50% of `vue 3` size (13.5kb)

SergkeiM:
Hi @reinink 

With 2.0 in beta, maybe is a good idea to include this as well?

RomainLanz:
I have asked that in the "Epic" issue, no answer yet https://github.com/inertiajs/inertia/issues/1804#issuecomment-2316871984

PR #1792, opened 2024-02-09
feat: smaller alternative for  `qs` -> `query-string`

SergkeiM:
Hi @reinink

The purpose of this PR is to reduce the size of `@inertiajs/core`.

Starting from version `qs@6.10.0` increased size by **60%**

If we check the size of [@inertiajs/core](https://bundlephobia.com/package/@inertiajs/core@1.0.14)

**12%** - belongs to `qs` (9.8kB min + gzip - 31.3kB min)

I believe this is a lot for simple helper to build `URL query`.

In this **PR**:

`qs` is replaced by [query-string](https://github.com/sindresorhus/query-string) (2.5kB min + gzip) x4 smaller then `qs`, but does the same thing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -912,12 +912,6 @@
       "integrity": "sha512-JCB8C6SnDoQf0cNycqd/35A7MjcnK+ZTqE7judS6o7utxUCg6imJg3QK2qzHKszlTjcj2cn+NwMB2i96ubpj7w==",
       "dev": true
     },
-    "node_modules/@types/qs": {
-      "version": "6.9.7",
-      "resolved": "https://registry.npmjs.org/@types/qs/-/qs-6.9.7.tgz",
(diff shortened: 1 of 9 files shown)


reinink:
Hey! Thanks for this suggestion, but I think we're going to stick with `qs` for now because this is just too risky of a change as there could be unexpected side effects for not a lot of gain. Might revisit this in the future, but for now going to leave it as is 👍

SergkeiM:
Hi @reinink as mentioned #1791 here we can release a beta or pre-release for people to test.

I would actually replace both libraries with just `UrlSearchParams`

jamesh-purr:
Hey @reinink 

To add to this, you say its not a lot of gain but there is a clear issue with the package being not optimised/needs improving its size.

I have opened a issue regarding this here - https://github.com/inertiajs/inertia/issues/1804

Google simply don't like the package size and is a factor in its rating on your site which affects performance and seo being that if the score in lower than it should be.

PR #1796, opened 2024-02-12
[1.x] Fix a bug with [scroll-region] elements not restoring correct scroll positions in react adapter

dsazup:
Hi, I'm using react and I run into a bug where I have a [scroll-region] element and after navigating back, the scroll position is not restored, even though the page has saved the positions correctly.

This happens because in the `restoreScrollPositions` method `this.scrollRegions()` returns an empty array. It thinks that there aren't any [scroll-region] elements, because the document body hasn't been updated yet. Even though there actually is an element. Simply wrapping the whole thing in `setTimeout` fixes it and positions are restored correctly.

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -136,7 +136,9 @@ export class Router {
 
   protected restoreScrollPositions(): void {
     if (this.page.scrollRegions) {
-      this.scrollRegions().forEach((region: Element, index: number) => {
+      // Similar to the above, we're using setTimeout to fix a bug in the React adapter where
+      // the rendering isn't fast enough, causing incorrect scroll positions in [scroll-region] elements.
+      setTimeout(() => this.scrollRegions().forEach((region: Element, index: number) => {
         const scrollPosition = this.page.scrollRegions[index]
         if (!scrollPosition) {
           return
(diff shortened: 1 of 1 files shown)


pedroborges:
I tested locally and reproduced the bug with the React adapter. Your solution fixes it, @dsazup. Thank you!

The core has been rewritten to support async requests and the `restoreScrollPositions` method has been [renamed and moved](https://github.com/inertiajs/inertia/blob/338b4c25a68ff0a62c1fedf96789d0ff4273f21c/packages/core/src/scroll.ts#L41-L60) to the `packages/core/src/scroll.ts` file. Let's wait until that work is merged into `master` to rebase and merge this one 👍

pedroborges (changes requested):
I tested locally and reproduced the bug with the React adapter. Your solution fixes it, @dsazup. Thank you!

The core has been rewritten to support async requests, the `restoreScrollPositions` method has been [renamed and moved](https://github.com/inertiajs/inertia/blob/338b4c25a68ff0a62c1fedf96789d0ff4273f21c/packages/core/src/scroll.ts#L41-L60) to the `packages/core/src/scroll.ts` file. Let's wait until that work is merged into `master` to rebase and merge this one 👍

pedroborges:
#1803 also fixes a scroll restoration bug in the React adapter. They conflict with each other but that will be easy to consolidate.

pedroborges:
Fixed by #1980 ✅

Issue #1802, opened 2024-02-20
React: ScrollReset is slightly out of sync with Page changes, causing a flash before navigating


oscarnewman:
### Version:

- `@inertiajs/react` version: 1.0.14

### Describe the problem:

I noticed an issue working on a new Inertia React/Laravel app  where page navigations via `Link` would suffer what looked like a flash of content before moving to the next page. 

(this is my fist time using Inertia and it's been an awesome experience, btw!)

Looking more closely, I found that the "scroll reset" step was occurring an instant before the page itself changed to the new content. This means that if a user is scrolled down at all on a page _and_ they navigate to another page via Inertia, they'll see an instant where their current page scrolls all the way to the top before they see the next page.

On smaller/lighter weight pages, this is less noticeable. On heavier pages or on mobile, this was causing decent Jank.

I've attached a video of a simple reproduction in the React playground, and one from my own app.

It's less consistent/harder to notice in the React playground. I think this might be because the `render` pass is just faster with less content and is less likely to get out of sync. I recommend scrubbing through the video frame by frame slowly to see what's happening.

In the playground:

https://github.com/inertiajs/inertia/assets/4020585/ed6e9667-507a-4e1b-b772-3a056b73f09a

In my own app:

https://github.com/inertiajs/inertia/assets/4020585/7ad0381c-0d90-4cc5-b542-2bafe2dade87

### Steps to reproduce:

1. Create a page that causes the browser viewport to scroll
2. Include a link to another page
3. Click that link while scrolled down, and observe a flash of scroll-to-top before the transition

#### Reproduction Repo:

I've updated the "React Playground" in the repo to use a Sticky header -- this makes it possible to click a navigation link while scrolled down on a page.

That change is visible here: https://github.com/inertiajs/inertia/compare/master...oscarnewman:inertia:reproduction

I'm able to reproduce the issue in about 1/4 navigations from the bottom of the "Article" page to any other page.

### Root cause / Solution

I dug into the Inertia code a bit to see if I could figure out what's happening. I think it primarily occurs in this `router.setPage` function (comments are my annotations):

```typescript
protected setPage(
    page: Page,
    {
      visitId = this.createVisitId(),
      replace = false,
      preserveScroll = false,
      preserveState = false,
    }: {
      visitId?: VisitId
      replace?: boolean
      preserveScroll?: PreserveStateOption
      preserveState?: PreserveStateOption
    } = {},
  ): Promise<void> {
[...]

Issue #1698, opened 2023-10-05
Does not reliably restore scroll position using browser/mouse nav buttons in Firefox.


benjivm:
### Version:

- `@inertiajs/vue3` version: `^1.0.11`

### Describe the problem:

Browser/mouse button page navigation (forward/back) does not restore scroll position. Oddly, it does appear to work fine if the content is static (not served from a controller, e.g.)

### Steps to reproduce:

Reproduction repo: https://github.com/benjivm/inertiajs-firefox-scroll-position

Open the page in Firefox (this issue does not affect Chrome).

Scroll down to about the middle of the page where the link and explanation are then navigate with your browser's nav buttons or mouse button.[

[This polyfill](https://github.com/janpaul123/delayed-scroll-restoration-polyfill) makes it more reliable but is jarring to use.

This relates to this issue #1459

Hasan-Mir:
I have the same issue (mostly in Firefox but I think I saw it a few times in Chrome as well).

ivalkenburg:
Having the same issue. It does not always scroll to top when navigating. Specially when further down the page. Very odd.

fritz-c:
Inertia's popstate listener (popstate being the event called when the back button is used) will asynchronously load the previous component, rendering it when it has loaded.
https://github.com/inertiajs/inertia/blob/f34f373692a964933e4272fc49fa48552301e568/packages/core/src/router.ts#L480-L492

The problem arises in that the browser wants to restore the previous scroll position as soon as the popstate listener exits, when the url changes, and at that point in time, the previous component has not loaded yet. So what you might be seeing is your pre-back page getting scrolled down (by browser native restoration) but hitting the bottom of the scroll bar, after which point the previous component gets rendered, just at the scroll position you managed to get to from the previous page.

I looked for an elegant solution around this issue, but failing to find one, made one that at least looks good from an end-user's perspective. It's for Vue+Typescript, but could be adapted for other environments. You would want to run this only once, probably in a global layout or something.

```ts
import { onMounted, onUnmounted } from 'vue';
import { debounce } from 'lodash-es';
import { router } from '@inertiajs/vue3';

/**
 * Restores the window scroll position when navigating back to a page.
 *
 * The 'popstate' event handler in inertia loads the component from the previous
 * page asynchronously, so browser-native scroll restoration is triggered before
 * the new component is mounted, resulting in the scrollbar banging against the
 * bottom of the window in the case that the page before navigating back is
[...]

PR #1803, opened 2024-02-20
[1.x] Fix a flash of scrolling-to-top on page changes by using `useLayoutEffect`

oscarnewman:
This is my (attempted) fix for https://github.com/inertiajs/inertia/issues/1802. I outlined the issue in detail there but tl;dr:

- Inertia/React creates a race condition when changing pages, where resetting the scroll position to (0,0) often happens an instant before the page component itself changes to the next page
- This creates a flash of unexpected content when you'll scrolled down a page and click a `<Link>`.

A huge caveat to all of this is that I'm brand new to Inertia as a user, let alone looking at this codebase, so please take anything I suggest with a grain of salt. 

Based on how [Remix/React-router handle this successfully](https://github.com/remix-run/react-router/blob/5d475a2af140fee92a8a49f0e40d683087aa2b78/packages/react-router-dom/index.tsx#L1842) with a `useLayoutEffect`, I've slightly adapted a few things:

1. each framework package (react/vue/etc) can now specific in it's `router.init` whether it will handle resetting scroll position itself (vs letting the router core decide when to do it)
2. the `swapComponent` callback within `router.init` now passes a new argument, `preserveScroll`, similar to how it passes `preserveState`. This and (1) mean React has the information needed to manage scroll resetting in it's own render lifecycle
3. `router.resetScrollPositions()` is now public
4. Existing calls to `router.resetScrollPositions()` adjacent to the `swapComponent` call have been replaced with `router.resetScrollPositionsIfNotHandledExternally()` -- which just does nothing if the `handleScrollResetsExternally` flag is set by the framework package. Other calls to `router.resetScrollPositions()` weren't changed as they don't interact with the `swapComponent` function and don't seem to suffer the same issue

I've also updated the React playground with a `sticky` header so that the fix can be validated there as well.

Testing so far:
1. Tested the fix in the playground ✅
2. Linked the rebuild inertia into my own app, and the more aggressive flashes are gone ✅
3. Ensured that `preserveScroll` still works on `Link` and `Router` calls ✅

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -40,19 +40,23 @@ export class Router {
   protected navigationType?: string
   protected activeVisit?: ActiveVisit
   protected visitId: VisitId = null
+  protected handleScrollResetExternally: boolean = false
 
   public init({
     initialPage,
     resolveComponent,
     swapComponent,
+    handleScrollResetExternally
   }: {
     initialPage: Page
     resolveComponent: PageResolver
     swapComponent: PageHandler
+    handleScrollResetExternally?: boolean
   }): void {
     this.page = initialPage
     this.resolveComponent = resolveComponent
     this.swapComponent = swapComponent
+    this.handleScrollResetExternally = !!handleScrollResetExternally
 
     this.setNavigationType()
     this.clearRememberedStateOnReload()
@@ -116,22 +120,30 @@ export class Router {
     })
   }
 
-  protected resetScrollPositions(): void {
-    window.scrollTo(0, 0)
-    this.scrollRegions().forEach((region) => {
-      if (typeof region.scrollTo === 'function') {
-        region.scrollTo(0, 0)
-      } else {
-        region.scrollTop = 0
-        region.scrollLeft = 0
+  public resetScrollPositions(): void {
+      window.scrollTo(0, 0)
+      this.scrollRegions().forEach((region) => {
+        if (typeof region.scrollTo === 'function') {
+          region.scrollTo(0, 0)
+        } else {
+          region.scrollTop = 0
+          region.scrollLeft = 0
+        }
+      })
+      this.saveScrollPositions()
+      if (window.location.hash) {
+        // We're using a setTimeout() here as a workaround for a bug in the React adapter where the
+        // rendering isn't completing fast enough, causing the anchor link to not be scrolled to.
+        setTimeout(() => document.getElementById(window.location.hash.slice(1))?.scrollIntoView())
       }
-    })
-    this.saveScrollPositions()
-    if (window.location.hash) {
-      // We're using a setTimeout() here as a workaround for a bug in the React adapter where the
-      // rendering isn't completing fast enough, causing the anchor link to not be scrolled to.
-      setTimeout(() => document.getElementById(window.location.hash.slice(1))?.scrollIntoView())
+  }
+
+  protected resetScrollPositionsIfNotHandledExternally(): void {
+    if (this.handleScrollResetExternally) {
+      return
     }
+
+    this.resetScrollPositions()
   }
 
   protected restoreScrollPositions(): void {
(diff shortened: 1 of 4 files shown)


packages/core/src/router.ts
```diff
@@ -484,7 +496,7 @@ export class Router {
       Promise.resolve(this.resolveComponent(page.component)).then((component) => {
         if (visitId === this.visitId) {
           this.page = page
-          this.swapComponent({ component, page, preserveState: false }).then(() => {
+          this.swapComponent({ component, page, preserveState: false, preserveScroll: true }).then(() => {
             this.restoreScrollPositions()
```
oscarnewman, line 500: As far as I can tell, `restoreScrollPositions()` doesn't suffer from this same issue in my testing. I'm not entirely sure why, to be honest.

packages/react/src/App.ts
```diff
@@ -25,17 +26,25 @@ export default function App({
...
     router.init({
       initialPage,
       resolveComponent,
-      swapComponent: async ({ component, page, preserveState }) => {
+      swapComponent: async ({ component, page, preserveState, preserveScroll }) => {
         setCurrent((current) => ({
           component,
           page,
           key: preserveState ? current.key : Date.now(),
+          preserveScroll: !!preserveScroll,
```
oscarnewman, line 44: both `preserveState` and `preserveScroll` are typed as a `PreserveStateOption` in the router core, although they actually seem to be resolved to booleans where it matters in the usage here?

This is where my knowledge of this codebase starts to feel super minimal -- so I may be way off here

playgrounds/react/resources/js/Components/Layout.tsx
```diff
@@ -5,7 +5,7 @@ export default function Layout({ children }) {
 
   return (
     <>
-      <nav className="flex items-center space-x-6 bg-slate-800 px-10 py-6 text-white">
+      <nav className="sticky top-0 flex items-center space-x-6 bg-slate-800 px-10 py-6 text-white">
```
oscarnewman, line 8: allows for reproduction to be tested -- you need to be able to click a link while in a scrolled-down state

packages/core/src/router.ts
```diff
@@ -40,19 +40,23 @@ export class Router {
   protected navigationType?: string
   protected activeVisit?: ActiveVisit
   protected visitId: VisitId = null
+  protected handleScrollResetExternally: boolean = false
```
pedroborges, line 43: Suggestion: `hasCustomScrollHandler`

packages/core/src/router.ts
```diff
@@ -134,6 +138,14 @@ export class Router {
     }
   }
 
+  protected resetScrollPositionsIfNotHandledExternally(): void {
```
pedroborges, line 141: Suggestion: `attemptScrollReset`

pedroborges (changes requested):
Just tested this locally and reproduced it, adding a red background to the article title and throttling CPU and network makes it easier to spot it. Nice work @oscarnewman 👏 

## Before fix

https://github.com/user-attachments/assets/6b71d527-ec85-4ce2-9c91-908ca30e5068

## After fix

https://github.com/user-attachments/assets/23adf332-f25c-405e-b7e4-3b5134b9016d

The core has been rewritten to support async requests, the code that handles scroll [moved](https://github.com/inertiajs/inertia/blob/338b4c25a68ff0a62c1fedf96789d0ff4273f21c/packages/core/src/scroll.ts) to the `packages/core/src/scroll.ts` file. Let's wait until that work is merged into `master` to rebase and merge this one 👍 

#1796 also fixes a scroll restoration bug in the React adapter. These PRs conflict with each other but that will be easy to consolidate.

packages/react/src/App.ts
```diff
@@ -25,17 +26,25 @@ export default function App({
...
         }))
       },
+      handleScrollResetExternally: true,
```
pedroborges, line 47: Alternatively this property could take a callback that the router could call internally.

pedroborges:
While searching for ways simplify this solution and trying to avoid handling scroll outside `core`, I came across a pattern called double RAF. It combines two `requestAnimationFrame` to ensure that a callback runs after the next repaint. Vue uses this pattern internally:

- [Double RAF w/ callback](https://github.com/vuejs/core/blob/dad673809929c084dcb8e42640eb7daa675d4ea4/packages/runtime-dom/src/components/Transition.ts#L326-L330)
- [Double RAF w/ promise](https://github.com/vuejs/core/blob/dad673809929c084dcb8e42640eb7daa675d4ea4/packages/vue/__tests__/e2e/e2eUtils.ts#L157-L165)

I tested it and it solves the scrolling glitch in React. Since this pattern is safe for the other adapters, we can implement a fix inside `core` without requiring any changes to the React adapter.

We can also replace it where we currently use a `setTimeout` to "delay" scrolling.

pedroborges:
Found other issues related to this one: #1698, #1816, and #1922.

PR #1835, opened 2024-03-28
Fix svelte 4 warning for ARIA role

Isaac-alencar:
#1627

If as is neither `'a'` nor `'button'`, it spreads the `{ role, tabindex }` object.
if role is `'button'`, it won't be included as an attribute on the element, ensuring that the correct role is applied based on the element type specified by the as variable.

### DIFF
--- a/packages/svelte/src/Link.svelte
+++ b/packages/svelte/src/Link.svelte
@@ -12,6 +12,8 @@
   export let only = []
   export let headers = {}
   export let queryStringArrayFormat = 'brackets'
+  export let role = 'link'
+  export let tabIndex = 0
 
   beforeUpdate(() => {
     if (as === 'a' && method.toLowerCase() !== 'get') {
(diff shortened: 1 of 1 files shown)


reinink:
Hey! Thanks for this contribution, but I just tested this, and I'm still getting an ARIA warning:

```
vite v4.5.2 building SSR bundle for production...
transforming (27) ../../packages/svelte/src/index.js8:17:43 AM [vite-plugin-svelte] /Users/jonathanreinink/Sites/inertiajs-inertia/packages/svelte/src/Link.svelte:27:0 A11y: <svelte:element> with click, dblclick, mousedown, mousemove, mouseout, mouseover, mouseup handlers must have an ARIA role
``

What am I missing? 🤔

reinink:
Think I'm going to just add a `svelte-ignore` for now to solve this, as I'm nervous about adding a `role` or `tabIndex`. I'm *pretty* sure this is a false positive, as @jessarcher suggested in #1635. Thanks either way for this contribution!! 🤙

Isaac-alencar:
> Think I'm going to just add a `svelte-ignore` for now to solve this, as I'm nervous about adding a `role` or `tabIndex`. I'm _pretty_ sure this is a false positive, as @jessarcher suggested in #1635. Thanks either way for this contribution!! 🤙

sounds good to me! thank you for the support on this one!

Issue #1844, opened 2024-04-09
Link component cannot be rendered as a custom component due to toLowerCase()


ammagv:
### Version:

- `@inertiajs/react` version: 1.0.14

### Describe the problem:

https://github.com/inertiajs/inertia/blob/2ebd86224d2c12fdad4e4b23276129832754cf6d/packages/react/src/Link.ts#L114

Because of the above line, it is not possible to use a custom component in the "as" of a Link component. Otherwise, it would be possible to use it as shown below:
```
import { Link } from '@inertiajs/react';
import { Button } from './components';

...
<Link as={Button}>Foo</Link>
...
```

Would it be possible to remove the above line to enable this use case?

_Originally brought up by @bobbypiper in https://github.com/inertiajs/inertia/discussions/998_

driesvints:
Heya thank you for this issue. This is more a feature request so I suggest to just attempt a PR to see if it gets accepted.

PR #1846, opened 2024-04-09
Remove toLowerCase() on Link "as" so custom components can be used

ammagv:
In response to this issue: https://github.com/inertiajs/inertia/issues/1844


ammagv:
In that case I assume there is no interest in changing this behaviour, I'll close this

PR #1857, opened 2024-04-15
Use `Omit` instead of `Exclude` in router types

### LINKED ISSUE #1856: Router methods type issues with wrong use of Exclude
In `router`, `get`, `reload`, `replace`, `post`, `put`, `patch`, `delete` typed wrong. It uses

```ts
Exclude<VisitOptions, FieldsToHide>
```
instead of
```ts
Omit<VisitOptions, FieldsToHide>
```

This causes behaviours like 'preserveState', 'preserveScroll' is included in `reload` functions `options` parameter type.

aktasumut34:
This pr fixes #1856

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -498,11 +498,7 @@ export class Router {
     }
   }
 
-  public get(
-    url: URL | string,
-    data: RequestPayload = {},
-    options: Omit<VisitOptions, 'method' | 'data'> = {},
-  ): void {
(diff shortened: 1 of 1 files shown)


reinink:
@aktasumut34 Hey thanks! Just tested this out and you're right, that seems to work better 👍

reinink:
This fix has been released as part of [v1.0.16](https://github.com/inertiajs/inertia/releases/tag/v1.0.16) 👍

PR #1761, opened 2023-12-26
include component CSS in head

buhrmi:
Fixes #1760

PR #1866, opened 2024-04-21
Svelte TypeScript support

punyflash:
This PR adds complete TypeScript support for Svelte adapter. Been tested by me on multiple projects and contributors from [@westacks/inertia-svelte](https://github.com/westacks/inertia-svelte). Works perfectly for Svelte 3 and 4.

TypeScript support based on #1614, however:
- Adapter bundled using SvelteKit to compile types for Svelte TS components (required if you don't use TS in your project)
- There is no `cjs` files, since Svelte compiler requires ESM modules anyway.

Also includes fixes to SSR CSS rendering to prevent yanking on initial page load: #1761 (will not be needed in Svelte 5)


packages/svelte/package.json
```diff
@@ -19,17 +20,54 @@
...
+        "./dist/index.d.ts"
+      ]
+    }
+  },
   "dependencies": {
-    "@inertiajs/core": "1.0.16",
-    "lodash.clonedeep": "^4.5.0",
-    "lodash.isequal": "^4.5.0"
+    "@inertiajs/core": "^1.0.16",
+    "lodash": "^4.5.0"
```
james-em, line 71: These days we shouldn't rely on lodash anymore as it is no longer maintained. Importing the whole library now is also worst than importing the 2 small one.

I believe "lodash.clonedeep" could be replaced by the native structuredClone https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

Another alternative would be switching from lodash to Remeda

cloneDeep: https://remedajs.com/docs/#clone
deepEqual: https://remedajs.com/docs/#isDeepEqual

This lib is maintained and free of vulnerabilities

packages/svelte/src/lib/createInertiaApp.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    return
+  }
+
+  // Svelte types are written for the DOM API and not the SSR API.
+  const { html, head, css } = (SSR as any).render({ id, initialPage })
+
+  return {
+    body: html,
+    head: [head, `<style data-vite-css>${css.code}</style>`],
```
james-em, line 88: This is different from the current implementation, is it expected?

packages/svelte/package.json
```diff
@@ -19,17 +20,54 @@
...
+    "dist",
+    "!dist/**/*.test.*",
+    "!dist/**/*.spec.*"
+  ],
   "peerDependencies": {
     "svelte": "^3.20.0 || ^4.0.0"
   },
+  "devDependencies": {
+    "@sveltejs/adapter-auto": "^3.2.0",
+    "@sveltejs/kit": "^2.5.5",
```
james-em, line 39:
```
"@sveltejs/adapter-auto": "^3.2.0",
"@sveltejs/kit": "^2.5.5",
```

Do we actually need svelte kit?

I believe a smaller config would be 

vite.config.ts
```js
import { defineConfig } from 'vite'
import { svelte } from '@sveltejs/vite-plugin-svelte'
export default defineConfig({
  plugins: [
    svelte()
  ],
})
```

svelte.config.js
```js
import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';

export default {
  preprocess: [vitePreprocess()]
};
```

packages/svelte/package.json
```diff
@@ -19,17 +20,54 @@
...
+    ".": {
+      "types": "./dist/index.d.ts",
+      "svelte": "./dist/index.js"
+    },
+    "./server": {
+      "types": "./dist/server.d.ts",
+      "svelte": "./dist/server.js"
+    }
+  },
+  "import": "./dist/index.js",
```
james-em, line 59: Not sure if it does anything

james-em (changes requested):

packages/svelte/package.json
punyflash: The whole library is [not imported](https://github.com/inertiajs/inertia/pull/1866/files#diff-42290667298535b040cc6d17c720543abdc4c74ed4e30d1276ac827dcbe7aaa0R3-R4), so there is no extra bundle size usage, however using 1 package instead of 2 is a bit cleaner.

Switching to other library or ditching it as a whole should be consideration for other PR.

packages/svelte/package.json
punyflash: That's default scaffolding for Svelte library projects. Refer to [official documentation](https://kit.svelte.dev/docs/packaging).

reinink:
@punyflash Thanks so much for putting this together! Going to do my best to get this reviewed myself and by @pedroborges (he knows more about Svelte than I and has helped with this adapter in the past).

@jakobbouchard would love your 👀 on this one too if you have time, no pressure though!

@james-em thanks for reviewing already 👍

punyflash:
Ok, I'll try to keep it up to date, but hopefully it not get staled as well

jamesst20:
@pedroborges @jakobbouchard 

Any news about this? I would hate to see this stale again :( 

I have generated `d.ts` files that can be imported in any project from this branch in the mean time. 
(Only edit I made was adding `export type VisitOptions = VisitOptions` to `useForm.d.ts` because I needed it.

[inertia-types.zip](https://github.com/inertiajs/inertia/files/15178005/inertia-types.zip)

I have been using it for the past week without any issue.

punyflash:
@jamesst20, as for now, you may use package [`@westacks/inertia-svelte`](https://github.com/westacks/inertia-svelte) instead of `@inertiajs/svelte`. I've been using and maintaining it over half a year. There is also Svelte 5 support in `next` release branch.

jamesst20:
> @jamesst20, as for now, you may use package [`@westacks/inertia-svelte`](https://github.com/westacks/inertia-svelte) instead of `@inertiajs/svelte`. I've been using and maintaining it a few months for now and works perfectly. There is also Svelte 5 support in `next` release branch. As soon as it gets merged, I'll work on Svelte 5 support there as well

Thanks, I agree it appears the official repository is not well maintained when we take a look at how many valid issues/prs are being issued and just left in the dark. It's a shame considering how great it could get over time with active development.

However it seems [@westacks/inertia-svelte](https://github.com/westacks/inertia-svelte) wasn't forked from this repository which is a bummer for us as an organization as we can't tell how far behind/ahead it is from the official repository. It's likely only ahead but there is no way to tell. By the way @james-em  is my organization account. I didn't realize I was signed on this account at the time.

jakobbouchard:
I didn't have any time to test it, but from going through the diffs, it seems to be fine! I wish I could've helped more, but I no longer work in web development, so I don't really have my environments setup anymore and not a lot of free time either.

reinink:
@punyflash Hey just did a bunch of testing on this PR, and things seem to work well! I also made a bunch of small formatting tweaks.

Still waiting on @pedroborges to have a look at this PR, but if he doesn't have a chance within the next couple days I think I'll just merge it in.

Thanks again for all your work on this!

packages/svelte/package.json
```diff
@@ -19,17 +20,54 @@
...
+    "@sveltejs/package": "^2.3.0",
+    "@sveltejs/vite-plugin-svelte": "^3.0.2",
+    "publint": "^0.2.7",
+    "svelte": "^4.2.12",
+    "svelte-check": "^3.6.8",
+    "tslib": "^2.6.2",
+    "typescript": "^5.4.3",
+    "vite": "^5.2.7"
+  },
+  "exports": {
```
kevin82222, line 49: when I run `npm run build` locally, it does not generate the `dist` folder and the corresponding `.js` and `.d.ts` files as described.
Could you please explain how to properly build these files? What settings or additional tools do I need to install to generate the file structure described in the exports configuration?
Are there any other points to note?

james-em: It should. Do you have any kind of error?

kevin82222: Oh sorry! It's my mistake, forgot `run install` before `build`
It works!
Thanks every working.

reinink:
@jakobbouchard Hey I noticed that your original PR didn't use SvelteKit to compile types — was there any particular reason for that?

packages/svelte/src/lib/useForm.ts
```diff
@@ -0,0 +1,278 @@
...
+  RequestPayload,
+  VisitOptions,
+} from '@inertiajs/core'
+import { router } from '@inertiajs/core'
+import type { AxiosProgressEvent } from 'axios'
+import cloneDeep from 'lodash/cloneDeep'
+import isEqual from 'lodash/isEqual'
+import { writable, type Writable } from 'svelte/store'
+
+interface InertiaFormProps<TForm extends Record<string, unknown>> {
```
jamesst20, line 17: We can improve a little the typing here using the same types that React uses.
See here https://github.com/jamesst20/inertia/blob/main/packages/svelte5/src/lib/useForm.svelte.ts#L19

I also was able to get rid of all ts-ignore

packages/svelte/src/lib/useForm.ts
```diff
@@ -0,0 +1,278 @@
...
+  isDirty: boolean
+  errors: Partial<Record<keyof TForm, string>>
+  hasErrors: boolean
+  progress: Progress | null
+  wasSuccessful: boolean
+  recentlySuccessful: boolean
+  processing: boolean
+  setStore(data: InertiaFormProps<TForm>): void
+  setStore(key: keyof InertiaFormProps<TForm>, value?: any): void
```
jamesst20, line 26: I believe `setStore` should not include the attributes or methods of the form. Should be typed to the data.
In my branch I did
```
  setStore(data: TForm): void
  setStore(key: keyof TForm, value?: FormDataConvertible): void
  defaults(): this
  defaults(fields: Partial<TForm>): this
  defaults(field?: keyof TForm, value?: FormDataConvertible): this
```

packages/svelte/src/lib/useForm.ts
```diff
@@ -0,0 +1,278 @@
...
+          }
+        },
+        onStart: (visit: PendingVisit) => {
+          this.setStore('processing', true)
+
+          if (options.onStart) {
+            return options.onStart(visit)
+          }
+        },
+        onProgress: (event: AxiosProgressEvent) => {
```
jamesst20, line 187: Looking internally, it appears event can be undefined too

jamesst20 (changes requested):

jamesst20:
> @jakobbouchard Hey I noticed that your original PR didn't use SvelteKit to compile types — was there any particular reason for that?

It seems he wrote a an ESBuild script by hand https://github.com/inertiajs/inertia/pull/1614/files#diff-e1d807461eb96d0bbba189479a9a91e7fa45dad230985c34563619bcb6d19e01R1 instead of using Sveltekit which is the documented approach. When you run `npm create svelte` you can pick the option to build a library and it setups Sveltekit.

He probably did that to follow what was done in other adapters like react https://github.com/inertiajs/inertia/blob/master/packages/react/build.js

Doing so is more likely to cause pain later. SvelteKit is recommending to bundle ESM instead of CJS

reinink:
@james-em Hey! Just merged this PR in, but before we tag the release would you mind submitting your recommended changes as a new PR (or multiple PRs if that makes more sense)?

@punyflash Huge thanks for all your efforts here! 💪

jamesst20:
> @james-em Hey! Just merged this PR in, but before we tag the release would you mind submitting your recommended changes as a new PR (or multiple PRs if that makes more sense)?
> 
> @punyflash Huge thanks for all your efforts here! 💪

Will do as quickly as possible! Thanks!

reinink:
@james-em Thanks buddy!

PR #1864, opened 2024-04-19
Pass the full page object to resolve component

james-em:
See discussion #1863

Issue #1670, opened 2023-09-09
Svelte page component instance is not refreshed


edgars-vasiljevs:
### Version:

- `@inertiajs/svelte` version: 1.0.11

### Describe the problem:

Page component is not refreshed (destroyed and recreated) when navigating to the same page.

Per documentation page component is supposed to refresh - "By default, page visits to the same page create a fresh page component instance."

### Steps to reproduce:

- fresh laravel + inertia + svelte install
- routes:
```php
Route::get('/1', function () {
    return Inertia::render('Index', [ 'name' => 'Foo' ]);
});
Route::get('/2', function () {
    return Inertia::render('Index', [ 'name' => 'Bar' ]);
});
```
- Index.svelte
```javascript
<script>
    import { useForm, inertia } from '@inertiajs/svelte'
    export let name;
    
    const form = useForm({
        'name': name
    });

    $: console.log('Name property updated?', name); // this runs
</script>

Name: <input type="text" bind:value={$form.name} >

<a use:inertia href="/1">Link to Foo</a>
<a use:inertia href="/2">Link to Bar</a>
```

Currently only exported properties are updated but page component instance stays the same.
This causes an issue where form is not updated when navigating to the same form but with different data.

Tried same scenario with Vue2 and it's updating form as expected. Component is being re-mounted every time.
- Index.vue
```javascript
<template>
  <div>
    Name: <input type="text" v-model="form.name"><br />
    <Link href="/1">Link to Foo</Link><br />
    <Link href="/2">Link to Bar</Link><br />
  </div>
</template>

<script>
import { Link } from '@inertiajs/vue2'
export default {
  components: { Link },
  props: { name: String },
  remember: 'form',
  data() {
    return {
      form: this.$inertia.form({
        name: this.name,
      }),
    }
  },
   mounted() {
    console.log('component is being (re)mounted')
   }
}
</script>
```

edgars-vasiljevs:
Not quite confident making PR but here is a fix I came up with and deployed to prod. It has been running for couple of weeks now and I haven't noticed any issues so far - https://github.com/edgars-vasiljevs/inertiajs-svelte/pull/1/files

It uses `#key` logic block to re-render page component whenever component's key changes. 
It checks whether swapped component is the same as current one and component has no children (is not layout).

Hope this gives more clues.

ianautiyal:
anyone any update on this?

danielstgt:
[...]

PR #1872, opened 2024-05-10
Svelte 5: Minimalest upgrade

jamesst20:
- Upgraded playground to Svelte 5
- Fixed SSR breaking changes
- packages.json: Allow Svelte 5 installation
- Improvements: Render CSS properly for Svelte 3 & 4 https://github.com/inertiajs/inertia/pull/1761
- Svelte 5 conventions used: None

**Half-Breaking change**
 
Svelte 5 changed how SSR rendering works.  Under Svelte 3&4, you can call MyComponent.render.
Under Svelte 5 you must import the render method from a package that doesn't exist under Svelte 3 & 4 making impossible to have compatibility for both unless we extract the component rendering outside.

I have updated the setup function to return the rendered result when in SSR mode.
```
// ssr.js
createServer((page) =>
  createInertiaApp({
    page,
    resolve: (name) => { ... },
    setup({ App, props }) {
      // Svelte 4: return App.render(props)
      // Svelte 5:
      return render(App, { props })
    },
  }),
)
```

I have added a warning when the implementation is not ok.

```
throw new Error(`setup(...) must return rendered result when in SSR mode.`)
```

I call this an half-breaking change because Svelte 4 apps will not break. I added a fallback with warning:
```ts
    if (!result && typeof (SSR as SSRComponentType).render !== 'function') {
      throw new Error(`setup(...) must return rendered result when in SSR mode.`)
    } else if (!result) {
      console.warn('Deprecated: in SSR mode setup(...) must be defined and must return rendered result in SSR mode.')
      console.warn('For Svelte 5: `return render(App, { props })` or for Svelte 4: `return App.render(props)`')
      result ||= (SSR as SSRComponentType).render!({ id, initialPage })
    }
```

**Warning to Svelte 5 users**

- You still need to use `$` syntax when importing reactive state from Inertia such as $page
- Super reactive state will not work! Meaning `$form.something.push({...})` will not trigger state refresh. This is likely a really huge bummer so you may consider using my fork adapter instead @jamesst20/inertia-svelte5. Hopefully Inertia maintainers change their mind about having a separate Svelte 5 adapter.
- You still need to do stuff like `$form.something = $form.something` to trigger state refresh on updated object without creating a full copy before updating even if you use Svelte 5

### DIFF
--- a/packages/svelte/src/lib/createInertiaApp.ts
+++ b/packages/svelte/src/lib/createInertiaApp.ts
@@ -88,7 +88,7 @@ export default async function createInertiaApp({
     } else if (!result) {
       console.warn('Deprecated: setup(...) must be defined and must return rendered result in SSR mode.')
       console.warn('For Svelte 5: `return render(App, { props })` or for Svelte 4: `return App.render(props)`')
-      result ||= (SSR as SSRComponentType).render!({ id, initialPage })
+      result = (SSR as SSRComponentType).render!({ id, initialPage })
     }
 
     const { html, head, css } = result
--- a/packages/svelte/src/lib/createInertiaApp.ts
+++ b/packages/svelte/src/lib/createInertiaApp.ts
@@ -1,22 +1,24 @@
 import { router, setupProgress, type InertiaAppResponse, type Page } from '@inertiajs/core'
 import type { ComponentType } from 'svelte'
-import SvelteApp from './components/App.svelte'
-import SSR from './components/SSR.svelte'
+import App from './components/App.svelte'
+import SSR, { type SSRProps } from './components/SSR.svelte'
 import store from './store'
 import type { ComponentResolver, InertiaComponentType } from './types'
+
+type SSRRenderResult = { html: string; head: string; css?: { code: string } }
+type SSRComponentType = ComponentType<SSR> & { render?: (props: SSRProps) => SSRRenderResult }
 
 interface CreateInertiaAppProps {
   id?: string
   resolve: ComponentResolver
   setup: (props: {
     el: Element
-    // @ts-ignore
-    App: ComponentType<SvelteApp>
+    App: ComponentType<App> | SSRComponentType
     props: {
       initialPage: Page
       resolveComponent: ComponentResolver
     }
-  }) => void | SvelteApp
+  }) => void | SSRRenderResult
   progress?:
     | false
     | {
@@ -67,24 +69,37 @@ export default async function createInertiaApp({
     if (progress) {
       setupProgress(progress)
     }
-
-    setup({
-      el,
-      App: SvelteApp,
-      props: {
-        initialPage,
-        resolveComponent,
-      },
-    })
-
-    return
   }
 
-  // Svelte types are written for the DOM API and not the SSR API.
-  const { html, head, css } = (SSR as any).render({ id, initialPage })
+  let result = setup({
+    // @ts-expect-error
+    el,
+    App: !isServer ? App : SSR,
+    props: {
+      id,
+      initialPage,
+      resolveComponent,
+    },
+  })
 
-  return {
-    body: html,
-    head: [head, `<style data-vite-css>${css.code}</style>`],
+  if (isServer) {
+    if (!result && typeof (SSR as SSRComponentType).render !== 'function') {
+      throw new Error(`setup(...) must return rendered result when in SSR mode.`)
+    } else if (!result) {
+      console.warn('Deprecated: in SSR mode setup(...) must be defined and must return rendered result in SSR mode.')
+      console.warn('For Svelte 5: `return render(App, { props })` or for Svelte 4: `return App.render(props)`')
+      result ||= (SSR as SSRComponentType).render!({ id, initialPage })
+    }
+
+    const { html, head, css } = result
+
+    return {
+      body: html,
+      head: [
+        head,
+        // Note: Svelte 5 no longer output CSS
+        ...(css?.code ? [`<style data-vite-css>${css?.code}</style>`] : []),
+      ],
+    }
   }
 }
--- a/packages/svelte/src/lib/createInertiaApp.ts
+++ b/packages/svelte/src/lib/createInertiaApp.ts
@@ -86,7 +86,7 @@ export default async function createInertiaApp({
     if (!result && typeof (SSR as SSRComponentType).render !== 'function') {
       throw new Error(`setup(...) must return rendered result when in SSR mode.`)
     } else if (!result) {
-      console.warn('Deprecated: in SSR mode setup(...) must be defined and must return rendered result in SSR mode.')
+      console.warn('Deprecated: setup(...) must be defined and must return rendered result in SSR mode.')
       console.warn('For Svelte 5: `return render(App, { props })` or for Svelte 4: `return App.render(props)`')
       result ||= (SSR as SSRComponentType).render!({ id, initialPage })
     }
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,55 +29,55 @@
       }
     },
     "node_modules/@ampproject/remapping": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.1.tgz",
-      "integrity": "sha512-lFMjJTrFL3j7L9yBxwYfCq2k6qqwHyzuUl/XBnif78PWTJYyL/dfowQHWE3sp6U6ZzqWiiIZnpTMO96zhkjwtg==",
+      "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
+      "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
       "dependencies": {
-        "@jridgewell/gen-mapping": "^0.3.0",
-        "@jridgewell/trace-mapping": "^0.3.9"
+        "@jridgewell/gen-mapping": "^0.3.5",
+        "@jridgewell/trace-mapping": "^0.3.24"
       },
       "engines": {
         "node": ">=6.0.0"
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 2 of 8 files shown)


harryqt:
Can you provide a full example of app.js and ssr.js?

punyflash:
I think it would be more reasonable to make it in a same way as Vue and React adapter works: just expose `SSR` component instead of `App` in setup function and let user return rendered content:

```ts
createServer(page => createInertiaApp({
    page,
    resolve: name => resolvePageComponent(`./pages/${name}.svelte`, import.meta.glob('./pages/**/*.svelte')),
    setup: ({App, props}) => render(App, { props })
}))
```

jamesst20:
> I think it would be more reasonable to make it in a same way as Vue and React adapters works: just expose `SSR` component instead of `App` in setup function in case app is server-rendered and let user return rendered content:
> 
> ```ts
> createServer(page => createInertiaApp({
>     page,
>     resolve: name => resolvePageComponent(`./pages/${name}.svelte`, import.meta.glob('./pages/**/*.svelte')),
>     setup: ({App, props}) => render(App, { props })
>     // // Svelte 4
>     // setup: ({App, props}) => App.render(props)
> }))
> ```

It wouldn't make sense to make the API exactly like React / Vue because then we should define a `render` property that would take care of the rendering and the setup would serve only one purpose and it would be to return `App` which `createInertiaApp` already own. Also, the render signature is different between Svelte 4 and 5 and the one from v4 comes from the component itself `App.render`.

However, I 100% agree this could be simplified to `setup(...)` only definition. I find it odd to have the same method in SSR/Non-SSR mode that servers 2 different purposes but it seems it's the way it currently is for every adapter so I applied your recommendations.

@reinink 

Before I go ahead and update all of my PRs, getting this one merged or approved prior would be great! :)
Also, the build is now failing with this error

```
[vite-plugin-svelte] [plugin vite-plugin-svelte] structuredClone is not defined
```

I guess the pipeline is simply too old and this is already addressed by https://github.com/inertiajs/inertia/pull/1874

packages/svelte/src/lib/createInertiaApp.ts
```diff
@@ -67,24 +69,37 @@ export default async function createInertiaApp({
...
+      id,
+      initialPage,
+      resolveComponent,
+    },
+  })
 
-    return
-  }
+  if (isServer) {
+    if (!result && typeof (SSR as SSRComponentType).render !== 'function') {
```
jrmajor, line 86: I think you can make a cleaner check using `import { VERSION } from 'svelte/compiler'`.

packages/svelte/src/lib/createInertiaApp.ts
```diff
@@ -67,24 +69,37 @@ export default async function createInertiaApp({
...
+      throw new Error(`setup(...) must return rendered result when in SSR mode.`)
+    } else if (!result) {
+      console.warn('Deprecated: in SSR mode setup(...) must be defined and must return rendered result in SSR mode.')
+      console.warn('For Svelte 5: `return render(App, { props })` or for Svelte 4: `return App.render(props)`')
+      result ||= (SSR as SSRComponentType).render!({ id, initialPage })
```
jrmajor, line 91:
```suggestion
      result = (SSR as SSRComponentType).render!({ id, initialPage })
```

packages/svelte/src/lib/createInertiaApp.ts
```diff
@@ -67,24 +69,37 @@ export default async function createInertiaApp({
...
+    },
+  })
 
-    return
-  }
+  if (isServer) {
+    if (!result && typeof (SSR as SSRComponentType).render !== 'function') {
+      throw new Error(`setup(...) must return rendered result when in SSR mode.`)
+    } else if (!result) {
+      console.warn('Deprecated: in SSR mode setup(...) must be defined and must return rendered result in SSR mode.')
```
jrmajor, line 89:
```suggestion
      console.warn('Deprecated: setup(...) must be defined and must return rendered result in SSR mode.')
```

packages/svelte/src/lib/createInertiaApp.ts
jamesst20: Thanks for the follow up, I will need to double check that but I believe I had tried and this import is either not available il all svelte major versions or only available in the NodeJS context while compiling/transpiling with Vite. Also the import is going to fetch for Inertia's version and not the user specified version in his project.

At first I wanted to make this "transparent" to the user and dynamically render the proper way a component based on that version but I was unsuccessful 

https://stackoverflow.com/questions/78470116/are-dynamic-imports-at-runtime-in-the-browser-esm-possible-with-svelte

Sillyvan:
With 2.0 announced and not seeing a svelte tab in the presentation im wondering what the plan here is?

pedroborges:
@Sillyvan I'm back maintaining the Svelte adapter. As soon as a couple open PRs are closed I'll be able focus my attention on bringing all the features planned for v2 to the Svelte adapter. That's the plan 👊

Sillyvan:
@pedroborges very nice to hear. Is the plan still staying on stores?
From what we heard so far Svelte 6 will be the one that might remove compatibility.

So will the adapter just be updated when that is around the corner? 
Maintainability wise this makes absolute sense we sadly miss out on some performance and some goodies but better than nothing

Thanks for your work

pedroborges:
@Sillyvan The plan for V2 is to support both Svelte 4 and 5, so no runes for now. `$page` will continue to be a store but recently I investigated https://github.com/inertiajs/inertia/issues/1670 and found out that this issue is cause by the use of an internal store to hold component name and props that are passed down to layouts and page components. This will be fixed in V2.

I'm super exited about Svelte 5 and we will definitely use on future versions. It will make the Svelte adapter code much simpler for sure 🙌

jamesst20:
It appears to me the project Inertia is dead and there are no more reasons for me to keep my work open and maintain it.

I am closing all my opened PRs which includes https://github.com/inertiajs/inertia/pull/1864 https://github.com/inertiajs/inertia/pull/1872 https://github.com/inertiajs/inertia/pull/1873 https://github.com/inertiajs/inertia/pull/1874 https://github.com/inertiajs/inertia/pull/1875 https://github.com/inertiajs/inertia/pull/1881

I will be maintaining unofficially my own fork of this library so my projects using Svelte 5 can still live until I find or create an appropriate replacement for this library.

Thanks to everyone who spent time looking at my PRs.

Sillyvan:
I totally get his frustration. The Svelte 5 PR ignored, how does he have 3 PRs fixing/improving open for 3+ months with 0 activity on it. He obviously cares a lot about this project

incon:
@jamesst20 it's a shame seeing you giving up. In general this package doesn't seem very friendly to outside contributes.

reinink:
> It appears to me the project Inertia is dead and there are no more reasons for me to keep my work open and maintain it.
> 
> I am closing all my opened PRs which includes #1864 #1872 #1873 #1874 #1875 #1881
> 
> I will be maintaining unofficially my own fork of this library so my projects using Svelte 5 can still live until I find or create an appropriate replacement for this library.
> 
> Thanks to everyone who spent time looking at my PRs.

Bummer dude 😕

We've been working *really* hard on Inertia.js v2.0 (as you can see in [Taylor's Laracon keynote](https://www.youtube.com/watch?v=AwWepVU5uWM)), so the project is far from dead — we're just in a bit of an awkward position between major versions.

I apologize that your work has not received more attention, that's partly my fault for not having enough time to review PRs recently. I don't work on Inertia.js full time like other open source maintainers (ie. Laravel, Livewire), so for me it actually requires time out of my personal life. Because of that things just take longer, as much as I wish it didn't.

Thanks for your contributions either way.

harryqt:
> We've been working _really_ hard on Inertia.js v2.0

Wow! The features are wonderful. Can't wait to test.

However I don't see any Svelte example, I guess you don't use it. Perhaps you can separate adapter packages to mono repo, and give commit permissions to others. That way community can maintain svelte adapter.

pedroborges:
@jamesst20, I totally get where you're coming from, and I'm really sorry for the frustration you've been feeling. We can definitely do better here. I haven't been as involved as I wished. I was away from the project for the last two years, but I'm back now and really grateful for everyone who's kept contributing in the meantime 🙇

The team has been heads down working on Inertia.js v2, especially leading up to the Laracon announcement last week. There are a lot of cool new features coming, and I'm excited to be more involved again!

Just last Friday I mentioned internally that getting #1881 and #1872 reviewed and approved was my top priority this week since I'm blocked by them too. I'd really appreciate it if you could reopen these two PRs. After that, I'll be focusing on fixing #1670 and adding all the new v2 stuff to the Svelte adapter so we can launch it with the other adapters next month.

Thanks again for all your hard work and contributions. It really means a lot!

jamesst20:
> @jamesst20, I totally get where you're coming from, and I'm really sorry for the frustration you've been feeling. We can definitely do better here. I haven't been as involved as I wished. I was away from the project for the last two years, but I'm back now and really grateful for everyone who's kept contributing in the meantime 🙇
> 
> The team has been heads down working on Inertia.js v2, especially leading up to the Laracon announcement last week. There are a lot of cool new features coming, and I'm excited to be more involved again!
> 
> Just last Friday I mentioned internally that getting #1881 and #1872 reviewed and approved was my top priority this week since I'm blocked by them too. I'd really appreciate it if you could reopen these two PRs. After that, I'll be focusing on fixing #1670 and adding all the new v2 stuff to the Svelte adapter so we can launch it with the other adapters next month.
> 
> Thanks again for all your hard work and contributions. It really means a lot!

I'm glad you are back ! :) I might re-contribute on the project later if I notice PRs ever starts to get merged again :)

Perhaps you could take a look at this issue? https://github.com/inertiajs/inertia/issues/775#issuecomment-2103680326

It's likely to affect more than Svelte 5 projects but short summary, when there are proxy in your variables (deep or not) `structuedClone` crashes and you can no longer page navigate. I'm still using the global workaround in my projects but it is ugly

clementmas:
Svelte 5 is officially out now and looks very promising. Is there any progress being made now to support it?

jamesst20:
> Svelte 5 is officially out now and looks very promising. Is there any progress being made now to support it?

@pedroborges Reopened a new pr and
Svelte 5 support was added by https://github.com/inertiajs/inertia/pull/1970

clementmas:
Ah ok, it's available with the beta version (1.3 or 2.0). Thanks

PR #1874, opened 2024-05-12
Upgrade: All NPM packages, depedencies, laravel, workflow

jamesst20:
**Upgrades every NPM packages for** 

- Core
- All adapters
- All playgrounds
- Vue 2 integration tests (Upgrade from Cypress 6 to 13 + very minor fixes)
- React: Minor typing improvements to `useForm` to allow compilation

**Upgrade Laravel**

- All playgrounds have been upgraded to Laravel 11.
- Fresh project from scratch: `composer create-project laravel/laravel`
- Installed latest version of Laravel Inertia
- Setup the same Inertia middleware
- Setup same routes in web.php
- Reused same Tailwind/PostCSS/(upgraded)packages.json
- Reused same exact `Resources` folder (Views, JS, CSS, ...)
- Used tsconfig.json from fresh project using vite template i.e `npm create vite@latest  myapp  -- --template vue-ts`

**Upgrade github workflow**

- Upgraded to NodeJS 20 LTS. Now possible with everything up to date

Capture d’écran, le 2024-05-12 à 00 18 15

PR #1875, opened 2024-05-12
Migrate from npm to pnpm

jamesst20:
Note: This is based from this branch https://github.com/inertiajs/inertia/pull/1874
This is just one small commit: https://github.com/inertiajs/inertia/pull/1875/commits/7dd0a030aab44f05f5186c3eccee517bebed716b

Replace npm by pnpm.

Pnpm is a much better package manager. If you are curious about its differences, I recommend you to google it.
Syntax and usage is pretty much identical

```
pnpm add -D package-name # npm install --save-dev package-name
pnpm install # pnpm install

# Build all
pnpm build:all
# or
pnpm --workspace-concurrency 1 -r build


# Publish all
pnpm -r publish
```

The reason why it's a good idea is because pnpm properly scopes packages unlike npm. As you can see in the single commit of this branch, I had to add missing packages to few repos. This should have never worked in the first place. This prevent weird bugs as well where the wrong package version is used for building as well.

Also, using @inertiajs/core from the workspace would be a better approach because sometimes both adapter and core need an update. i.e adding a new type that would be missing from the core in the currently published version.

Pnpm automatically re-alias properly packages referenced from the workspace.

https://pnpm.io/workspaces 
```
Referencing workspace packages through aliases
Let's say you have a package in the workspace named foo. Usually, you would reference it as "foo": "workspace:*".

If you want to use a different alias, the following syntax will work too: "bar": "workspace:foo@*".

Before publish, aliases are converted to regular aliased dependencies. The above example will become: "bar": "npm:foo@1.0.0".
```


jamesst20:
It appears to me the project Inertia is dead and there are no more reasons for me to keep my work open and maintain it.

I am closing all my opened PRs which includes https://github.com/inertiajs/inertia/pull/1864 https://github.com/inertiajs/inertia/pull/1872 https://github.com/inertiajs/inertia/pull/1873 https://github.com/inertiajs/inertia/pull/1874 https://github.com/inertiajs/inertia/pull/1875 https://github.com/inertiajs/inertia/pull/1881

I will be maintaining unofficially my own fork of this library so my projects using Svelte 5 can still live until I find or create an appropriate replacement for this library.

Thanks to everyone who spent time looking at my PRs.

PR #1896, opened 2024-06-14
[React] Fix wrong type in transform

rejin-r:
Simple fix for a wrong type for the transform function in the react package.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -18,7 +18,7 @@ export interface InertiaFormProps<TForm extends FormDataType> {
   wasSuccessful: boolean
   recentlySuccessful: boolean
   setData: setDataByObject<TForm> & setDataByMethod<TForm> & setDataByKeyValuePair<TForm>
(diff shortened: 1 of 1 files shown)


packages/react/src/useForm.ts
```diff
@@ -18,7 +18,7 @@ export interface InertiaFormProps<TForm extends FormDataType> {
   wasSuccessful: boolean
   recentlySuccessful: boolean
   setData: setDataByObject<TForm> & setDataByMethod<TForm> & setDataByKeyValuePair<TForm>
-  transform: (callback: (data: TForm) => TForm) => void
+  transform: (callback: (data: TForm) => object) => void
```
rejin-r, line 21: The return value cannot be of type `TForm` as that defeats the purpose of transform. It should be anything but TForm. 😄

thecrypticace (approved):

reinink:
Thanks! 🙏

theblindfrog:
@reinink is it possible to tag a release for this? I know y'all are likely flat out on 2.0; but this'd be great in a 1.2.1

PR #1901, opened 2024-06-21
Introduce `formDataArrayFormat` option

skryukov:
This PR adds a new `formDataArrayFormat` option to `Link` and `router.visit` that duplicates the logic of `queryStringArrayFormat`, but for formData arrays serialization.

### Rationale

Currently, multipart form data arrays are always serialized using indices:

```
------WebKitFormBoundarynVnO6UtO8ts1Qa5q
Content-Disposition: form-data; name="model[files][0]"; filename="aliens.png"
Content-Type: image/png


------WebKitFormBoundarynVnO6UtO8ts1Qa5q
Content-Disposition: form-data; name="model[files][1]"; filename="ufo.jpg"
Content-Type: image/jpeg
```

Rails (or more precisely [Rack](https://github.com/rack/rack/blob/main/lib/rack/query_parser.rb#L116-L122)) parses this as hashes with keys `'0'` and `'1'`. To support scenarios like a file input with the `multiple` attribute, we need to allow the tweaking of this behavior.

This PR introduces the `formDataArrayFormat` option to `router.visit`. With this option set to `'brackets'`, we serialize the keys in the format required by Rack/Rails:

```js
  form.post(`/models`, {
    formDataArrayFormat: 'brackets',
  })
```

```
------WebKitFormBoundaryxrZa60F5B7gxV5aG
Content-Disposition: form-data; name="model[files][]"; filename="aliens.png"
Content-Type: image/png


------WebKitFormBoundaryxrZa60F5B7gxV5aG
Content-Disposition: form-data; name="model[files][]"; filename="ufo.jpg"
Content-Type: image/jpeg
```

### Notes

I have set `formDataArrayFormat` to `'indices'` by default to preserve the current behavior. We might want to synchronize `formDataArrayFormat` and `queryStringArrayFormat` values at some point (or even introduce more general option), but this could be a breaking change.

### DIFF
--- a/packages/core/src/formData.ts
+++ b/packages/core/src/formData.ts
@@ -1,28 +1,47 @@
-import { FormDataConvertible } from './types'
+import { FormDataConvertible, SerializationArrayFormat } from './types'
 
 export function objectToFormData(
   source: Record<string, FormDataConvertible>,
-  form: FormData = new FormData(),
-  parentKey: string | null = null,
+  form: FormData,
+  parentKey: string | null,
+  formDataArrayFormat: SerializationArrayFormat,
 ): FormData {
   source = source || {}
 
   for (const key in source) {
     if (Object.prototype.hasOwnProperty.call(source, key)) {
-      append(form, composeKey(parentKey, key), source[key])
+      append(form, composeObjectKey(parentKey, key), source[key], formDataArrayFormat)
     }
   }
 
   return form
 }
 
-function composeKey(parent: string | null, key: string): string {
-  return parent ? parent + '[' + key + ']' : key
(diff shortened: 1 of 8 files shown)


skryukov:
CI fails in master as well, see #1902 for a possible fix.

skryukov:
Since 2.0 is coming, what if we deprecate `queryStringArrayFormat` and instead of `formDataArrayFormat` introduce just `arrayFormat` or `arraySerializationFormat` that controls both formData and query serialization?

skryukov:
@pascalbaljet, may I ask you for guidance on this one? If we consider this a bug, we can also just use `queryStringArrayFormat` instead of `formDataArrayFormat`.

skryukov:
Closing in favor of #2690

PR #1902, opened 2024-06-21
Bump `vue-tsc` and update `package-lock.json`

skryukov:
This PR updates `vue-tsc` to fix the `Search string not found: "for (const existingRoot of buildInfoVersionMap.roots) {"` error. Failed CI action can be found here: https://github.com/inertiajs/inertia/actions/runs/9617696317/job/26530006801.

I also noticed that the `package-lock.json` wasn't synced with the `package.json` ([see CI logs](https://github.com/inertiajs/inertia/actions/runs/9617696317/job/26530006801#step:5:37)). To prevent such errors, I have updated GitHub Actions to use [npm ci](https://docs.npmjs.com/cli/v10/commands/npm-ci), which fails under such circumstances. We can revert this change if it seems to be unnecessary.


thecrypticace:
I think we want to use `npm ci` but we'll need to change our node_module caching strategy before we do that as it deletes the cached node_modules before installing so I've reverted that bit for now.

thecrypticace:
Thanks! ✨

PR #1903, opened 2024-06-22
[1.x] Update .gitattributes to exclude playgrounds/ from linguist inference

EmmanuelOga:
I realized the project is being classified as php based, but I think it would be more acurate to say it is based on javascript?

This is useful when people search for projects by language, if I'm looking for a javascript project on github, it would right now be excluded from search results, as it is being tagged as mainly php based.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 # exclude playgrounds/ since otherwise the project gets classified as mainly php based.
 # https://github.com/github-linguist/linguist/blob/master/docs/overrides.md#summary
-playgrounds/ linguist-vendored
+playgrounds/** linguist-vendored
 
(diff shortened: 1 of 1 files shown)


derrickreimer (approved):
Seems like a good idea to me!

driesvints (approved):

pedroborges (commented):
I'm not familiar with Linguist but from reading [the docs](https://github.com/github-linguist/linguist/blob/main/docs/overrides.md#vendored-code) it looks like the correct way to override to all files and directories in the directory would be `playgrounds/**`.

joetannenbaum (approved):
Updated based on docs and @pedroborges suggestion. Thank you!

Issue #1679, opened 2023-09-16
Link component does not work if the user presses middle click


Temepest74:
### Version:

- `@inertiajs/vue3` version: 1.0.11/3.3.4

### Describe the problem:

Inertia <Link /> does not open a new tab when i use the mouse middle click

Issue #1329, opened 2022-10-27
Middle mouse button click does not open link in new tab when using Safari


TilW-CEBEX:
### Versions:

- `@inertiajs/inertia` version: 0.11.1
- `@inertiajs/inertia-react` version: 0.8.1
- Safari version 16.0

### Describe the problem:

Clicking an InertiaLink with the middle mouse button opens the page in the same tab.
I would expect it to behave like a normal a tag, which opens the page in a new tab when using the middle mouse button.

### Steps to reproduce:

Open Safari and click on a link which uses the Link component from Inertia using the middle mouse button.

reinink:
Hey! Thanks so much for your interest in Inertia.js and for sharing this issue/suggestion.

In an attempt to get on top of the issues and pull requests on this project I am going through all the older issues and PRs and closing them, as there's a decent chance that they have since been resolved or are simply not relevant any longer. My hope is that with a "clean slate" me and the other project maintainers will be able to better keep on top of issues and PRs moving forward.

Of course there's a chance that this issue is still relevant, and if that's the case feel free to simply submit a new issue. The only thing I ask is that you please include a super minimal reproduction of the issue as a Git repo. This makes it much easier for us to reproduce things on our end and ultimately fix it.

Really not trying to be dismissive here, I just need to find a way to get this project back into a state that I am able to maintain it. Hope that makes sense! ❤️

PR #1908, opened 2024-06-24
Skip intercepting non-left button clicks on links

### LINKED ISSUE #1772: Middle mouse button click does not open link in new tab when using Safari
### Version:

- `@inertiajs/react` version: 1.0.14

### Describe the problem:

Clicking an InertiaLink with the middle mouse button opens the page in the same tab.
I would expect it to behave like a normal a link, which opens the page in a new tab when using the middle mouse button.

### Steps to reproduce:

Open https://inertiajs.com in Safari and click on a link using the middle mouse button. Result: It opens in the current tab, not a new tab.
Open https://inertiajs.com in Chrome and click on a link using the middle mouse button. Result: It opens in a new tab.

derrickreimer:
Fixes #1772 #1329 #1679

All major browsers except Safari trigger the `auxclick` instead of `click` when the middle button is clicked on a mouse. WebKit has a [long-standing bug](https://bugs.webkit.org/show_bug.cgi?id=22382) that does not adhere to this behavior, so Safari dispatches a plain old `click` event instead. But alas, we can key off of the [`button` property on `MouseEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button#value) to detect if this is an auxiliary click, and skip intercepting in that case.

I had to tweak the TypeScript types here to handle receiving a `MouseEvent` and discriminate accordingly.

### DIFF
--- a/packages/core/src/shouldIntercept.ts
+++ b/packages/core/src/shouldIntercept.ts
@@ -8,9 +8,6 @@ export default function shouldIntercept(event: MouseEvent | KeyboardEvent): bool
     (isLink && event.ctrlKey) ||
     (isLink && event.metaKey) ||
     (isLink && event.shiftKey) ||
-    // The auxiliary button (middle button) was pressed. All major browsers
-    // except Safari trigger the `auxclick` instead of `click`, so this is to
-    // ensure we do not intercept middle clicks in Safari.
-    (isLink && 'button' in event && event.button === 1)
+    (isLink && 'button' in event && event.button !== 0)
   )
 }
--- a/packages/core/src/shouldIntercept.ts
+++ b/packages/core/src/shouldIntercept.ts
@@ -1,4 +1,4 @@
-export default function shouldIntercept(event: KeyboardEvent): boolean {
+export default function shouldIntercept(event: MouseEvent | KeyboardEvent): boolean {
   const isLink = (event.currentTarget as HTMLElement).tagName.toLowerCase() === 'a'
   return !(
     (event.target && (event?.target as HTMLElement).isContentEditable) ||
@@ -7,6 +7,10 @@ export default function shouldIntercept(event: KeyboardEvent): boolean {
     (isLink && event.altKey) ||
     (isLink && event.ctrlKey) ||
     (isLink && event.metaKey) ||
-    (isLink && event.shiftKey)
+    (isLink && event.shiftKey) ||
+    // The auxiliary button (middle button) was pressed. All major browsers
+    // except Safari trigger the `auxclick` instead of `click`, so this is to
+    // ensure we do not intercept middle clicks in Safari.
+    (isLink && 'button' in event && event.button === 1)
   )
 }


packages/core/src/shouldIntercept.ts
```diff
@@ -7,6 +7,10 @@ export default function shouldIntercept(event: KeyboardEvent): boolean {
     (isLink && event.altKey) ||
     (isLink && event.ctrlKey) ||
     (isLink && event.metaKey) ||
-    (isLink && event.shiftKey)
+    (isLink && event.shiftKey) ||
+    // The auxiliary button (middle button) was pressed. All major browsers
+    // except Safari trigger the `auxclick` instead of `click`, so this is to
+    // ensure we do not intercept middle clicks in Safari.
+    (isLink && 'button' in event && event.button === 1)
```
thecrypticace, line 14: Should this be `event.button !== 0` instead? (as in "is not left click")

derrickreimer: That makes sense! I was erring on the side of caution here by surgically targeting the known edge case, but realistically there's probably not much risk expanding to `!== 0`.

thecrypticace (approved):

reinink:
Thanks @derrickreimer! 🙌

shengslogar:
Nice! I'm a bit surprised to see #1679 and this fix applied to both the Vue and React wrappers. I'm using `@inertiajs/vue3` `1.1.0` in my own project and just confirmed middle click is working prior to this patch. 🧐

In fact, native `<a>` tags in both React ([demo](https://reactplayground.vercel.app/#N4IgLgziBcBmCGAbCBTANCAbrK1QEsA7AExQA8A6AK1xHwFsAHAewCcwACAQUcY9lbN6HAOQUA9D0bUIZEQB1CDFuw4AlFPADGnAUNGtNOhUqZtOwDlsPwwKNc2acAvv0HCRNnQFpiQ8VqI+CiEYCaKWsyEEJyRofBEKKwcALwcfloArvQhYBQA5ihgAKKIKDmhAEIAngCSxAAUno5hAJQRUTEcgk6pVjZ2Dk4NcWAJhEnthD15hiRJDYoc6kZ51pp2peW5i4TLyxraeQDKYKz4OgCyzKRoS-uEmYiId3v7hzoU67YoWxVgDSkU2WU1aIAwUhkZBgdDMqg+YDQHEsmVQpx+HFceg8XjCig60U4Uj6DVaqQAfMj7rBMoQdPgohwABbwEhlADCQS0AGsGihMLkycB7st+bkKIxDALQgARFAIJ4A1oAbhFHCQSQBAANAhduShiBwACTAMWhCgAI0yYDAUWcWpV92c93uhjAmVYe12+w4AB54MzDLAUvIQABiUMcKKcvUpYAstkoGM85zktXLZPcjiXFBq33ieBpt5TZ2ERTkFScUgKxBE3ggZzOIA)) and Vue.js ([demo](https://play.vuejs.org/#eNp9kU9PAjEQxb/KpJoICVkOesKVqISDJv6JeuyBtcxCods27RRJNvvdbbsBORhu7XtvOr/ptOzB2mIXkE1Y6YWTlsAjBTvlWjbWOIIWHNbQQe1MA1cxesU113XQgqTRsK70UuFMSbEdtN+ByOhuCC3XAJVCR4OFSB4u4fJgL4a3XHdcl+O+Y+wVL4SNVRVhvAGUFdznusI63KGmO85OOnEG60gVxQvOplmCFyzHVSwuxycvsREjL4yu5arYeKPjmBmNM2EaKyPgm01jeM4mPXTyKqXMz3PWyAUcHXSxRrH9R9/4fdI4e3fo0e2Qs6NHlVsh9fb88xX38Xw0G7MMKqbPmB/ojQqJsY89hvQL7iSXaZ/ysqReffn5nlD7w1AJNCW7nOcsLnB2ZvQ/3OviJtfFVbHuFz58uYM=)) don't even trigger a `click` event when middle clicked, defaulting to the browser opening the `href` in a new tab. So now I'm not quite sure how this was an issue to begin with?

derrickreimer:
Interesting! I am seeing the `onclick` handler being triggered by middle click in both the React & Vue demos in Safari (17.5).

CleanShot 2024-06-25 at 19 34 59

I'm not really sure if the way click triggering has changed in Vue's internals over time, but at any rate, this fix should cover off Inertia intercepting any aux clicks that may make there way into the `onclick` handler.

shengslogar:
After some debugging, it turns out the original problem comes down to the `event.which > 1` statement of this line, and is specific to the React wrapper only:

https://github.com/inertiajs/inertia/blob/7526daa664ab23dcb75fda410a76721043c9798d/packages/core/src/shouldIntercept.ts#L6

For regular mouse events, `event.which === 2` for middle clicks, and `1` for left clicks, both in Vue and React.

However, React events are "synthetic events" and don't provide the `which` property, so it's always undefined (https://react.dev/reference/react-dom/components/common#react-event-object-properties). (That also means the TypeScript type here of `MouseEvent | KeyboardEvent` as-is, is technically incorrect for the React wrapper.)

Screenshot of console-logged React event, with "nativeEvent" and "which: 2" highlighted
[Updated React demo](https://reactplayground.vercel.app/#N4IgLgziBcBmCGAbCBTANCAbrK1QEsA7AExQA8A6AK1xHwFsAHAewCcwACAQUcY9lbN6HAOQUA9D0bUIZEQB1CDFuw4AlFPADGnAUNGtNOhUqZtOwDlsPwwKNc2acAvv0HCRNnQFpiQ8VqI+CiEYCaKWsyEEJyRofBEKKwcALwcfloArvQhYBQA5ihgAKKIKDmhAEIAngCSxAAUno5hAJQRUTEcgk6pVjZ2Dk4NcWAJhEnthD15hiRJDYoc6kZ51pp2peW5i4TLyxraeQDKYKz4OgCyzKRoS-uEmYiId3v7hzoU67YoWxVgDSkU2WU1aIAwUhkZBgdDMqg+YDQHEsmVQpx+HFceg8XjCig60U4Uj6DVaqQAfMj7rBMoQdPgohwABbwEhlADCQS0AGsGihMLkycB7st+bkKIxDALQgARFAIJ4A1oAbhFVk6zDKFEQzHyfOlYBVaqQSQBAAMAO5Mi5M6AcAAkwDFoQoVptziRhFs+AFxQNrutWltDqd-q9YB9v39bqDzjNRreznu90MYEyrD2u32HAAPPBmYZYCl5CAAMQljhRTkXbkpYAstkoas85zktXLZvcjiXFBqnPieBtt5TJOERTkFScUgKxBE3ggZzOIA), [updated Vue demo](https://play.vuejs.org/#eNp9UU1LAzEQ/StDFGyhbIV6qmtRSw8KfqAeA3ZJp7vbZpOQZLeFZf+7k6ytPUhvmffeZN6badmDMUlTI5uy1AlbGg8OfW1mXJWV0dZDCxbX0MHa6gquSHrFFVfrWglfagVFplYS57IU2wE2qPwQWq4AMonWD5a7ohTFFC7byCWx7JbD2yARWjktMZE6/20luOMqHfdOyAMVHisjM49UAaQZ3IswKzE2ttxxduKAMyjILYEXnM0iBC+YjjNqTscnP7ER847mr8s82TitKH50zZnQlSnJ+5sJ8Rxn0z5P4DIp9e45Yt7WODrgokCx/QffuH3AOHu36NA2yNmR85nN0ff04vMV9/Q+kpVe1ZLUZ8gPpN3VwWMve6zDFuyJLrp9ikcsVf7lFnuPyh1CBaNB2UU9Z3TY+Znof3YnyU3so1PRFr8btOFPWiARyeSadT9e18bq)

If we revert this PR and update the following line to pass `event.nativeEvent`, all should work as originally intended:

https://github.com/inertiajs/inertia/blob/3a63529262288d9b3079360b67be50d7ef3b5d27/packages/react/src/Link.ts#L73

It's still unclear to me what issue was behind #1679. [`which` in Vue 3.3.4 behaves the same as described above](https://stackblitz.com/edit/vue3-vite-starter-x52t9e?file=src%2FApp.vue), and `shouldIntercept` hasn't changed in the last few years. Apart from asking the original issue owner for a repro, I would be willing to write that specific case off (e.g., perhaps they were defining `<Link as="button" />`, which would render a `<button>` tag and cause Inertia to rightfully always intercept).

shengslogar:
It is also worth mentioning that all of this is still specific to Safari, as you rightfully pointed out in your last comment. A click event, and subsequently `shouldIntercept`, is NEVER called when middle clicking links in Chrome and Firefox.

derrickreimer:
Ah nice investigative work! Yep, updating the event we pass to `shouldIntercept` makes sense to me. Seems as though the now-deprecated `which` check could be removed as well?

shengslogar:
Oh, good call! I didn't realize `which` was deprecated. In that case, your use of `button` should be a drop-in replacement. 👍 

I also like your explicit use of `!== 0`, since it's unclear to me when [some of the other options](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button#value) (browser back/forward?) might be triggered.

PR #1909, opened 2024-06-25
Allow SSR server clustering

RobertBoes:
Currently the Inertia SSR server would only run on a single thread, so on high load this might have implications, since the Node process could only take 100% on a single core/thread.

Node has the cluster module https://nodejs.org/api/cluster.html, which would make it possible to start multiple Node servers on the same port, then each request would be dealt with by each thread in a round-robin way.

With this PR it's possible to enable clustering by using;

```js
createServer(
  (page) => createInertiaApp( ... ),
  13714,
  true,
)
```

It's disabled by default. Kind of a downside of JS is that we can't use positional arguments, so you'd need to pass the port number along as well.

Massive thanks to ChristianQode who discovered the issue and asked about it on the Inertia.js Discord server:

> Hi guys! I'm using Laravel Inertia + Vue with SSR. Sometimes the site is becoming really slow. I figured out this is because of the node SSR process is over 100% of CPU usage. However, this is only a single core, there are 7 cores idle. I want to make the node process capable of using multiple CPU cores to spread the load. What is the best way to do it? Should I spawn multiple processes with the `inertia:start-ssr command`, or should I implement node clustering somehow? Does anyone have a suggestion?
> https://discord.com/channels/592327939920494592/758259460920573992/1255111006384820234


driesvints:
@RobertBoes seems the react build fails here. Not sure if that's got to do with this PR.

jamesst20:
> @driesvints Yeah, looking into why they're failing.. Tests first failed because it was using quite an old Node version which didn't support the features I used, now it fails early. So my suspicion is that the Inertia test suite doesn't work on newer node versions (though, tbh, v18 isn't _that_ new). Also not sure if it's just the React tests, the other ones are cancelled because the React one fails to build 🤔

It's likely failing because it is using npm under the hood which doesn't manage well dependencies across multiple workspaces.

There is already a PR to switch to PNPM https://github.com/inertiajs/inertia/pull/1875
and there is already a PR to update all dependencies https://github.com/inertiajs/inertia/pull/1874

harryqt:
I have the exact same issue, thank you so much for the PR, hope this gets merged soon.

PR #1913, opened 2024-07-05
Prepend title instead of appending to meta items

RobertBoes:
Currently when using SSR and the Inertia head component, the title is appended to the meta items.

Taking an example from the docs:
```html
<Head>
  <title>Your page title</title>
  <meta name="description" content="Your page description">
</Head>
```

This would render in the HTML as;
```html
<meta name="description" content="Your page description">
<title inertia>Your page title</title>
```

Since the title is somewhat of a special element the user doesn't have any control over where it's placed in the HTML. Rendering it as the first element would make more sense to me.
(edit: removed the part about SEO, as that's not of a concern)


derrickreimer:
Interesting, I've never heard of this as an SEO factor before. Is there a source on that intel?

PR #1926, opened 2024-07-19
add hidden '_method' field to data if http method is not 'get' or 'post'

YakutD:
When submitting a form using methods other than GET and POST, browser redirect errors may occur. For example, if you have a secure DELETE route and you are logged out, the browser will redirect you to the authentication page via DELETE method. Having __method_ in _data_ solves this problem. Also, I leave the option to override this property in the transform method.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -70,6 +70,14 @@ export default function useForm<TForm extends FormDataType>(
 
   const submit = useCallback(
     (method, url, options = {}) => {
+      const _data = data
+
+      method = method.toLowerCase()
+      if(method != 'get' && method != 'post'){
+        _data['_method'] = method
+        method = 'post'
+      }
+
       const _options = {
         ...options,
         onCancelToken: (token) => {
@@ -158,9 +166,9 @@ export default function useForm<TForm extends FormDataType>(
       }
 
       if (method === 'delete') {
-        router.delete(url, { ..._options, data: transform(data) })
+        router.delete(url, { ..._options, data: transform(_data) })
       } else {
-        router[method](url, transform(data), _options)
+        router[method](url, transform(_data), _options)
       }
     },
     [data, setErrors],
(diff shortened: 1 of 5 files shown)


RobertBoes:
Inertia would already convert those redirect responses to a 303 response (https://inertiajs.com/redirects#303-response-code), which would instruct the browser to perform the follow-up request using GET.

YakutD:
> Inertia would already convert those redirect responses to a 303 response (https://inertiajs.com/redirects#303-response-code), which would instruct the browser to perform the follow-up request using GET.

What version of the adapter made this possible?

RobertBoes:
That behaviour has been in the Laravel adapter since the first released version; https://github.com/inertiajs/inertia-laravel/blob/v0.1.0/src/Middleware.php#L28

YakutD:
> That behaviour has been in the Laravel adapter since the first released version; https://github.com/inertiajs/inertia-laravel/blob/v0.1.0/src/Middleware.php#L28

When I encountered this problem I was working on a project with the adapter "inertiajs/inertia-laravel": "^0.6.8". And I have to use _method in the form.

RobertBoes:
> When I encountered this problem I was working on a project with the adapter "inertiajs/inertia-laravel": "^0.6.8". And I have to use _method in the form.

Then the issue probably lies somewhere else, as Inertia would already take care of it when the middleware is applied

YakutD:
> > When I encountered this problem I was working on a project with the adapter "inertiajs/inertia-laravel": "^0.6.8". And I have to use _method in the form.
> 
> Then the issue probably lies somewhere else, as Inertia would already take care of it when the middleware is applied

Are you sure that the request even gets into this middleware if the redirect is initiated by the exception handler?

pedroborges:
I'm closing this PR since there's no issue with Inertia.js. @YakutD use [Discussions](https://github.com/inertiajs/inertia/discussions) if you still need help debugging this problem on your project.

PR #1937, opened 2024-07-28
feat: Add isEmpty computed property to useForm

swarakaka:
This PR introduces a new `isEmpty` computed property to the `useForm` hook in Inertia. 
This addition allows developers to easily check if all fields in a form are empty, 
which can be useful for form validation, conditional rendering, or other logic 
that depends on the form's state.

### DIFF
--- a/packages/vue3/src/useForm.ts
+++ b/packages/vue3/src/useForm.ts
@@ -7,6 +7,7 @@ type FormDataType = object
 
 interface InertiaFormProps<TForm extends FormDataType> {
   isDirty: boolean
+  isEmpty: boolean
   errors: Partial<Record<keyof TForm, string>>
   hasErrors: boolean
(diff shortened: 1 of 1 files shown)


pedroborges:
@swarakaka a create form starts with empty default values and you can already use the `isDirty` property to detect if current data differs from the default ones. Also, it's hard for the library to determine what the user considers empty values, this PR doesn't take empty arrays or nested objects into account, for example. Usually, validation on Inertia.js projects is handled on the backend. If you need to provide real-time validation you could use [Laravel Precognition](https://laravel.com/docs/11.x/precognition#main-content).

I'm closing this but please feel free to reopen it if you can provide a good use case for `isEmpty` that's is not possible with `isDirty` already. In that case your PR would also need to include the code for the other adapters.

Issue #1746, opened 2023-12-12
Link component as doesn't render other components


crazyhl:
### Version:

- `@inertiajs/vue3` version: 1.0.14

### Describe the problem:

Link component only render html tag component , but doesn't render other third vue component。
here has same issue [https://github.com/inertiajs/inertia/issues/1654](https://github.com/inertiajs/inertia/issues/1654) and [https://github.com/inertiajs/inertia/issues/1550](https://github.com/inertiajs/inertia/issues/1550)

I use [https://github.com/inertiajs/inertia/issues/1654#issuecomment-1689049845](https://github.com/inertiajs/inertia/issues/1654#issuecomment-1689049845) to test Vue's dynamic `<component>` tag。but Vue's dynamic `<component>` tag not has `as` attribute, it use `is` attribute。 when i use `is` it works.
```
<v-btn>abc</v-btn>
    <Component is="v-btn">abc1</Component>
    <Link href="/" as="button">Home</Link>
    <Link href="/" as="v-btn">Home1</Link>
```
<img width="257" alt="image" src="https://github.com/inertiajs/inertia/assets/5474833/cb57c1b5-be85-424b-a2dd-efe3799a0cd3">

but, `Link` is not correct rendered.
<img width="1080" alt="image" src="https://github.com/inertiajs/inertia/assets/5474833/5d20cac3-2195-4368-85dc-e58907824b83">

I also use `:as='v-btn'` or `:as='VBtn'` , it doesn't work.
 
the console has error
<img width="1718" alt="image" src="https://github.com/inertiajs/inertia/assets/5474833/cd283205-41c2-4280-838d-a3604e67ed3d">

<img width="1747" alt="image" src="https://github.com/inertiajs/inertia/assets/5474833/f3a03843-8fe0-4964-a019-e12ce14254d3">

Issue #1668, opened 2023-06-21
How do I make a custom component act as an inertia link?


jdion84:
Using React Bootstrap and it allows you to specify hrefs in the components e.g.:

```jsx
<Navbar.Brand href="/">
    {app.name}
</Navbar.Brand>
```

How would I make this `Navbar.Brand` component act like an inertia `Link`?

jessarcher:
Hi there,

Thanks for reporting the problem you are encountering, but it looks like this is a question that may be better suited for a support channel. We only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use the functionality provided by this repository, you can try one of the following channels:

* [Laracasts Forums](https://laracasts.com/discuss)
* [Laravel.io Forums](https://laravel.io/forum)
* [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
* [Discord](https://discordapp.com/invite/KxwQuKb)
* [Larachat](https://larachat.co/)
* [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

However, this issue will not be locked, and everyone is free to discuss solutions to your problem!

Issue #1654, opened 2023-08-22
Link component's as doesn't render other components


kazon-m:
### Version:

- `@inertiajs/vue3` version: 1.0.11

### Describe the problem:

When I use as prop to render a different tag, it works. But if I pass another component to as, it just shows name of the passed component. This issue doesn't exist in vue2 adapter (exactly in version 1.0.6). I can use <Link as="v-btn"> and it renders v-btn component, but in vue3 adapter it seems to be broken.

### Steps to reproduce:

```
<script setup>
import { Link } from '@inertiajs/vue3'
import MyComponent from '@/Path/To/Component/MyComponent .vue'
</script>

<template>
    <Link as="v-btn" href="/page">Button</Link>
    <Link as="v-tab" href="/page">Tab</Link>
    <Link as="my-component" href="/page">Label</Link>

    <my-component>My component</my-component>
</template>
```

**Expected result:**

```
<button type="button" class="v-btn v-theme--light v-btn--density-default v-btn--size-default"><span class="v-btn__overlay"></span><span class="v-btn__underlay"></span><span class="v-btn__content">Button</span></button>
<button type="button" class="v-btn v-theme--light v-btn--density-default v-btn--size-default v-tab" tabindex="-1" role="tab" aria-selected="false"><span class="v-btn__overlay"></span><span class="v-btn__underlay"></span><span class="v-btn__content" data-no-activator="">Tab<div class="v-tab__slider"></div></span></button>
<div class="my-comp">Label</div>
<!-- no anchors, but when clicked, it should navigate just like in vue2 adapter  -->

<div class="my-comp">My component</div>
```

**Actual result:**

```
<v-btn>Button</v-btn>
<v-tab>Tab</v-tab>
<my-component>Label</my-component>
<!-- no anchors, but when clicked, it navigates, however it looks ugly because components aren't rendered -->

<div class="my-comp">My component</div>
```

jessarcher:
I think this might be more to do with how Vue 3 works, rather than Inertia. To help debug the issue, if you replace `<Link>` with Vue's dynamic `<component>` tag, does it render the correct element?

E.g:

```vue
<script setup>
import { Link } from '@inertiajs/vue3'
import MyComponent from '@/Path/To/Component/MyComponent.vue'
</script>

<template>
    <component as="v-btn" href="/page">Button</component>
    <component as="v-tab" href="/page">Tab</component>
    <component as="my-component" href="/page">Label</component>
</template>
```

[...]

Issue #1550, opened 2023-05-12
Link component's `as` doesn't render other components


kenan-recebli:
### Version:

- `@inertiajs/vue3` version: 1.0.6

### Describe the problem:

When I use `as` prop to render a different tag, it works. But if I pass another component to `as`, it just shows name of the passed component. This issue doesn't exist in vue2 adapter (exactly in version 1.0.6). I can use `<Link as="v-btn">` and it renders `v-btn` component, but in vue3 adapter it seems to be broken.

### Steps to reproduce:

```vue
<script setup>
import { Link } from '@inertiajs/vue3'
import MyComponent from '@/Path/To/Component/MyComponent .vue'
</script>

<template>
    <Link as="v-btn" href="/page">Button</Link>
    <Link as="v-tab" href="/page">Tab</Link>
    <Link as="my-component" href="/page">Label</Link>

    <my-component>My component</my-component>
</template>
```

#### Expected result:

```html
<button type="button" class="v-btn v-theme--light v-btn--density-default v-btn--size-default"><span class="v-btn__overlay"></span><span class="v-btn__underlay"></span><span class="v-btn__content">Button</span></button>
<button type="button" class="v-btn v-theme--light v-btn--density-default v-btn--size-default v-tab" tabindex="-1" role="tab" aria-selected="false"><span class="v-btn__overlay"></span><span class="v-btn__underlay"></span><span class="v-btn__content" data-no-activator="">Tab<div class="v-tab__slider"></div></span></button>
<div class="my-comp">Label</div>
<!-- no anchors, but when clicked, it should navigate just like in vue2 adapter  -->

<div class="my-comp">My component</div>
```

#### Actual result:

```html
<v-btn>Button</v-btn>
<v-tab>Tab</v-tab>
<my-component>Label</my-component>
<!-- no anchors, but when clicked, it navigates, however it looks ugly because components aren't rendered -->

<div class="my-comp">My component</div>
```

##### Workaround 

As a workaround I'm using a regular component and navigate via router.get, and even created a component, so I just pass URL and component's name, but it's still very inconvenient.

craigrileyuk:
This isn't an issue with inertia, it's a change to how `script setup` makes the components available to the template.

As easier workaround when working with Vuetify is to use the following plugin (install it using `app.use(convertRouterToInertia)` in your app's entry file):

```js
import { computed, useAttrs, ref } from "vue";
import { router } from "@inertiajs/vue3";
[...]

PR #1940, opened 2024-08-11
Allow custom component on Link "as" prop in React and Vue3

guizoxxv:
This PR adds the ability to use a custom component in the `as` prop of the Link component. Currently it only accepts a string value. The proposed changes would allow to use the link as:

* **Vue 3**
```vue
<script setup>
import { Link } from '@inertiajs/vue3';
import CustomComponent from '../Components/CustomComponent.vue';
</script>

<template>
  <div>
    <Link href="/article" :as="CustomComponent">Custom Component Link</Link>
    <Link href="/article" as="button">Button Link</Link>
  </div>
</template>
```
* **React**

```jsx
import { Link } from '@inertiajs/react';
import CustomComponent from '../Components/CustomComponent';

const Home = () => {
  return (
    <>
        <Link href="/article">
          HTML Tag Link (button)
        </Link>
        <Link href="/article" as={CustomComponent}>
          Custom Component Link
        </Link>
    </>
  );
};

export default Home;
```

This feature was requested in #1550, #1654, #1668 and #1746.

### DIFF
--- a/packages/react/src/Link.ts
+++ b/packages/react/src/Link.ts
@@ -7,12 +7,12 @@ import {
   router,
   shouldIntercept,
 } from '@inertiajs/core'
-import { createElement, forwardRef, useCallback } from 'react'
+import React, { ElementType, createElement, forwardRef, useCallback } from 'react'
 
 const noop = () => undefined
 
 interface BaseInertiaLinkProps {
-  as?: string
+  as?: string | ElementType
   data?: Record<string, FormDataConvertible>
   href: string
   method?: Method
@@ -115,13 +115,13 @@ const Link = forwardRef<unknown, InertiaLinkProps>(
       ],
     )
 
-    as = as.toLowerCase()
+    const isAnchor = as === 'a' || as === 'A'
     method = method.toLowerCase() as Method
     const [_href, _data] = mergeDataIntoQueryString(method, href || '', data, queryStringArrayFormat)
     href = _href
     data = _data
 
-    if (as === 'a' && method !== 'get') {
+    if (isAnchor && method !== 'get') {
       console.warn(
         `Creating POST/PUT/PATCH/DELETE <a> links is discouraged as it causes "Open Link in New Tab/Window" accessibility issues.\n\nPlease specify a more appropriate element using the "as" attribute. For example:\n\n<Link href="${href}" method="${method}" as="button">...</Link>`,
       )
@@ -131,7 +131,7 @@ const Link = forwardRef<unknown, InertiaLinkProps>(
       as,
       {
         ...props,
-        ...(as === 'a' ? { href } : {}),
+        ...(isAnchor ? { href } : {}),
         ref,
         onClick: visit,
       },
--- a/packages/vue3/src/link.ts
+++ b/packages/vue3/src/link.ts
@@ -1,8 +1,8 @@
 import { mergeDataIntoQueryString, Method, PageProps, Progress, router, shouldIntercept } from '@inertiajs/core'
-import { defineComponent, DefineComponent, h, PropType } from 'vue'
+import { Component, defineComponent, DefineComponent, h, PropType } from 'vue'
 
 export interface InertiaLinkProps {
-  as?: string
+  as?: string | Component
   data?: object
   href: string
   method?: Method
@@ -29,7 +29,7 @@ const Link: InertiaLink = defineComponent({
   name: 'Link',
   props: {
     as: {
-      type: String,
+      type: [String, Object] as PropType<string | Component>,
       default: 'a',
     },
     data: {
@@ -75,11 +75,11 @@ const Link: InertiaLink = defineComponent({
   },
   setup(props, { slots, attrs }) {
     return () => {
-      const as = props.as.toLowerCase()
+      const isAnchor = props.as === 'a' || props.as === 'A'
       const method = props.method.toLowerCase() as Method
       const [href, data] = mergeDataIntoQueryString(method, props.href || '', props.data, props.queryStringArrayFormat)
 
-      if (as === 'a' && method !== 'get') {
+      if (isAnchor && method !== 'get') {
         console.warn(
           `Creating POST/PUT/PATCH/DELETE <a> links is discouraged as it causes "Open Link in New Tab/Window" accessibility issues.\n\nPlease specify a more appropriate element using the "as" attribute. For example:\n\n<Link href="${href}" method="${method}" as="button">...</Link>`,
         )
@@ -89,7 +89,7 @@ const Link: InertiaLink = defineComponent({
         props.as,
         {
           ...attrs,
-          ...(as === 'a' ? { href } : {}),
+          ...(isAnchor ? { href } : {}),
           onClick: (event) => {
             if (shouldIntercept(event)) {
               event.preventDefault()


pedroborges (changes requested):
Thanks for this PR! I tested it today and everything works smoothly except for one small accessibility issue. When a custom component passed to the `as` prop renders an anchor tag, it doesn't receive the `href` attribute, which breaks accessibility:

```vue
<Link :as="CustomAnchor" href="/" class="hover:underline">Home</Link>

<a class="hover:underline">Home</a>
```

The link still functions due to the `onclick` handler, but without the `href`, it’s not fully accessible. Could you ensure that the `href` attribute is passed when the custom component renders an anchor element?

guizoxxv:
@pedroborges Thanks for the review.

I didn't find a effective way to discover the HTML element of the custom component main node without mounting it, Something like this:

```ts
const getCustomComponentTagName = (customComponent: Component): string => {
  const app = createApp(customComponent);
  const instance = app.mount(document.createElement('div'));
  const tagName = instance.$el.tagName; // single root component example
  app.unmount();

  return tagName;
}
```

I'm not sure if this temporary rendering would be viable considering performance/resources. In that case, do you know an altertive way?

pedroborges:
@guizoxxv Yeah, this solution is not good for performance. Here’s an alternative approach that should be more performant:

- Always pass `href` to the custom components.
- Store the rendered component in a `ref` to access the DOM element once it's rendered.
- Use `nextTick` to wait until the component is fully rendered, then check its `tagName`.
- If the `tagName` is not `A`, simply remove the `href` attribute.

This way, you can avoid unnecessary temporary rendering and still maintain accessibility when an anchor element is used. What do you think of this approach? Let me know if you need help implementing it 😉

guizoxxv:
You mean something like this (Vue3 example)?

```ts
setup(props, { slots, attrs }) {
  const elementRef = ref(null)

  nextTick(() => {
    if (elementRef.value) {
      const element = elementRef.value.$el

      if (element.tagName !== 'A') {
        element.removeAttribute('href')
      }
    }
  })

  return () => {
    const isAnchor = props.as === 'a' || props.as === 'A'
    const isCustomComponent = typeof props.as !== 'string'
    const method = props.method.toLowerCase() as Method
    const [href, data] = mergeDataIntoQueryString(method, props.href || '', props.data, props.queryStringArrayFormat)

    if (isAnchor && method !== 'get') {
      console.warn(
        `Creating POST/PUT/PATCH/DELETE <a> links is discouraged as it causes "Open Link in New Tab/Window" accessibility issues.\n\nPlease specify a more appropriate element using the "as" attribute. For example:\n\n<Link href="${href}" method="${method}" as="button">...</Link>`,
      )
    }

    return h(
      props.as,
      {
        ...attrs,
        ...(isAnchor || isCustomComponent ? { href } : {}),
        ...(isCustomComponent ? { ref: elementRef } : {}),
        onClick: (event) => {
          if (shouldIntercept(event)) {
            event.preventDefault()

            router.visit(href, {
              data: data,
              method: method,
              replace: props.replace,
              preserveScroll: props.preserveScroll,
              preserveState: props.preserveState ?? method !== 'get',
              only: props.only,
              except: props.except,
              headers: props.headers,
              // @ts-expect-error
              onCancelToken: attrs.onCancelToken || (() => ({})),
              // @ts-expect-error
              onBefore: attrs.onBefore || (() => ({})),
              // @ts-expect-error
              onStart: attrs.onStart || (() => ({})),
              // @ts-expect-error
              onProgress: attrs.onProgress || (() => ({})),
              // @ts-expect-error
              onFinish: attrs.onFinish || (() => ({})),
              // @ts-expect-error
              onCancel: attrs.onCancel || (() => ({})),
              // @ts-expect-error
              onSuccess: attrs.onSuccess || (() => ({})),
              // @ts-expect-error
              onError: attrs.onError || (() => ({})),
            })
          }
        }, 
      },
      slots,
    )
  }
},
```

I like it. Another idea I had is adding a new prop for this scenario, for example `keep-href`:

```vue
<Link :as="CustomAnchor" href="/" keep-href>Home</Link>
```

However, this add more burden to the user, while yours don't. Also, it feels counter intuitive that without it, the `href` would be removed. I think yours is best, let me know if the implementation in Vue3 is correct and I'll try to replicate in the others.

pedroborges:
This sounds like a good plan! I’m with you on not adding extra burden on the user. Also, I think we might not even need `nextTick` here. Since we're dealing with DOM elements, `onMounted` might already do the job.

You could try moving the `nextTick` logic into `onMounted` and see if it works just as well. It should simplify things further.

guizoxxv:
Yeah, I jump bumped into another problem. In Vue 2, because the Link is a functional component, which is stateless, I cannot use lifecycle hooks.

pedroborges:
@guizoxxv Vue 2 supports lifecycle hooks in a different syntax, see the [`mounted` method](https://v2.vuejs.org/v2/api/#mounted).

guizoxxv:
@pedroborges 
My understanding is that, functional components are stateless, they do not work with reactive data and don't have access to lifecycle hooks - https://v2.vuejs.org/v2/guide/render-function#Functional-Components.

Typescript interface `FunctionalComponentOptions` (used in Vue 2 Link component) does not contain lifecycle hooks. I removed this type just for testing, added the `mounted` hook and it didn't do anything. After changing `functional` to `false` it worked.

pedroborges:
Thanks for the clarification, @guizoxxv, and my apologies for the confusion! Functional components wasn't a thing when I last worked with Vue 2, so I wasn’t fully aware of their limitations.

That being said, I also forgot to mention that we’re dropping support for Vue 2 in Inertia V2. Since your PR introduces new behavior, we'll be merging it into V2. For now, you can go ahead and make these changes just for the React and Vue 3 adapters. Once the V2 branch is live, we’ll rebase and merge your work.

Thanks again for your contribution, and I appreciate your patience!

guizoxxv:
No problem. I had more experience with Vue 2 - only started with Vue 3 last year =). It makes sense to drop support for Vue 2 since it reached end of life. I reverted my code to keep the Vue 2 file as it was.

I had a bit of a strugle in React due to the ref type, which can be a function. I ended up using this function `combinedRef`, but I'm not sure if is the best approach. Also, it requires the user to foward the ref in the custom component. Example:

```jsx
import { forwardRef, Ref } from "react"

const CustomComponent = forwardRef(({ children, ...props }, ref: Ref<HTMLButtonElement>) => {
  return (
    <button ref={ref} {...props}>
      {children}
    </button>
  )
})

export default CustomComponent
```

SuperDJ:
Would potentially close https://github.com/vuetifyjs/vuetify/issues/11573 as well.

crazyhl:
When will it be merged into the mainline?

SuperDJ:
Any updates on this? Because without this using persistent layouts requires workarounds.

Workaround (partial) example for Vuetify:
```vue
<template>
<v-btn :href="url" @click="e => routerLink(e, url)">Link</v-btn>
</template>

<script setup>

const url = 'some url';

function routerLink(e, url)
{
   e.preventDefault();
  return router.visit(url);
}
<script>
```

Without providing the href, ctrl + click for example will not work.

pascalbaljet:
Closed in favor of #2443

PR #1763, opened 2024-01-03
Fix Svelte 4 slot rendering issues

dunhamjared:
Fix #1733 
Fix #1621

# Overview

When routing between layouts, Svelte ([v4.1.0 and above](https://github.com/inertiajs/inertia/issues/1621#issuecomment-1775805874)) will not slot in the pages correctly. 

This PR implements additional logic with a key block to guarantee accurate page rendering while maintaining the persistent layouts feature.

# Notes

- I found a PR that includes this fix: https://github.com/inertiajs/inertia/pull/1705
- This issue may be related: https://github.com/inertiajs/inertia/issues/1670

Issue #1733, opened 2023-11-21
Routing between default layout and persistent layout doesn't slot in the page


kobe-ra:
### Version:

- "@inertiajs/svelte": "^1.0.14",

### Describe the problem:

I have MasterLayout that is for logged in users. Then there is GuestLayout for guests. When user logs out and is redirected to Login page, GuestLayout loads, but the Login page doesn't. I then manually refresh the browser window, everything loads normally on the Login page. After submitting login form with useForm helper, user is redirected to Home page, but again - MasterLayout is loaded and works properly, while Page itself is not present at all.
[Video](https://youtu.be/EvUrAtNy3jo)

### Steps to reproduce:

app.js
```
import "./bootstrap";
import MasterLayout from "./Layouts/MasterLayout.svelte";

import { createInertiaApp } from "@inertiajs/svelte";

createInertiaApp({
    resolve: (name) => {
        const pages = import.meta.glob("./Pages/**/*.svelte", { eager: true });
        let page = pages[`./Pages/${name}.svelte`];
        console.log(page)
        // return { ...page, layout: typeof page.layout === 'undefined' ? MasterLayout : page.layout }
        return {
            default: page.default,
            // layout: (_) => page.layout || MasterLayout,
            layout: page.layout || MasterLayout,
        };
    },
    setup({ el, App, props }) {
        // new App({ target: el });
        new App({ target: el, props });
    },
});
```

GuestLayout.svelte
```
<script>
    console.log('guest layout')
</script>

<slot />
```

Login.svelte
```
<script context="module">
    export { default as layout } from "@/Layouts/GuestLayout.svelte";
</script>

<script>
    import { useForm } from "@inertiajs/svelte";
    import Input from "@/Components/Forms/Input.svelte";

    let form = useForm({
        login: "",
        password: "",
    });

    const submit = () => {
        $form.post(route("login"));
    };

    console.log("login page");
</script>

<div class="container h-screen">
    ...
</div>
```

MasterLayout.svelte
```
<script>
    import Navigation from "@/Layouts/Navigation.svelte";

    console.log('master layout')
</script>

<div id="app">
    <Navigation>
        <slot />
    </Navigation>
</div>
```

Home.svelte
```
<script>
    console.log('home page')
</script>

<svelte:head>
    <title>{import.meta.env.VITE_APP_NAME}</title>
</svelte:head>

data data
```

Josefikos:
i'm facing same issue for a while

[...]

Issue #1621, opened 2023-07-26
Issues with Svelte 4


punyflash:
### Version:

- `@inertiajs/svelte` version: 1.0.9
- `svelte` version: 4

### Describe the problem:

When using Svelte 4, props are not being populated on inertia router visits. Also when adding svelte-eslint to the project pages are completely empty after visit.

### Steps to reproduce:

Just make a visit using inertia link or router to the route that requires any props

reinink:
Hey! I am not able to reproduce this issue. We actually have a Svelte playground included in the project, which is on Svelte 4.x, and I just tested it and it works fine:

https://github.com/inertiajs/inertia/tree/master/playgrounds/svelte

I am going to close this issue for now, but feel free to provide us with a minimal reproduction of the issue you're experiencing as a Git repo and we can have another look 👍

punyflash:
@reinink here is an example repository: https://github.com/punyflash/inertia-test.

## A bit detailed explanation
I created 2 pages: `/page/1` and `page/2` to show how it works. It seems like page is not being updated after persistent layout change

### Page 1

```svelte
<script lang="ts" context="module">
    // import LayoutInsideTag from "@/components/LayoutInsideTag.svelte";
    import LayoutOutsideTag from "@/components/LayoutOutsideTag.svelte";

    export const layout = [LayoutOutsideTag];
</script>

<script lang="ts">
    import { fade } from 'svelte/transition';
    import { inertia } from "@inertiajs/svelte";
    export let title = '';
    export let data = '';
</script>

<svelte:head>
    <title>{title}</title>
</svelte:head>

<section in:fade>
    <h1 class="text-3xl">{title}</h1>
    <div>{data}</div>
    <a href="/page/2" class="underline" use:inertia>Go to page 2</a>
</section>
```

### Page 2

```svelte
<script lang="ts" context="module">
    // import LayoutOutsideTag from "@/components/LayoutOutsideTag.svelte";
    import LayoutInsideTag from "@/components/LayoutInsideTag.svelte";

    export const layout = [LayoutInsideTag];
</script>

<script lang="ts">
    import { fade } from 'svelte/transition';
    import { inertia } from "@inertiajs/svelte";
    export let title = '';
    export let data = '';
</script>

<svelte:head>
    <title>{title}</title>
</svelte:head>

<section in:fade>
    <h1 class="text-3xl">{title}</h1>
    <div>{data}</div>
[...]

PR #1943, opened 2024-08-13
[1.x] Fix <Render /> component to respect "preserveState"

pedroborges:
Issues with the page component rendering in the Svelte adapter have been reported a couple time (see: #1621, #1733). The last attempt to solve this (#1763) created another issue because it doesn't respect `preserveState` and reuses page components even when `preserveState` is set to `false`. This PR fixes this issue.

I have updated the Svelte playground with a reproducible to help me find a solution and also help the reviewer. We can revert these changes to the playground before merging this PR.

## Before
Navigating between `/foo` and `/bar`, which use the same page component, doesn't refresh the page component, but only updates its props. The same page component is reused instead of being destroyed and a new one being mounted.

Screen Shot 2024-08-13 at 16 03 28

## After
Navigating between `/foo` and `/bar` will use a fresh instance of the page component each time, see the "mounted" and "destroyed" logs.

Screen Shot 2024-08-13 at 16 13 54

----

This PR still doesn't fully solve the issue described in #1670. Notice on the blue arrows on the "After" screenshot that in some cases the component prop is updated shortly after the component is initialized. In my tests, this only happened when I tried to use a prop inside the `<script>` block. Using props in reactive expressions and lifecycle hooks such as `onMount` works just fine.

```svelte
<script>
  import { useForm } from '@inertiajs/svelte'

  export let name

  const form = useForm({
    'name': name,
  })
</script>
```

After a lot of debugging, I found out that this happens because we are using a Svelte store to persist component and props state. When the store is updated their listeners can be slightly out-of-sync. This causes the internal `<App />` component to pass outdated props down to the component tree.

I have a working solution and will open a separate PR to fix that because there might be a small breaking change.


packages/svelte/src/lib/components/Render.svelte
```diff
@@ -23,20 +23,17 @@
...
-  {#key key}
+  <!--
+  Add the `key` only to the last (page) component in the tree.
+  This ensures that the page component re-renders when `preserveState` is disabled,
+  while the layout components are persisted across page changes. -->
+  {#key children?.length === 0 ? $store.key : null}
     {#if children.length > 0}
       <svelte:component this={component} {...props}>
-        {#each children as child, index (component && component.length === index ? $store.key : null)}
+        {#each children as child}
```
pedroborges, line 36: Since children always contains a single element, there's no need to use a key in the `#each` block in this case.

PR #1948, opened 2024-08-21
Added a prop external to give the option of making a link external or…

rajvirkolpolok:
I added a prop external to give the Link Tag an option of making it act as a link or a tradition a tag. I faced a problem building a CMS where i had routes which can be both internal or external coming from the user and i had to check it every time.

### DIFF
--- a/packages/react/src/Link.ts
+++ b/packages/react/src/Link.ts
@@ -11,13 +11,24 @@ import { createElement, forwardRef, useCallback } from 'react'
 
 const noop = () => undefined
 
+// Hook to determine if a link is internal or external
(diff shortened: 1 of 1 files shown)


pedroborges:
Hey @rajvirkolpolok, how are you verifying whether a URL is internal or external? I'm not sure we want to handle this inside Inertia, my recommendation is to check the URL yourself and only use the Inertia <Link /> component for internal ones.

rajvirkolpolok:
Hey @pedroborges, Link Tags from Next and React handle this on their own afaik. So for consistency purposes as well, I believe this is better.

PR #1957, opened 2024-09-02
Fix `setNavigationType` for Safari 10

hivokas:
Simple fix for Safari 10. 

Screenshot 2024-09-02 at 2 02 34 PM

It happens because `this.navigation` is present, but `this.navigation.getEntriesByType` is missing.

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -70,7 +70,7 @@ export class Router {
 
   protected setNavigationType(): void {
     this.navigationType =
-      window.performance && window.performance.getEntriesByType('navigation').length > 0
+      window.performance && window.performance.getEntriesByType && window.performance.getEntriesByType('navigation').length > 0
         ? (window.performance.getEntriesByType('navigation')[0] as PerformanceNavigationTiming).type
         : 'navigate'
   }


derrickreimer (approved):
It looks like Safari was one of the latest to add support for this (in 2017). https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType#browser_compatibility

Arguably this affects an extremely small % of users, but I don't really see a drawback in adding one additional guard to this ternary.

reinink:
@hivokas Hey thanks for this! We've actually already corrected this in our v2.0 branch ([see here](https://github.com/inertiajs/inertia/blob/async-requests/packages/core/src/navigationType.ts)), but I think it would be smart to fix this in 1.x as well 👍

hivokas:
> Arguably this affects an extremely small % of users, but I don't really see a drawback in adding one additional guard to this ternary.

Yeah, it's indeed an extremely small % of users, but useful for projects which aim to support a wide variety of browsers.

hivokas:
> @hivokas Hey thanks for this! We've actually already corrected this in our v2.0 branch ([see here](https://github.com/inertiajs/inertia/blob/async-requests/packages/core/src/navigationType.ts)), but I think it would be smart to fix this in 1.x as well 👍

@reinink Welcome!

By the way, I've checked the code from the link you provided: https://github.com/inertiajs/inertia/blob/338b4c25a68ff0a62c1fedf96789d0ff4273f21c/packages/core/src/navigationType.ts#L5-L9

Unless I'm missing something, I don't think `window?.performance.getEntriesByType` handles this Safari 10 edge case. Should it be `window?.performance?.getEntriesByType` instead? (because Safari has `windows.performance`, but doesn't have `window.performance.getEntriesByType`).

jamesst20:
> > @hivokas Hey thanks for this! We've actually already corrected this in our v2.0 branch ([see here](https://github.com/inertiajs/inertia/blob/async-requests/packages/core/src/navigationType.ts)), but I think it would be smart to fix this in 1.x as well 👍
> 
> @reinink Welcome!
> 
> By the way, I've checked the code from the link you provided:
> 
> https://github.com/inertiajs/inertia/blob/338b4c25a68ff0a62c1fedf96789d0ff4273f21c/packages/core/src/navigationType.ts#L5-L9
> 
> Unless I'm missing something, I don't think `window?.performance.getEntriesByType` handles this Safari 10 edge case. Should it be `window?.performance?.getEntriesByType` instead? (because Safari has `windows.performance`, but doesn't have `window.performance.getEntriesByType`).

Then in this case both will be wrong as you can't call a function if it is undefined even with the `?` operator on the previous assignment.

Proper signature would be
```
window.performance.getEntriesByType?.(...)
```

Screenshot 2024-09-09 at 11 37 59 AM

pedroborges (approved):
Looks good to me, thanks @hivokas!

joetannenbaum (approved):
Looks good, thanks!

hivokas:
Thanks!

hivokas:
@joetannenbaum even though it works in 1.x, it doesn't seem to work in 2.x

PR #1958, opened 2024-09-03
[1.x] Fix type augmentation for Vue 3

daniser:
This PR changes Vue 3 type augmentation to be placed in `vue` module instead of `@vue/runtime-core` [as it supposed to be](https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties).
Many infrastructure packages like [Vue Router](/vuejs/router/pull/2295), [Vue i18n](/intlify/vue-i18n/pull/1919) and [Nuxt](/nuxt/nuxt/pull/28542) already did so.

### DIFF
--- a/packages/vue3/src/types.ts
+++ b/packages/vue3/src/types.ts
@@ -12,7 +12,7 @@ declare module '@inertiajs/core' {
   }
 }
 
-declare module '@vue/runtime-core' {
+declare module 'vue' {
   export interface ComponentCustomProperties {
     $inertia: typeof router
     $page: Page


pedroborges (approved):
Looks good to me! I can confirm this is how [Laravel Breeze](https://github.com/laravel/breeze/blob/ac6227c0054095bb65be133c8e277df1ea925983/stubs/inertia-vue-ts/resources/js/types/global.d.ts#L14) augments Vue 3 types.

joetannenbaum (approved):
Looks good, thanks!

daniser:
@pedroborges @joetannenbaum for some reason this was merged into `v1` (yet unreleased) but not into current `master` branch. Because of this I still get random type related errors from vue-tsc.

Issue #1878, opened 2024-05-19
Make `data-page` hidration a new request


PillFall:
### Version:

- `@inertiajs/vue3` version: 0.x and 1.x

I believe this applies to another Inertia drivers as well.

### Describe the problem:

I published https://github.com/inertiajs/inertia/discussions/1566 a while ago in the discussion page with no response, so I decided to put this as an issue, due to this could be considered a bug or design flaw.

When the page load for the first time, all the data needed for rendering are written as JSON inside the data-page attribute in the root element. This is ok, I know the JSON parser is pretty fast.

But...

From what I have been able to test, if this data is relatively big (like 1000 chars), it could make the DOM parsing significantly slower, triggering bad performance, increasing the FCP in more than 3 seconds, on low-end devices.

The only thing I could attribute this is all the page data stored in the `data-page` attribute. If I load the same page without the bulked `data-page` attribute, the FCP in under 1.5 seconds at most. I used VueJS in both tests.

I propose, instead of loading the data in the data-page attribute for the first load, fetch the data as a new XHR request, as the subsequent page visits are. This wouldn't differ from what an SPA does.

I know the backend (probably) must handle the same exact request twice, but we can control the backend processing speed and power, but not the frontend. And this can be easily mitigated with a middleware.

This will also allow us to send even more information (and maybe even quicker) to the frontend to process. And we get in addition a cleaner and smaller HTML, avoiding data duplication in the DOM.

peter-emad99:
what about inline script tag with defer attribute ?

RobertBoes:
Making an additional request on load seems rather pointless to me, it would also require a rework of the Inertia protocol and is quite a significant breaking change. Inertia is inherently backend-driven, so all the data is already present on the initial request (such as authentication, authorization checks etc), only then to make another request for pretty much the same data.

One option I'd see, similar to what @peter-emad99 is suggesting; make the way initial data is resolved user-configurable. So instead of these lines https://github.com/inertiajs/inertia/blob/71792799b852a5e6cb882d74f4abd117b15c989e/packages/vue3/src/createInertiaApp.ts#L32-L33 that could be done through a callback. I'd imaging something along the lines of this:
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0" />
    @vite('resources/js/app.js')
    <script>
[...]

PR #1964, opened 2024-09-06
Allow passing initialData callback

RobertBoes:
solves #1960 and #1878 (although not through a new request)

This would allow users to customize how the initial data is rendered. Currently it is always added through the `data-page` attribute.

With this PR the following is possible;

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0" />

	<!-- Rendering the props in a script tag -->
	<script>
		window.__inertia_data__ = {{ Illuminate\Support\Js::from($page) }};
	</script>

    @vite('resources/js/app.js')
    @inertiaHead
  </head>
  <body>
	<!-- Just render a div, without the need for data-page -->
    <div id="app"></div>
  </body>
</html>
```

Then in your main script you'd do the following:
```js

createInertiaApp({
  resolve: name => {
    const pages = import.meta.glob('./Pages/**/*.vue', { eager: true })
    return pages[`./Pages/${name}.vue`]
  },
  // Pass the initialData callback, as we provide the data and Inertia doesn't have to grab it from the DOM
  initialData: () => window.__inertia_data__,
  setup({ el, App, props, plugin }) {
    createApp({ render: () => h(App, props) })
      .use(plugin)
      .mount(el)
  },
})
```

As a note: This would also require a bit of rework for SSR, as that would still output the div with the `data-page`. Haven't gotten to fixing that yet, but if there's no interest in this there's no point in researching that

### DIFF
--- a/packages/react/src/createInertiaApp.ts
+++ b/packages/react/src/createInertiaApp.ts
@@ -35,8 +35,9 @@ type BaseInertiaAppOptions = {
 type CreateInertiaAppSetupReturnType = ReactInstance | void
 type InertiaAppOptionsForCSR<SharedProps extends PageProps> = BaseInertiaAppOptions & {
   id?: string
-  page?: Page | string
+  page?: Page
   render?: undefined
+  initialData?: (el: HTMLElement) => Page
   progress?:
     | false
     | {
@@ -51,8 +52,9 @@ type InertiaAppOptionsForCSR<SharedProps extends PageProps> = BaseInertiaAppOpti
 type CreateInertiaAppSSRContent = { head: string[]; body: string }
 type InertiaAppOptionsForSSR<SharedProps extends PageProps> = BaseInertiaAppOptions & {
   id?: undefined
-  page: Page | string
+  page: Page
   render: typeof renderToString
+  initialData?: undefined
   progress?: undefined
   setup(options: SetupOptions<null, SharedProps>): ReactInstance
 }
@@ -68,6 +70,7 @@ export default async function createInertiaApp<SharedProps extends PageProps = P
   resolve,
   setup,
   title,
+  initialData = (el) => JSON.parse(el.dataset.page),
   progress = {},
   page,
   render,
@@ -76,7 +79,7 @@ export default async function createInertiaApp<SharedProps extends PageProps = P
 > {
   const isServer = typeof window === 'undefined'
   const el = isServer ? null : document.getElementById(id)
-  const initialPage = page || JSON.parse(el.dataset.page)
+  const initialPage = page || initialData(el)
   // @ts-expect-error
   const resolveComponent = (name) => Promise.resolve(resolve(name)).then((module) => module.default || module)
 
@@ -88,6 +91,7 @@ export default async function createInertiaApp<SharedProps extends PageProps = P
       el,
       App,
       props: {
+        // @ts-expect-error
         initialPage,
         initialComponent,
         resolveComponent,
--- a/packages/svelte/src/lib/createInertiaApp.ts
+++ b/packages/svelte/src/lib/createInertiaApp.ts
@@ -17,6 +17,7 @@ interface CreateInertiaAppProps {
       resolveComponent: ComponentResolver
     }
   }) => void | SvelteApp
+  initialData?: (el: HTMLElement | null) => Page
   progress?:
     | false
     | {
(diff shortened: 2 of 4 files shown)


pedroborges:
I've reviewed the discussions around this topic, and I see the benefits of allowing more flexibility in how the initial data is rendered. I'd like to propose a slight adjustment to enhance the developer experience (DX) while maintaining backward compatibility:

- **Configurable Initial Data Rendering**: Introduce a configuration option in the Laravel Inertia adapter that lets users choose between `attribute` or `script` for rendering initial data.
    - If `attribute` is selected, the JSON-encoded props would continue to be rendered in the `data-page` attribute, as is currently done, using the `@inertia` directive.
    - If `script` is selected, the props would instead be output as a JavaScript object in a `<script>` tag, leveraging the `@inertiaHead` directive.
- **Fallback Mechanism in `createInertiaApp`**: Provide a default implementation of the `initialData` option in `createInertiaApp` that first looks for data in the `<script>` tag, and falls back to the `data-page` attribute if necessary. This keeps things flexible while allowing users to override the `initialData` option if they need a custom approach.

This would offer a cleaner and more flexible setup for users who prefer handling their data in a `<script>` tag while preserving the current behavior for those who prefer the `data-page` attribute.

RobertBoes:
Not really sure I understand the proposal, you want to either load it from the data attribute or from a script (a predefined window variable?). That would be less flexible to the end-user, right?
Or do you mean to pretty much keep this implementation, along with also checking a window variable. So the end user still has the option to overwrite the method and provide their own data?

jamesst20:
Just wondering, because we can already pass the `page` attribute to `createInertiaApp` as seen here https://github.com/inertiajs/inertia/blob/master/packages/svelte/src/lib/createInertiaApp.ts#L40 

Wouldn't it be simpler to simply allow methods to be there?

```typescript
page: string | () => string | () => Promise<string>
```

Replace
```avascript
  const initialPage = page || JSON.parse(el?.dataset.page ?? '{}')
```
with
```javascript
const initialPage = typeof page === function ? await page() : (page || JSON.parse(el?.dataset.page ?? '{}'))
```
 
```

**Comment by pedroborges** [comment]
@RobertBoes as @jamesst20 mentioned it seems that the current `page` argument in `createInertiaApp` can already address this use case. My proposal was aimed at improving the developer experience (DX) by not requiring the user to manually add the `<script>` tag. Instead, I suggest we add a `inertia.pageSource` option to the Laravel adapter and leverage the existing `@inertiaHead` directive to output the script tag when this option is set to `script`.

The only change inside `createInertiaApp` on the adapters would be:

```diff
- const initialPage = page || JSON.parse(el.dataset.page)
+ const initialPage = page || window.__inertiaPage || JSON.parse(el.dataset.page)
```

This way, the library supports the two most common use cases, script and data attribute, out-of-the-box and users would still have the flexibility to override `page` if they want to.

jamesst20:
> @RobertBoes as @jamesst20 mentioned it seems that the current `page` argument in `createInertiaApp` can already address this use case. My proposal was aimed at improving the developer experience (DX) by not requiring the user to manually add the `<script>` tag. Instead, I suggest we add a `inertia.pageSource` option to the Laravel adapter and leverage the existing `@inertiaHead` directive to output the script tag when this option is set to `script`.
> 
> The only change inside `createInertiaApp` on the adapters would be:
> 
> ```diff
> - const initialPage = page || JSON.parse(el.dataset.page)
> + const initialPage = page || window.__inertiaPage || JSON.parse(el.dataset.page)
> ```
> 
> This way, the library supports the two most common use cases, script and data attribute, out-of-the-box and users would still have the flexibility to override `page` if they want to.

That would work but can we expect everyone https://inertiajs.com/community-adapters to update their implementation aswell?

I still believe supporting supporting a function/promise has its worth even if this could be done outside the createInertiaApp and passed as a prop

pedroborges:
Hey @RobertBoes, I can confirm that this functionality is already supported via the `page` option in `createInertiaApp`, which allows you to pass the initial page data directly, like so:

```javascript
createInertiaApp({
  page: window.initialPage,
  // ...
})
```

Does this approach satisfy your use case, or is there something additional you're looking to achieve with this PR? Let me know 😉

RobertBoes:
Hey @pedroborges, yeah looks like that also works. I'll go with that then.

PR #1969, opened 2024-09-10
[1.x] Fix props reactivity

### LINKED ISSUE #1670: Svelte page component instance is not refreshed

pedroborges:
This PR builds on #1943 and introduces a complete fix for issue #1670. ~It also includes a minor breaking change, so I've kept it separate for easier review.~

## The Issue

Currently, when navigating between pages that share the same page component, props can fall out of sync. For instance, navigating between user edit profile pages reuses the page component, but the props (like `name`) [aren’t updated properly](https://github.com/inertiajs/inertia/issues/1670#issuecomment-2119452838) when used inside the `<script>` block—unless they're wrapped in reactive expressions or lifecycle functions like `onMount`.

Example:

```svelte
<script>
  import { onMount } from 'svelte'
  import { useForm } from '@inertiajs/svelte'

  export let name

  const form = useForm({
    'name': name, // ⛔️ outdated
  })
  
  $: console.log(name) // ✅ updated
  
  onMount(() => console.log(name)) // ✅ updated
</script>
```

~This happens because we're using a Svelte store to persist the component and props state, and sometimes the store’s listeners get out of sync. As a result, the internal `<App />` component ends up passing outdated props down to the component tree.~

It’s possible we were using the wrong combination of the store with reactive expressions, or perhaps it was due to using the store inside both the `<App />` and `<Render />` components. However, the store itself wasn’t the issue.

## The Solution

~The fix is straightforward: instead of relying on the store, we’ll pass the props directly to the `<App />` component and update them with the `$set` method.~

~To implement this, there’s a small, but necessary, breaking change in user-land:~ I found another solution without introducing breaking changes. This solution has been tested on Svelte 4 and 5 projects.

```diff
createInertiaApp({
  // ...
- setup({ el, App }) {
+ setup({ el, App, props }) {
-   new App({ target: el, hydrate: true })
+   return new App({ target: el, props, hydrate: true })
  },
})
```

~By making this change, we guarantee that props are up-to-date when the new page components are instantiated.~

By removing the store from the `<Render />` component and refactoring the `<App />` component to use a single reactive expression, we ensure that props are always up-to-date when new page components are instantiated.

----

Fixes #1670


pedroborges:
#1970 is drafted and we intend to support Svelte 4 and 5. With Svelte 5 removing class-based components and the `$set` method, we need to adapt our prop update approach this PR introduces. Svelte now recommends using `$state` for programmatically updated props.

Two potential solutions:

1. Modified `createInertiaApp` setup:

    ```js
    createInertiaApp({
      // ...
      setup({ el, App, props }) {
        const stateProps = $state(props)
        
        if (el.dataset.serverRendered === 'true') {
          hydrate(App, { target: el, props: stateProps })
        } else {
          mount(App, { target: el, props: stateProps })
        }
        
        return stateProps
      }
    })
    ```

    This allows internal detection and updating of reactive props, but requires more user setup.

2. Svelte 5-specific `createInertiaApp`:

    ```js
    createInertiaAppForSvelte5({
      // ...
      setup({ el, App, props }) {
        // props is already reactive
        
        if (el.dataset.serverRendered === 'true') {
          hydrate(App, { target: el, props })
        } else {
          mount(App, { target: el, props })
        }
      }
    })
    ```

    This handles reactive props internally, simplifying usage for Svelte 5 projects.

While Svelte 5 is stable, it's not officially released yet. Developers using it are early adopters who likely understand they're working with pre-release software. This means we have some leeway to implement temporary solutions during the transition period, even if they require a bit more setup from users.

These solutions haven't been tested yet. Feedback and additional ideas are welcome.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ For changes prior to v1.0.0, see the [legacy releases](https://legacy.inertiajs.
 - Fix form helper `transform` return type in React adapter ([#1896](https://github.com/inertiajs/inertia/pull/1896))
 - Use updater function in `setData` in `useForm` hook in React adapter ([#1859](https://github.com/inertiajs/inertia/pull/1859))
 - Skip intercepting non-left button clicks on links ([#1908](https://github.com/inertiajs/inertia/pull/1908), [#1910](https://github.com/inertiajs/inertia/pull/1910))
+- [Svelte] Fix props reactivity ([#1969](https://github.com/inertiajs/inertia/pull/1969))
 - [Svelte] Fix `<Render />` component to respect `preserveState` option ([#1943](https://github.com/inertiajs/inertia/pull/1943))
 - [Svelte] Fix 'received an unexpected slot "default"' warning ([#1941](https://github.com/inertiajs/inertia/pull/1941))
```
reinink, line 19: Minor, but this is generally the format we used to identify the adapter.

```suggestion
- Fix props reactivity in Svelte adapter ([#1969](https://github.com/inertiajs/inertia/pull/1969))
- Fix `<Render />` component to respect `preserveState` option in Svelte adapter ([#1943](https://github.com/inertiajs/inertia/pull/1943))
- Fix 'received an unexpected slot "default"' warning in Svelte adapter ([#1941](https://github.com/inertiajs/inertia/pull/1941))
```

packages/svelte/src/lib/components/App.svelte
```diff
@@ -1,17 +1,60 @@
...
+    }
+
+    return h(layout, pageProps, child ? [child] : [], key)
+  }
 </script>
 
-<Render {...components} />
+{#if props}
+  <Render {...props} />
+{/if}
```
reinink, line 60: Missing line return at end of this fine

packages/svelte/src/lib/components/App.svelte
```diff
@@ -1,17 +1,60 @@
...
+        .reverse()
+        .reduce((child, layout) => h(layout, pageProps, [child], key))
+    }
+
+    return h(layout, pageProps, child ? [child] : [], key)
+  }
 </script>
 
-<Render {...components} />
+{#if props}
```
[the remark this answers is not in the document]
pedroborges: The simple answer is never! The long answer is:

1. The [store is initialized](https://github.com/inertiajs/inertia/blob/24cbac37f6e441e6a68ce31e0643e06809616141/packages/svelte/src/lib/store.ts#L11-L15) with `null` values.
2. The store [values are set](https://github.com/inertiajs/inertia/blob/24cbac37f6e441e6a68ce31e0643e06809616141/packages/svelte/src/lib/createInertiaApp.ts#L47-L51) to initial data before the router inits and the `<App />` component is mounted.
3. For this reason, [we check](https://github.com/inertiajs/inertia/blob/24cbac37f6e441e6a68ce31e0643e06809616141/packages/svelte/src/lib/components/App.svelte#L14) for component and page on the `resolveProps` function and return `null` if they are missing. In addition, this check is needed to satisfy TS.

So, this is equivalent to [this check](https://github.com/inertiajs/inertia/blob/24cbac37f6e441e6a68ce31e0643e06809616141/packages/vue3/src/app.ts#L83) on the Vue 3 adapter.

PR #1970, opened 2024-09-11
[1.x] Add SSR support for Svelte 5

pedroborges:
The primary goal for this PR is to maintain Svelte 4 as the default version in the adapter, ensuring stability for existing projects. However, I've also added support for Svelte 5, taking advantage of its backward compatibility with Svelte 4. This allows developers to upgrade their projects to Svelte 5 without needing to modify the adapter or deal with any breaking changes.

TL;DR:
The adapter remains on Svelte 4, but you can use either Svelte 4 or Svelte 5 in your projects seamlessly.

In Svelte 5 users no longer need to explicitly enable client side hydration in their `vite.config.js` file:

```diff
+     svelte(),
-     svelte({
-       compilerOptions: {
-         hydratable: true,
-       },
-     }),
```


pedroborges:
@jamesst20 please help us test this PR if you get a chance 😉

packages/svelte/src/lib/createInertiaApp.ts
```diff
@@ -50,12 +52,27 @@ export default async function createInertiaApp({
...
+      return (SSR as SSRComponent).render({ id, initialPage })
+    })() as SvelteRenderResult
+
+    return css
+      ? {
+          body: html,
+          head: [head, `<style data-vite-css>${css.code}</style>`],
+        }
+      : {
+          body: `<div data-server-rendered="true" id="${id}" data-page="${escape(JSON.stringify(initialPage))}">${html}</div>`,
```
pedroborges, line 73: I tried other ways to achieve this but couldn't using the `SSR` component, [this solution](https://github.com/jamesst20/inertia/blob/c8493ab9358fb520461940ef7f043987f0b39d17/packages/svelte/src/lib/createInertiaApp.ts#L52) was suggested by @jamesst20.

packages/svelte/tsconfig.json
```diff
@@ -20,6 +20,6 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "preserveConstEnums": true,
-    "removeComments": true
+    "removeComments": false
```
pedroborges, line 23: Needed to preserve the `/* @vite-ignore */` comment that prevents Vite from statically analyzing and bundling `svelte/server` during development.

packages/svelte/package.json
```diff
@@ -31,21 +31,21 @@
     "!dist/**/*.spec.*"
   ],
   "peerDependencies": {
-    "svelte": "^3.20.0 || ^4.0.0"
+    "svelte": "^3.20.0 || ^4.0.0 || ^5.0.0 || ^5.0.0-next.244"
```
pedroborges, line 34: [Svelte 3.59.2](https://www.npmjs.com/package/svelte/v/3.59.2), the latest version, was released over a year ago. We should consider whether it's worth continuing support for it. As far as I know, it still works with the current Svelte adapter, but with the upcoming v2.0 features and Svelte 5 on the horizon, we don't have the capacity to test across all versions to ensure full compatibility.

packages/svelte/src/lib/createInertiaApp.ts
buhrmi: Wouldn't this solution also work for Svelte 4, meaning we can get rid of SSR.svelte?

pedroborges: @buhrmi That was a great idea, just tested and it totally works ✅

jamesst20:
> @jamesst20, please help us test this PR if you can spare a moment 😉

Appears to be fine on my end! The approach of checking the Svelte version at runtime is the first approach I had taken in the past but for some reason it failed to work because the detected version at runtime was the one @inertiajs/svelte was built with instead of the version being used in the playground. I wonder where I failed. Not sure if this is the right commit in the history but it looked like this https://github.com/jamesst20/inertia/commit/db5401558f474f87bf1bce80209f2163d50b6e62#diff-ee8bc50b28e24eefd929600798a4a4c3283bd349a8bd968b0bef56c08271f628R7 

It appears your implementation works as expected which is nice !

pedroborges:
Thank you for testing it out, @jamesst20! I appreciate your feedback. Yes, passing the module path to `dynamicImport` was intentional—it was necessary after several failed attempts.

packages/svelte/package.json
pedroborges: Since this PR will be merged into Inertia 1.3, we'll maintain support for Svelte 3 for now and drop it with the release of Inertia V2.

jamesst20:
> Thank you for testing it out, @jamesst20! Yes, passing the module path to `dynamicImport` was intentional—it was necessary after several failed attempts.

Yeah this is crazy the right idea was there I just couldn't get it to work and yet it was so simple.

I wonder here https://github.com/jamesst20/inertia/commit/b9a902d0203ed0083be6adc3a1c3e9f9196fa6c7#diff-ee8bc50b28e24eefd929600798a4a4c3283bd349a8bd968b0bef56c08271f628R50 if I had changed 

```javascript
const { render } = await import('svelte/server')
return render(SSR, { props: { id, initialPage } })
```

to 

```javascript
const svelteImport = 'svelte/server'
const { render } = await import(svelteImport)
return render(SSR, { props: { id, initialPage } })
```

if this would also have worked. Anyway this is good news as breaking changes that relates to Inertia can be completely avoided

jamesst20:
Out of subject a little bit because I can see there are build errors. I had some troubles as well and was forced to move from NPM to PNPM at some point. I am not sure if this is what is happening here. From my testing I realized that when using NPM, the specified version in the packages.json from each "workspace" were not necessary always respected. I also realized that some of these workspaces were depending on some dependencies that were not **explicitly** declared in the packages.json but because some other project in the workspace were using them directly or indirectly from a subdepedency, it worked. If you are ever interested in taking a look, I had a PR opened to migrate to PNPM. Unfortunatly that PR was based on top of another one which updated every packages (https://github.com/inertiajs/inertia/pull/1874)  but the relevent files are `{packages,playgrounds}/**/package.json` and `pnpm-workspace.yaml` and `.npmrc` and the github workflow itself. https://github.com/inertiajs/inertia/pull/1875/files#diff-0b810c38f3c138a3d5e44854edefd5eb966617ca84e62f06511f60acc40546c7

This had received 0 attention but I still think it was worth something https://github.com/inertiajs/inertia/pull/1873

.github/workflows/build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 16.15
+          node-version: 20.15
```
pedroborges, line 20: Necessary because I got a `ReferenceError: [plugin vite-plugin-svelte] structuredClone is not defined` error on the build workflow while building the Svelte 5 playground.

Issue #1922, opened 2024-07-14
Scroll restoration issue


mrleblanc101:
### Version:

- `@inertiajs/vue3` version: 1.2.0

### Describe the problem:

If you click on a link at the bottom of a page, then use any mean to go to the previous page (back button, trackpad gesture, history.back()), the scroll position is not always properly restored.

As you can see in this exemple, the first back does not restore the correct position, but the second back does.
![2024-07-13 22 20 50](https://github.com/user-attachments/assets/670d1f75-cac4-420b-aa94-05457525eb3c)

### Steps to reproduce:

Here is the website where you can reproduce the issue: https://tee-time.ca/search?city_id=65&date=2024-07-14
If the is no results, try changing the date in the top-left corner.

mrleblanc101:
Probably relates to #1915

benjivm:
Same as #1698 does not appear to affect Chrome, at least in my testing and repro repo.

pedroborges:
Fixed by #1980 ✅

Issue #1816, opened 2024-03-02
Intermittent scroll restoration issues when using code-splitting


mbise1993:
### Version:

- `@inertiajs/react` version: 1.0.14

### Describe the problem:

Scroll restoration fails intermittently on back navigation when using code-splitting (i.e. resolving components asynchronously). I've observed this both with browser-default scroll restoration and by forcing usage of Intertia's scroll restoration mechanism by adding the `scroll-region` attribute to my page's scrolling element.

I've already spent some time digging into this (and [asked about it in the Discord](https://discord.com/channels/592327939920494592/1192987019064909946/1192987019064909946) awhile back) and from what I can tell, the issue is a race condition between when the component we're navigating back to finishes rendering and when the scroll restoration mechanism executes. It seems that when using async components, it is possible for the the browser/Inertia to attempt to restore the scroll position _before_ the component has completely loaded and rendered. I've verified this by switching my app to use eager loading for component resolution - after that change, the scroll position is restored as expected on back navigations. I'm not sure if this issue appears with other frameworks as well, but I imagine it could since they are also using Vite.

The simplest solution I've found is wrapping the `scrollTo()` call and `scrollTop`/`scrollLeft` setters in the router in `requestAnimationFrame` or `setTimeout(() => {...}, 0)`. I'm not sure if this is a fool-proof solution, but its worked so far in my testing. I'd be happy to submit a PR for whatever solution you think would be best.

### Steps to reproduce:

1. Clone the reproduction repo here: https://github.com/mbise1993/inertia-scroll-restoration-demo
2. Follow the instructions in the README to run the app and see the error

benjivm:
Same as https://github.com/inertiajs/inertia/issues/1698 does not appear to affect Chrome, at least in my testing and repro repo.

pedroborges:
Fixed by #1980 ✅

PR #1980, opened 2024-09-19
[1.x] Fix `resetScrollPositions` and `restoreScrollPositions` router methods

### LINKED ISSUE #1211: Scroll regions inside page component are not restored after form submit.

pedroborges:
This PR resolves two scroll-related problems in Inertia:

1. **Scroll Reset on Page Navigation**
    - **Issue:** Using Inertia’s `Link` causes the scroll position to reset briefly before loading new content, leading to a flash of the previous page and jank on heavier or mobile pages.
    - **Details:** [#1802](https://github.com/inertiajs/inertia/issues/1802)
    - **Previous Attempt:** [#1803](https://github.com/inertiajs/inertia/pull/1803) tried to fix this but added complexity. This PR provides a simpler solution.

2. **Scroll Position Restoration for `[scroll-region]` Elements**
    - **Issue:** In the React adapter, `[scroll-region]` elements don't restore their scroll positions when navigating back because `restoreScrollPositions` doesn't detect them before the document body updates.
    - **Details:** [#1816](https://github.com/inertiajs/inertia/issues/1816), [#1802](https://github.com/inertiajs/inertia/issues/1802), [#1211](https://github.com/inertiajs/inertia/issues/1211)
    - **Previous Attempt:** [#1796](https://github.com/inertiajs/inertia/pull/1796) used `setTimeout`, but `nextFrame` is a better approach.

### Solution

Used a **double `requestAnimationFrame` (RAF)** method to ensure callbacks run after the next repaint. This approach is used internally by Vue:

- **Double RAF with Callback:** [Vue Transition Component](https://github.com/vuejs/core/blob/dad673809929c084dcb8e42640eb7daa675d4ea4/packages/runtime-dom/src/components/Transition.ts#L326-L330)
- **Double RAF with Promise:** [Vue E2E Utils](https://github.com/vuejs/core/blob/dad673809929c084dcb8e42640eb7daa675d4ea4/packages/vue/__tests__/e2e/e2eUtils.ts#L157-L165)

This fix is implemented in the core, making it safe for React, Vue, and Svelte adapters without needing adapter-specific changes.

----

Fixes #1211, #1459, #1698, #1796, #1802, #1803, #1816, #1922

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -33,6 +33,11 @@ import { hrefToUrl, mergeDataIntoQueryString, urlWithoutHash } from './url'
 
 const isServer = typeof window === 'undefined'
 const cloneSerializable = <T>(obj: T): T => JSON.parse(JSON.stringify(obj))
+const nextFrame = (callback: () => void) => {
+  requestAnimationFrame(() => {
+    requestAnimationFrame(callback)
+  })
+}
 
 export class Router {
   protected page!: Page
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -126,21 +126,21 @@ export class Router {
   }
 
   protected resetScrollPositions(): void {
-    nextFrame(() => window.scrollTo(0, 0))
-    this.scrollRegions().forEach((region) => {
-      if (typeof region.scrollTo === 'function') {
-        region.scrollTo(0, 0)
-      } else {
-        region.scrollTop = 0
-        region.scrollLeft = 0
+    nextFrame(() => {
+      window.scrollTo(0, 0)
+      this.scrollRegions().forEach((region) => {
+        if (typeof region.scrollTo === 'function') {
+          region.scrollTo(0, 0)
+        } else {
+          region.scrollTop = 0
+          region.scrollLeft = 0
+        }
+      })
+      this.saveScrollPositions()
+      if (window.location.hash) {
+        document.getElementById(window.location.hash.slice(1))?.scrollIntoView()
       }
     })
-    this.saveScrollPositions()
-    if (window.location.hash) {
-      // We're using a setTimeout() here as a workaround for a bug in the React adapter where the
-      // rendering isn't completing fast enough, causing the anchor link to not be scrolled to.
-      setTimeout(() => document.getElementById(window.location.hash.slice(1))?.scrollIntoView())
-    }
   }
 
   protected restoreScrollPositions(): void {
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -92,7 +92,7 @@ export class Router {
     if (!this.page.url.includes(hash)) {
       this.page.url += hash
     }
-    this.setPage(page, { preserveScroll: true ,preserveState: true }).then(() => fireNavigateEvent(page))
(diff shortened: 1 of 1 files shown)


packages/core/src/router.ts
```diff
@@ -121,25 +126,27 @@ export class Router {
...
+      this.saveScrollPositions()
+      if (window.location.hash) {
+        document.getElementById(window.location.hash.slice(1))?.scrollIntoView()
       }
     })
-    this.saveScrollPositions()
-    if (window.location.hash) {
-      // We're using a setTimeout() here as a workaround for a bug in the React adapter where the
-      // rendering isn't completing fast enough, causing the anchor link to not be scrolled to.
-      setTimeout(() => document.getElementById(window.location.hash.slice(1))?.scrollIntoView())
```
pedroborges, line 137: `setTimeout` is no longer needed

joetannenbaum (approved):
Looks great, thanks @pedroborges!

PR #1984, opened 2024-09-21
[1.x] Fix history navigation issue on Chrome iOS

### LINKED ISSUE #1954: Browser Back/Forward buttons issue on iOS Chrome
### Version:

- `@inertiajs/vue2` version:  1.2.0

### Describe the problem:
Browser history not working properly on **iOS Chrome v 128.0.6613.98**
When i navigate through my web app and then click browsers "Back" button or using  gesture "swipe from left" Chrome does not follow correct path (not navigating to previous page)
I use <Link> component from @inertiajs/vue2 package
When i check on iOS Safari or Brave browser everything works fine

Browser back and forward and swipe actions should work properly based on history

### Steps to reproduce:

1. Open https://demo.inertiajs.com/login or https://inertiajs.com/  or another web app which uses inertiajs on **Chrome iOS**
2. Try to navigate through web app (open some links)
3. hit the back button or swipe from left gesture

pedroborges:
On Chrome iOS, calling `this.storeScrollPositions` (which uses `history.replaceState`) right before `this.setPage` in the router `visit` method caused timing issues with `history.pushState`, leading to navigation problems.

Wrapping `window.history.pushState` with `setTimeout` inside `router.pushState` fixes the issue. Using `setTimeout` defers `history.pushState` to the next event loop tick, preventing conflicts between `history.replaceState` and `history.pushState`. This change doesn’t introduce any side effects since `setTimeout` with a delay of 0 just delays the execution without altering functionality.

Fixes #1954.

## Before

https://github.com/user-attachments/assets/f7ad94e9-46fb-4bec-9897-8d54758cd951

## After

https://github.com/user-attachments/assets/2bb29406-ceb8-41ef-b4bd-8a5bff2f75f0

We could enhance this by detecting Chrome on iOS and only applying `setTimeout` there, but the current fix works universally without issues 🤷‍♂️

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -492,7 +492,7 @@ export class Router {
 
   protected pushState(page: Page): void {
     this.page = page
-    window.history.pushState(cloneSerializable(page), '', page.url)
+    setTimeout(() => window.history.pushState(cloneSerializable(page), '', page.url))
   }
 
   protected replaceState(page: Page): void {
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -492,6 +492,8 @@ export class Router {
 
   protected pushState(page: Page): void {
     this.page = page
+    // Defer history.pushState to the next event loop tick to prevent timing conflicts.
+    // Ensure any previous history.replaceState completes before pushState is executed.
     setTimeout(() => window.history.pushState(cloneSerializable(page), '', page.url))
   }
 


joetannenbaum (approved):
Looks good, thanks!

AbdullahAgsar:
This solution didn’t work. I cloned it from the master branch and used these changes in my project with npm link. While the issue persists on iOS browsers, I also started experiencing problems on browsers on my macOS device. @pedroborges

pedroborges:
@AbdullahAgsar Yeah, I noticed last week that this fix caused some side effects on macOS browsers as well. Could you try the fix from #1992? I just tested it again and it resolves the navigation issues for me.

AbdullahAgsar:
@pedroborges I tried the solution from [#1992](https://github.com/inertiajs/inertia/pull/1992), but it didn’t work. I continue to experience issues with both forward and backward navigation in macOS browsers. Additionally, the issue in iOS browsers still persists.

pedroborges:
@AbdullahAgsar Thanks for the feedback. Just to confirm, are you testing this in your project or one of the playgrounds?

Also, would you be able to provide a minimal reproducible repository? I’d love to investigate the issue further.

AbdullahAgsar:
@pedroborges I tested it in my own project. I will try to create a blank project where the same issue occurs, and once completed, I will share the GitHub link with you.

pedroborges:
@AbdullahAgsar Thanks, I’d really appreciate your help in reproducing the issue. I just tested #1992 on the [Vue](https://github.com/inertiajs/pingcrm) and [Svelte](https://github.com/inertiajs/pingcrm-svelte) versions of Ping CRM, and it resolved the navigation issue in those projects as well.

AbdullahAgsar:
Hello @pedroborges, I’ve been trying to solve the issue I’ve been facing for a long time using your solution. The issue wasn’t only present in iOS Chrome but also in iOS Edge. After making some adjustments along with your solution, I was able to come up with a fix for both browsers.

Initially, in the resolve block, I was using the following code:
```
const pages = import.meta.glob('./Pages/**/*.vue', { eager: true })
return pages[`./Pages/${name}.vue`]
```

But I had modified it to:
```
const pages = import.meta.glob('../../Pages/**/*.vue')
return pages[`../../Pages/${name}.vue`]()
```

After reverting this code back to its original state, most of the issues were resolved. However, I was still experiencing problems with the back navigation, like adding the same page to the history twice. To resolve these, I removed the Chrome check from the replaceState() function and adjusted it like this:
`window.history.replaceState(cloneSerializable(page), '', page.url)`

For the Edge browser, I created a variable called isEdgeIOS. With these changes, I was able to fix all the back navigation issues on both browsers.

Screenshot 2024-10-08 at 13 51 54

Screenshot 2024-10-08 at 13 52 20

PR #1985, opened 2024-09-21
[1.x] Fix isDirty after form.defaults() call in Vue

### LINKED ISSUE #1862: Form defaults() then reset() doesn't work as expected
### Version:

- `@inertiajs/vue3` version: 1.0.6

### Describe the problem:

The form does not reset **isDirty** state after calling **defaults()** and then calling **reset()**
This behavior is observed if there are only primitive values in the form
I assume that watcher doesn't see changes and does not change the IsDirty state
```js
watch(
  form,
  (newValue) => {
    form.isDirty = !isEqual(form.data(), defaults)
    if (rememberKey) {
      router.remember(cloneDeep(newValue.__remember()), rememberKey)
    }
  },
  { immediate: true, deep: true },
)
```
### Steps to reproduce:

Create a form as in the playground:

```js
const form = useForm('NewUser', {
  name: '',
  company: '',
  role: '',
})
```

Try to change some value

image

Next, call defaults() and reset(), isDirty will not change its state

https://github.com/inertiajs/inertia/assets/1456499/81c37c7e-c0d9-4c7d-8c5b-9e8e27574674

pedroborges:
When calling `defaults()` on a form, the `isDirty` state does not reset as expected. This occurs because the watcher only monitors changes to the `form` reactive properties, while the `defaults()` method modifies the default properties, which are not part of the watched `form`. As a result, `isDirty` is never recalculated and remains `true`.

This fix ensures that `isDirty` accurately reflects the form's state after resetting to defaults similarly to how the React adapter does it.

Fixes #1862.

### DIFF
--- a/packages/vue2/src/useForm.ts
+++ b/packages/vue2/src/useForm.ts
@@ -79,6 +79,7 @@ export default function useForm<TForm extends FormDataType>(...args): InertiaFor
 
       if (typeof key === 'undefined') {
         defaults = this.data()
+        this.isDirty = false
       } else {
         defaults = Object.assign({}, cloneDeep(defaults), value ? { [key]: value } : key)
       }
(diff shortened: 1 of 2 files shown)


joetannenbaum (approved):
Looks good, thank you!

PR #1988, opened 2024-09-25
inject css in head to prevent content shift when using SSR, fixes #1760

### LINKED ISSUE #1760: Svelte SSR does not render component CSS
### Version:

- `@inertiajs/svelte` version: 1.0.14

### Describe the problem:

When using Svelte SSR, the server side renderer does not render the component CSS, which results in content shifting around once the CSS is loaded via Javascript.

buhrmi:
Svelte 4 removed the CSS output from the component's `render` method. However, the `css: 'inject'` compiler option can be used instead to inject that CSS into the `<head>`, as documented [here](https://svelte-5-preview.vercel.app/docs/breaking-changes#components-are-no-longer-classes-server-api-changes). According to Rich Harris, this solution is only for toy projects. So I guess we'll just stay a toy project forever.

### DIFF
--- a/playgrounds/svelte5/vite.config.js
+++ b/playgrounds/svelte5/vite.config.js
@@ -9,6 +9,10 @@ export default defineConfig({
       ssr: 'resources/js/ssr.js',
       refresh: true,
     }),
-    svelte(),
+    svelte({
+      compilerOptions: {
+        css: 'injected'
+      }
+    }),
   ],
 })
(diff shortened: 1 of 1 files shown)


pedroborges:
@buhrmi My understanding from issue #1760 is that the `css: 'inject'` compiler option is typically needed when you are styling directly within components. Since that's not the case for the Svelte playground, I'm not sure if the change is necessary. Also, I haven't noticed any CSS-related issues in the current playground setup.

Could you clarify if I'm missing something here? If the issue you were facing has been resolved, feel free to close the PR 😉

buhrmi:
@pedroborges Yeah, you're right. It doesn't make sense to add this to the playground. For some reason, my mind was set on providing a "feature-complete" reference project, where styling in components also works. But yeah, since the playground isn't using any regular styling (only tailwind), it doesn't make sense. We should however find a place in the documentation to tell people that `css: 'inject'` is needed for SSR when using component styling.

PR #1992, opened 2024-09-27
[1.x] Fix history navigation issue

pedroborges:
I noticed an issue introduced by #1984 when using the Vue 2/3 and Svelte adapters. On the playgrounds for these adapters, navigating through **Home** → **Users** → **Article** → **Form** and then trying to go back in history skips the **Article** page. 

### Root cause:

After investigating, I found that `router.replaceState` (which remembers the form state) was being called **before** `router.pushState`, effectively replacing the **Article** history state.

before-fix

### Solution:

I tried several approaches but determined that the safest solution is to scope the fix from #1984 specifically to Chrome on iOS, where the initial issue occurs. This avoids any unintended side effects in other browsers.

after-fix

----

I’ve tested the fix across all adapters on **Chrome/Safari/Firefox** on macOS and **Chrome/Safari** on iOS, and everything is functioning as expected.

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -31,8 +31,8 @@ import {
 } from './types'
 import { hrefToUrl, mergeDataIntoQueryString, urlWithoutHash } from './url'
 
-const isChromeIOS = /CriOS/.test(navigator.userAgent)
 const isServer = typeof window === 'undefined'
+const isChromeIOS = !isServer && /CriOS/.test(window.navigator.userAgent)
 const cloneSerializable = <T>(obj: T): T => JSON.parse(JSON.stringify(obj))
 const nextFrame = (callback: () => void) => {
   requestAnimationFrame(() => {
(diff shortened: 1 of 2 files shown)


packages/core/src/router.ts
```diff
@@ -31,6 +31,7 @@ import {
 } from './types'
 import { hrefToUrl, mergeDataIntoQueryString, urlWithoutHash } from './url'
 
+const isChromeIOS = /CriOS/.test(navigator.userAgent)
```
reinink, line 34: Have you tested this in SSR by any chance to make sure it works? I'm worried it might blow up because `navigator` isn't available. Also, should it be `window.navigator`?

pedroborges: You are right, I'll fix this and test once again with SSR.

pedroborges: Fixed ✅

reinink (approved):

PR #1999, opened 2024-10-08
fix chromeIOS and edgeIOS browser back button problems

AbdullahAgsar:


RobertBoes:
Seems like this might require more than just a list of browsers. All third party browsers on iOS are essentially glorified Safari webviews. So this issue would likely exist in every single browser on iOS apart from Safari itself. Perhaps it would be a good idea to change the check to "is ios and not safari"

AbdullahAgsar:
@RobertBoes This issue was also present in Safari itself in the past few weeks, but it was resolved within just a few days. I’m not entirely sure, but I think it received an update. I believe the issue will be completely resolved once the browsers receive a new version.

packages/core/src/router.ts
```diff
@@ -495,24 +496,18 @@ export class Router {
...
 
   protected replaceState(page: Page): void {
     this.page = page
-    if (isChromeIOS) {
-      // Defer history.replaceState to the next event loop tick to prevent timing conflicts.
-      // Ensure any previous history.pushState completes before replaceState is executed.
-      setTimeout(() => window.history.replaceState(cloneSerializable(page), '', page.url))
-    } else {
-      window.history.replaceState(cloneSerializable(page), '', page.url)
-    }
```
reinink, line 515: Curious, why did you delete this block of code? Wouldn't it just need to be updated to `isChromeIOS || isEdgeIOS` as well?

AbdullahAgsar: During my testing, I observed that when navigating from one page to another, the replaceState function is called first, followed by pushState, and then replaceState is triggered again. For example, when transitioning from the home page to the dashboard, the browser would first process the home page with replaceState, then execute pushState for the dashboard page, and afterward call replaceState again. With this sequence, the back navigation worked, but I had to click the back button twice to return to the previous page. After adjusting the replaceState function, the issue of adding the same page twice was resolved, and now a single back button click is enough to return to the previous page.

Kokkey:
any updates on this ?
I'm having the same issues

AbdullahAgsar:
> any updates on this ? I'm having the same issues

i don't have any solution for this issues

pascalbaljet:
I tried to reproduce the issue on iPadOS + Edge, but I can't reproduce the original bug from #1984. Here's the user-agent with the full versions.

> Mozilla/5.0 (iPad; CPU OS 18_5_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) EdgiOS/137.0.3296.82 Version/18.0 Mobile/15E148 Safari/604.1

Is this issue still relevant? If so, can you give me instructions on how to reproduce it?

RyanDaDeng:
I have same problem here

pascalbaljet:
Closing this one in favor of #2629.

PR #2002, opened 2024-10-09
Fix Svelte version import

pedroborges:
[Sam reported](https://x.com/saml_dev/status/1843828546051158381) that the new svelte adapter adds ~500 kb to his app bundle size.

After some investigation I found the culprit, it’s the way we import the `VERSION` constant from the Svelte compiler in the `createInertiaApp.ts`. This PR fixes that.

## Before

Screen Shot 2024-10-09 at 03 49 25

## After

Screen Shot 2024-10-09 at 03 50 28


joetannenbaum (approved):
Looks good! Thanks for looking into this

PR #2016, opened 2024-10-11
[2.x] Export `usePrefetch` for react

nick-potts:
the `usePrefetch` wasn't exported in the first beta version.


pedroborges (approved):
Good catch! Thanks, @nick-potts!

PR #2019, opened 2024-10-11
[2.x] Fix `onClick` in React adapter

### LINKED ISSUE #2018: Inertia V2 beta - onClick doesn't get called anymore
### Version:

- `@inertiajs/react` version: #.#.#

### Describe the problem:

After switching from v1 to v2 all onClick handlers for the <Link /> elements don't work anymore.

### Steps to reproduce:

```ts
<Link href='/' onClick={() => alert(1)}>Click here</Link>
<Link as="button" method="post" href='/' onClick={() => alert(1)}>Click here</Link>
```

pedroborges:
Right now, `onClick` handlers passed to the React `<Link />` component are only called when the prefetch mode is `click` as reported in #2018. This is PR fixes that.

Closes #2018.

### DIFF
--- a/packages/react/src/Link.ts
+++ b/packages/react/src/Link.ts
@@ -166,6 +166,8 @@ const Link = forwardRef<unknown, InertiaLinkProps>(
 
     const regularEvents = {
       onClick: (event) => {
+        onClick(event)
+
         if (shouldIntercept(event)) {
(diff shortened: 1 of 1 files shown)


joetannenbaum (approved):
Looks good, thank you!

PR #2026, opened 2024-10-13
Svelte fix when using deferred props

### LINKED ISSUE #2012: Deferred svelte component renders the default slot too early
### Version:

- `@inertiajs/svelte` version: 2.0.0-beta.1

### Describe the problem:

There is a bug with the `Deferred` svelte component. It will render the components of the default slot before the props for the Svelte component have been updated. This is because the `$page.props` store is updated first, or at least a render happens after the store is updated but before the regular `export let` prop is updated. 

Since the `Deferred` component [decides what to render based on the `$page` store](https://github.com/inertiajs/inertia/blob/370173404f97d796f3b072d619d821b95d86a6eb/packages/svelte/src/components/Deferred.svelte#L13), it renders the default slot too early.

Note that passing a primitive as the deferred prop does not throw an error on the frontend, because it will just render undefined for that single render cycle. But when you pass an object as a deferred prop and try to access its properties, it throws when trying to access a property on undefined.

### Steps to reproduce:

I created a repro here: https://github.com/saml-dev/inertia2-repro
It includes the `.env` and `database.sqlite` to make your lives easier
1. clone above repo
2. `npm i`
3. `composer install`
4. `npm run dev` (and `php artisan serve` if not using herd/valet :D )
5. open the app to `/` and use the two links to see how one works and one doesn't

Relevant files in the code are:
- `Page1.svelte` and `Page2.svelte` in `resources/js/Pages`
- `routes/web.php` renders each page with the same deferred prop

saml-dev:
This fixes #2012

### DIFF
--- a/packages/react/test-app/Pages/DeferredProps/Page1.jsx
+++ b/packages/react/test-app/Pages/DeferredProps/Page1.jsx
@@ -3,13 +3,13 @@ import { Deferred, Link, usePage } from '@inertiajs/react'
 const Foo = () => {
   const { foo } = usePage().props
 
-  return foo
+  return foo.text
 }
 
 const Bar = () => {
   const { bar } = usePage().props
 
-  return bar
+  return bar.text
 }
 
 export default () => {
--- a/packages/svelte/src/page.ts
+++ b/packages/svelte/src/page.ts
@@ -1,6 +1,16 @@
 import { derived } from 'svelte/store'
 import store from './store'
 
-const page = derived(store, ($store) => $store.page)
+let initialized = false
+const page = derived(store, ($store, set) => {
+  // This ensures that page props and the $page store are updated
+  // in the same render, such as when using deferred props.
+  if (initialized && typeof window !== 'undefined') {
+    window.queueMicrotask(() => set($store.page))
+    return
+  }
+  set($store.page)
+  initialized = true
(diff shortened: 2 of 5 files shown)


saml-dev:
I just added tests. When I updated `server.js` to return an object to Page1 instead of text, it caused the svelte and react tests to fail, but strangely I couldn't get the Vue tests to fail. I updated `Page1.vue` anyway but let me know if I'm missing something there.

pedroborges:
Thanks for working on this, @saml-dev. After reviewing this change, I realized it could potentially cause significant compatibility issues for existing projects. Instead, I developed an approach that makes `$page` updates [more predictable](https://github.com/inertiajs/inertia/pull/2036) and implemented [another fix](https://github.com/inertiajs/inertia/pull/2037) for the `<Deferred />` component.

I would greatly appreciate it if you could test these PRs and provide your feedback. 🙏

reinink:
@saml-dev Hey thanks so much for this contribution, but as @pedroborges mentioned he's solved this in a slightly different way in #2037, which I just merged in. Thanks so much either way! 🙏

saml-dev:
Hey no problem! It was a good exercise to get familiar with the Inertia code base. Hopefully I can contribute another way in the future!

PR #2029, opened 2024-10-14
Correctly encrypt history state, fixes #2028

### LINKED ISSUE #2028: History encryption breaks if the data contains emojis
### Describe the problem:

History encryption breaks if a prop contains an emoji.

### Steps to reproduce:

Have a prop with an emoji somewhere in the value.
Navigate to another page.
Navigate back to page containing emojis.
Encrypted history fails to restore.

buhrmi:


packages/core/src/encryption.ts
```diff
@@ -47,9 +47,7 @@ const encryptData = async (iv: Uint8Array, key: CryptoKey, data: any) => {
 
   const textEncoder = new TextEncoder()
   const str = JSON.stringify(data)
-  const encoded = new Uint8Array(str.length)
```
skryukov, line 50: If we still want to use more performant `encodeInto` and not deal with UTF-8 issues, we can use `str.length * 3`: https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encodeInto#buffer_sizing

buhrmi: I'd rather have smaller history state. Already dumping lots of stuff in there, don't want it 3x the size

joetannenbaum:
Thanks for your pull request! I didn't see this until after I had one already open, sorry about that.

I'm going to close this in favor of #2030 since that one still uses `encodeInto` and just trims off anything we don't write.

PR #2030, opened 2024-10-14
[2.x] Fix multi byte strings in history encryption + test

### LINKED ISSUE #2028: History encryption breaks if the data contains emojis

joetannenbaum:
Fixes the buffer size to account for multi-byte characters when encrypting history.

Fixes #2028

### DIFF
--- a/packages/core/src/encryption.ts
+++ b/packages/core/src/encryption.ts
@@ -47,9 +47,9 @@ const encryptData = async (iv: Uint8Array, key: CryptoKey, data: any) => {
 
   const textEncoder = new TextEncoder()
   const str = JSON.stringify(data)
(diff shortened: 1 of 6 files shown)


RobertBoes:
As @buhrmi mentioned in his PR, is it really needed to take up 3 bytes for all characters, even if the content does not contain multibyte characters?

joetannenbaum:
We allot for the maximum (3x the characters) but only encode what we've written.

PR #2032, opened 2024-10-14
[2.x] $page url should not include the host

### LINKED ISSUE #2020: $page.url breaking change
### Version:

- `@inertiajs/vue3` version: lagest

### Describe the problem:

$page.url returns http://localhost:8000/ instead of /

### Steps to reproduce:

Update to v2 and {{ $page.url }}

joetannenbaum:
This was an accidental breaking change in 2.x, previously `$page.url` did not include the full URL, now it does. This PR fixes that.

Fixes #2020

### DIFF
--- a/packages/core/src/response.ts
+++ b/packages/core/src/response.ts
@@ -233,7 +233,7 @@ export class Response {
 
     setHashIfSameUrl(this.requestParams.all().url, responseUrl)
 
-    return responseUrl.href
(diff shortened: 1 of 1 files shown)


pedroborges (approved):

reinink (approved):
Yeah, for better or worse I've always tried to keep the host out of the URLs in Inertia.

PR #2034, opened 2024-10-15
[2.x] Fix Svelte `usePrefetch` in SSR mode

pedroborges:


packages/svelte/src/usePrefetch.ts
```diff
@@ -33,8 +33,13 @@ export default function usePrefetch(options: VisitOptions = {}) {
   })
 
   onDestroy(() => {
-    removePrefetchedListener()
-    removePrefetchingListener()
+    if (removePrefetchedListener) {
```
joetannenbaum, line 36: Do need these checks? Doesn't it default to a void func on lines 17-18?

pedroborges: Lines 17-18 only defines the types, not `noop` functions. SSR breaks without these checks:

```
TypeError: removePrefetchedListener is not a function
```

I simplified it by using optional chaining on the function calls.

reinink: I reverted this back to what you had — I think that was more clear 👍

joetannenbaum: I apparently don't know the diff between a colon and an equals sign, sorry 🤦‍♂️

PR #2036, opened 2024-10-15
[2.x] Refactor `createInertiaApp` in Svelte adapter

pedroborges:
While investigating issue #2012 with @saml-dev, I identified the need for more predictable updates to the `$page` store. This PR ensures that the `$page` store updates are more predictable to prevent reactivity issues, aligns the Svelte adapter with the other adapters, and simplifies the SSR (Server-Side Rendering) configuration.

> Please note that this PR does not fully resolve issue #2012. I will submit a separate PR that builds on these changes to address the remaining aspects of the issue.

Data resolution for the `<App />` component has been moved from the `createInertiaApp` function into the component itself. The internal `$store` has been removed, and the state it managed is now local within the `<App />` component. This change ensures that re-renders and `$page` store updates occur in a more predictable manner.

Additionally, I've simplified the SSR setup by allowing it to be managed by the user, consistent with the other adapters. This reduces the complexity within the adapter and provides more flexibility for custom server-side rendering configurations. I'll update the docs with instructions on how to add the `setup` callback to the `resources/js/ssr.js` file.

I have tested these changes on both playgrounds, including in SSR mode, as well as in two other projects, one using Svelte 4 and another using Svelte 5.

@joetannenbaum I welcome any feedback on how I have set up the Svelte-specific test. While the current test is functional, I know that it can be improved and plan to refactor it when I have the opportunity.


joetannenbaum (approved):
Thanks for chatting through this with me, looks good.

PR #2039, opened 2024-10-16
[2.x] Restore `router.resolveComponent()` method

pascalbaljet:
First of all, thanks for all the great work on v2! ❤️

In v1, the `resolveComponent` method was available on the router instance. This PR restores that for v2.

I am using this in [my modal package](https://inertiaui.com/inertia-modal/docs) to resolve the modal component, similarly to resolving a page. Without having this method publicly available, I must duplicate/reference the `resolveComponent` method in the main `app.js` file. That would be a shame, as it's so clean and nice in v1 :)


crankeye:
Hopefully this PR gets merged because it's much more clean to do it this way. 

For anyone else who's upgrading from Inertia 1.x to 2.x you can move the resolve function from `createInertiaApp()` in your `app.js` to it's own file. Then you can import the function in both the `app.js` and wherever you need to call `router.resolveComponent`.

Issue #1953, opened 2024-08-30
Poor tree-shaking


rkyoku:
### Version:

- `@inertiajs/svelte` version: 1.2.0

### Describe the problem:

**1. The router**

Just importing the router (`@inertiajs/core`) instantly adds 110KB to the final, minified uncompressed, "lib" in Svelte.

==> allow for a better tree-shaking of the router, as it shouldn't weigh 110KB minified (it probably includes some axios stuff and also other modules that increase weight)

I don't yet have a workaround for reducing output file when using the `router` (which is almost mandatory in any real app), but would definitely be grateful for one

**2. useForm**

Also, importing any module from `@inertiajs/svelte` prevents tree-shaking from removing unneeded stuff (e.g. `useForm`, which adds 30KB minified uncompressed because of `lodash`, even when we never use `useForm`)

==> maybe either use `lodash-es`, or don't include `useForm` in `index.js`

What I ended up doing is importing single `.js` files directly, e.g.:

```js
// not good:
// import { inertia } from '@inertiajs/svelte'
// good:
import { default as inertia } from '/node_modules/@inertiajs/svelte/src/link.js'
```

### Steps to reproduce:

Just import `Inertia` and check output file weight (using lib mode)

```js
import { router } from '@inertiajs/core';
// just make sure tree-shaking believes we are actually using the router
if (router) console.log('router');
```

pedroborges:
Thanks for reporting it @rkyoku, I'll be looking into this as part of the v2 work 👍

Qoraiche:
> Thanks for reporting it @rkyoku, I'll be looking into this as part of the v2 work 👍

@pedroborges This might be a bit off-topic, but could you let me know when version 2 will be released?

pascalbaljet:
> Thanks for reporting it @rkyoku, I'll be looking into this as part of the v2 work 👍

Thanks, @pedroborges! I think/hope this will also fix the issue of having to 'dedupe' when using the Inertia router in third-party packages:

```js
// vite.config.js
resolve: {
    dedupe: ['@inertiajs/vue3']
}
```

Without the `dedupe` option, the package will create a new instance of the router instead of reusing the one in your main app. This new instance, of course, doesn't have any state, which causes problems with navigation.

For future readers: I posted a full description of this problem on [Twitter/X](https://x.com/pascalbaljet/status/1803009939763319133).

pascalbaljet:
[...]

PR #2045, opened 2024-10-18
[2.x] Removal of NProgress dependency

### LINKED ISSUE #2031: [2.x] Should `nprogress` remain a dependency?
I wanted to open this to ensure awareness and have it tracked.

The core package requires `nprogress` which seems more or less abandoned. The last 0.2.0 release on npm was 9 years ago. 

4 years ago there was an attempt to release 1.0.0, which was tagged on GitHub but never made it as "latest" to npm. 

Issues and merge requests are ignored; new packages spammed.

Not sure how I feel about such a dependency in general. Additionally, not everyone needs progress bars. And support for the navigation api (native spinners, yay!) in all major browsers seems only months away.🤏

I was wondering if Inertia v2 would be a good moment to drop mentioned dependency and make it (and it's Inertia wrapper) optional?

joetannenbaum:
As mentioned in #2031, `nprogress` has been inactive for quite a long time and with the 2.x release it's probably smart to remove it as a dependency and bring it in house.

This also allowed for a refactor, bringing it up to more modern web standards. Among the improvements:

- Remove all vendor prefixes from css (all are now baseline)
- Use `classList` instead of manually parsing class attributes
- Use `style` property instead of manually parsing style attribute

It also allowed for tighter integration with the behavior we needed out of the component. Wins all around!

Fixes #2031

### DIFF
--- a/packages/core/src/progress-component.ts
+++ b/packages/core/src/progress-component.ts
@@ -1,0 +1,375 @@
+/* NProgress, (c) 2013, 2014 Rico Sta. Cruz - http://ricostacruz.com/nprogress
+ * @license MIT */
+
+import { ProgressSettings } from './types'
+
+const baseComponentSelector = 'nprogress'
+
+const settings: ProgressSettings = {
+  minimum: 0.08,
+  easing: 'linear',
+  positionUsing: 'translate3d',
+  speed: 200,
+  trickle: true,
+  trickleSpeed: 200,
+  showSpinner: true,
+  barSelector: '[role="bar"]',
+  spinnerSelector: '[role="spinner"]',
+  parent: 'body',
+  color: '#29d',
+  includeCSS: true,
+  template: [
+    '<div class="bar" role="bar">',
+    '<div class="peg"></div>',
+    '</div>',
+    '<div class="spinner" role="spinner">',
+    '<div class="spinner-icon"></div>',
+    '</div>',
+  ].join(''),
+}
+
+let status: number | null = null
+
+const configure = (options: Partial<ProgressSettings>) => {
+  Object.assign(settings, options)
+
+  if (settings.includeCSS) {
+    injectCSS(settings.color)
+  }
+}
+
+/**
+ * Sets the progress bar status, where `n` is a number from `0.0` to `1.0`.
+ */
+const set = (n: number) => {
+  const started = isStarted()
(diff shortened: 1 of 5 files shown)


NickSdot:
@joetannenbaum thanks for taking this on!

To be honest, I kinda was hoping to also get rid of some bundle size and unused ballast as a result of my question/proposal. 😅

Would you mind to gimme an idea of why having https://github.com/inertiajs/progress as an _optional_ thing wouldn't be sufficient? Afaik that was the plan for years, no?

Thank you in advance! 🙏

RobertBoes:
Another nice option would be if the progress stuff is treeshakable, then it wouldn't end up in the bundle if the option isn't enabled

packages/core/src/progress-component.ts
```diff
@@ -0,0 +1,375 @@
...
+const baseComponentSelector = 'nprogress'
+
+const settings: ProgressSettings = {
+  minimum: 0.08,
+  easing: 'linear',
+  positionUsing: 'translate3d',
+  speed: 200,
+  trickle: true,
+  trickleSpeed: 200,
+  showSpinner: true,
```
reinink, line 15: Kind of wish we could just remove the spinner entirely. Wish I knew how many people were actually using it.

reinink: Apparently some people like it: #96

NickSdot:
> Personally don't think it's unreasonable for Inertia to ship with a built-in progress bar, and I like how that keeps the installation story simpler.
> 
> 
> 
> Plus it's less than 400 lines of code, so not a big deal either way.

Almost impossible to argue against the legend himself, though. 😅 

However, I politely disagree. The installation story is really not _that_ much simpler. And, a 400 lines here, a 400 line there, these that - you know where I am coming from. 

Anyway, I understand and accept that is your call! I'd appreciate if you would take Roberts comment above and https://github.com/inertiajs/inertia/issues/1953 into account then.

Thanks for all your hard work! ❤️

parth391:
Hello,

How to configure template in new progress?  like we used to do in nProgress `NProgress.configure({});`

Thanks.

dreftas:
How to add `nonce-...` attribute to those inline styles ? Because these are now only ones left after configuring content security policy.

PR #2048, opened 2024-10-20
[V2] Improve WhenVisible component to merge data and params props for a more flexible API

### LINKED ISSUE #2025: [V2] Improve WhenVisible component to merge data and params props for a more flexible API
### Versions:

- `@inertiajs/react` version: v2-beta

### Describe the problem:

In the current implementation of the `WhenVisible` component, it is not possible to use both data and `params` props simultaneously. The issue lies in the `getReloadParams` function, which either returns data or params but not both. To provide a more flexible and user-friendly API, it would be ideal to merge the data prop into params when both are provided.

### Steps to reproduce:

1. Use the WhenVisible component with both data and params props.
2. The getReloadParams function will only consider data and ignore params.

```jsx
<WhenVisible fallback={"Loading..."} data={["todos", "pagy"]} params={{
  data: {
    teste: true,
    page: pagy.page + 1,
  },
  preserveUrl: true,
}} />
```
### Actual code

https://github.com/inertiajs/inertia/blob/88a765b3a3736a63b6619863b9bd872a5327f0d6/packages/react/src/WhenVisible.ts#L24

### Expected behavior:

When both data and params are provided, the function should merge them into a single object, allowing both to work together.

A initial implementation for React can be something like:

```jsx
const getReloadParams = (): Partial<ReloadOptions> => {
  const reloadParams: Partial<ReloadOptions> = { ...params }

  if (data) {
    reloadParams.only = (Array.isArray(data) ? data : [data]) as string[]
  }

  return reloadParams
}
```

PedroAugustoRamalhoDuarte:
## Enhancements

- **Merge `data` and `params` props:**  
  This PR introduces a more flexible API that merges the `data` and `params` props. Now, you can pass both in a single configuration like this:

  ```jsx
  <WhenVisible fallback="Loading..." data={["todos", "pagy"]} params={{
    data: {
      teste: true,
      page: pagy.page + 1,
    },
    preserveUrl: true,
  }} />
 ```


PedroAugustoRamalhoDuarte:
@joetannenbaum do you think this pr makes sense?

joetannenbaum:
Hey @PedroAugustoRamalhoDuarte, thanks for the PR. I'm not sure this makes sense to me to be honest.

The `data` prop is just shorthand for fetching specific data via the `only` property of `params`, if you include a `data` prop on the component _and_ include a `data` property in `params` that seems confusing to me, I wouldn't quite understand what was going on by looking at it.

If it's only `data`, it indicates "this is the data I'd like". If it's only `params`, it indicates "I'd like to get more granular about this request," which can itself include an `only` property.

As a side note, I believe children are currently optional for the component, so that should already work.

PedroAugustoRamalhoDuarte:
Thanks @joetannenbaum for the response. The major issue that I am trying to solve is that now I cannot send query params with WhenVisible component, and that is very useful for example for pagination

PedroAugustoRamalhoDuarte:
> Hey @PedroAugustoRamalhoDuarte, thanks for the PR. I'm not sure this makes sense to me to be honest.
> 
> The `data` prop is just shorthand for fetching specific data via the `only` property of `params`, if you include a `data` prop on the component _and_ include a `data` property in `params` that seems confusing to me, I wouldn't quite understand what was going on by looking at it.
> 
> If it's only `data`, it indicates "this is the data I'd like". If it's only `params`, it indicates "I'd like to get more granular about this request," which can itself include an `only` property.
> 
> As a side note, I believe children are currently optional for the component, so that should already work.

Thanks for the side note, maybe other PR already made this fix!!

PedroAugustoRamalhoDuarte:
Yes, that’s exactly it. I think a better API would merge the two, as many people naturally expect that behavior is the past:
- https://discord.com/channels/592327939920494592/915652207859351593/1323349840712699904
- https://discord.com/channels/592327939920494592/592327939920494594/1319401182954917971
- https://discord.com/channels/592327939920494592/592327939920494594/1319401656269668352

However, in the current code (modified in another PR after I created this one), I realized that the data prop is no longer necessary. We can now rely entirely on params, and it works as expected.

Thank you so much for taking the time to review this PR—I truly appreciate it!

joetannenbaum:
Closing this as it seems the conversation is resolved.

PR #2051, opened 2024-10-21
[2.x] Take over scroll restoration from browser

joetannenbaum:
This PR switches the browser's scroll restoration into manual mode, taking the process over from the browser. 

We're doing this primarily because of history encryption, as the process is async and we don't know how long it will take, we can then manually restore the document's scroll position whenever decryption has finished.


joetannenbaum:
This PR also separates out scroll data in the history state, preventing it from being encrypted with the rest of the payload as it is not sensitive data.

RomainLanz:
Hey there! 👋🏻 

It seems like this PR might have broken the SSR of a component with the latest release `2.0.1`. I'm encountering the following error during SSR rendering:

```
window is not defined
```

The error traces back to this line in the source: `node_modules/@inertiajs/core/src/eventHandler.ts:21:16`

I attempted to debug the issue locally and tried a monkey patch, but the published code is minified, making it difficult to modify or apply any meaningful changes directly.

Would it be possible to investigate this further or publish a non-minified version for easier debugging?

Thank you for your help!

RomainLanz:
(Created a new issue for that bug)

evansolomon:
We ran into this on a project that is using both inertia and react router. The fact that `history.ts` sets `window.history.scrollRestoration` at the module scope level means that if this code is imported anywhere in the dependency graph, even if history encryption is not used, the scroll restoration behavior is permanently changed. It took a while to track down the problem in back navigation for a page I was working on and once we did it was a relatively easy fix. But it does seem to me that inertia is being a little aggressive in setting this property regardless of whether the encryption feature is used. I'm curious if you've considered this and what you think.

PR #2052, opened 2024-10-22
use default empty object in useForm Vue and Svelte

### LINKED ISSUE #2011: Breaking change with empty useForm, 2.0 with svelte
### Version:

- `@inertiajs/svelte` version: 2.0.0-beta.1

### Describe the problem:

I use the `useForm` helper in my Svelte app for forms that have no data. With Inertia 1.0.16, I could just use `useForm()`, passing nothing. In 2.0, it throws unless you pass an empty object like `useForm({})`. That's no biggie obviously, it just may be worth mentioning in the upgrade docs as a breaking change if it stays this way for release :)

Now that I think about it, this might apply to React and Vue too. I haven't tested them.

### Steps to reproduce:
Just use `useForm()` anywhere in a svelte app and you'll see this error:
```
@inertiajs_svelte.js?v=74ba9348:8751 Uncaught (in promise) TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at Object.data (@inertiajs_svelte.js?v=74ba9348:8751:21)
    at @inertiajs_svelte.js?v=74ba9348:8883:59
    at Object.subscribe2 [as subscribe] (chunk-GRQJ6R46.js?v=be2c4f16:46:5)
    at useForm (@inertiajs_svelte.js?v=74ba9348:8882:10)
    at instance (BucketCard.svelte:25:25)
    at init (chunk-SGFLL5OL.js?v=be2c4f16:2190:23)
    at new BucketCard (BucketCard.svelte:22:43)
    at createProxiedComponent (svelte-hooks.js?v=be2c4f16:341:9)
    at new ProxyComponent (proxy.js?v=be2c4f16:242:7)
```

saml-dev:
Fixes #2011

This allows `useForm()` in Svelte and Vue similar to how the [React useForm](https://github.com/inertiajs/inertia/blob/master/packages/react/src/useForm.ts#L49) is set up.

NOTE I verified this works in Svelte via manual testing but I haven't tested the Vue version.

### DIFF
--- a/packages/svelte/src/useForm.ts
+++ b/packages/svelte/src/useForm.ts
@@ -57,7 +57,7 @@ export default function useForm<TForm extends FormDataType>(
   maybeData?: TForm | (() => TForm),
 ): Writable<InertiaForm<TForm>> {
   const rememberKey = typeof rememberKeyOrData === 'string' ? rememberKeyOrData : null
-  const inputData = typeof rememberKeyOrData === 'string' ? maybeData : rememberKeyOrData
+  const inputData = (typeof rememberKeyOrData === 'string' ? maybeData : rememberKeyOrData) ?? {}
   const data: TForm = typeof inputData === 'function' ? inputData() : (inputData as TForm)
   const restored = rememberKey
     ? (router.restore(rememberKey) as { data: TForm; errors: Record<keyof TForm, string> } | null)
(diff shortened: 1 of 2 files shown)


joetannenbaum:
Thank you!

PR #2060, opened 2024-10-23
[2.x] Remove `data` option from `useForm` options type

### LINKED ISSUE #2056: useForm `submit` should accept `data`
### Version:

- `@inertiajs/react` version: latest

### Describe the problem:

When using `submit`, `put`, `get`, `post`, `update` or `delete` from the `useForm` hook, the options parameter is typed as `VisitOptions`. This is because under the hood, it uses `router.visit`.

`VisitOptions` allows passing `data` as a parameter like this:

```ts
router.post(`/endpoint`, { data });
```

I think there is either a bug with the type or with the code, as it's not possible to pass `data` when we are using the `useForm` hook.

```ts
const {put} = useForm({
   foo: "bar",
});

// Below code is not working, even though the type allows it
// The data that is sent is always the one from the form, not the overwritten one
const onClick  = async () => {
   await put('/endpoint', {
      data: {
         abc: "def"
      }
   });
};
```

I would be in favor of fixing the code to allow passing additional data through this pattern, by merging anything that is passed in-here with the form data. This would allow injecting data at the exact moment the request is prepared to be sent, for cases where we need user confirmation (such as click) to set the data prior to sending.

I'll happily open a PR if you think it can be merged.

### Steps to reproduce:

- Install inertia in one of your project
- Instanciate a new React component and use the `useForm` hook with some data
- Try to pass additional data at the call moment, the hook always use data from the internal state/form

pedroborges:
The inclusion of `data` in `VisitOptions` can be misleading when using `useForm`. We should update the type definitions to exclude the `data` property in this context to prevent confusion.

Closes #2056.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -6,7 +6,7 @@ import useRemember from './useRemember'
 type setDataByObject<TForm> = (data: TForm) => void
 type setDataByMethod<TForm> = (data: (previousData: TForm) => TForm) => void
 type setDataByKeyValuePair<TForm> = <K extends keyof TForm>(key: K, value: TForm[K]) => void
-type FormDataType = object
+type FormDataType = Record<string, FormDataConvertible>
 type FormOptions = Omit<VisitOptions, 'data'>
 
 export interface InertiaFormProps<TForm extends FormDataType> {
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -26,12 +26,12 @@ export interface InertiaFormProps<TForm extends FormDataType> {
   clearErrors: (...fields: (keyof TForm)[]) => void
   setError(field: keyof TForm, value: string): void
   setError(errors: Record<keyof TForm, string>): void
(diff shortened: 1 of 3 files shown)


joetannenbaum:
Thanks, Pedro!

talelmishali:
The PR introduces breaking changes to interfaces as it changes the shape of `FormDataType` from `object` to `Record<string, FormDataConvertible>`.  

For example:

```ts
interface LoginFormProps {
    email: string;
    password: string;
    remember: boolean;
}
```

```ts
const { data, setData, post, errors, processing } = useForm<LoginFormProps>({
    email: '',
    password: '',
    remember: true,
});
```

Will now throw an error `TS2344: Type 'LoginFormProps' does not satisfy the constraint 'FormDataType'. Index signature for type 'string' is missing in type 'LoginFormProps'` caused by changing `type FormDataType = object` to `type FormDataType = Record<string, FormDataConvertible>`.

Can we maybe consider using a less restrictive type that still maintains safety but allows backward compatibility?

```ts
type FormDataType = object & {
    [K in string]?: FormDataConvertible;
}
```

PR #2070, opened 2024-10-28
Fix playground dependencies

gobeli:
The paths to the inertia-laravel dependency inside the svelte playgrounds are specified for a local setup which does not work if you don't clone the repos exactly the same way. I would suggest using the same version specified in the react playground, without local repositories.


joetannenbaum:
Good catch, sorry about that. Thanks for the PR!

PR #2074, opened 2024-11-02
[2.x] Add some additional poll helpers

grantholle:
This PR adds `toggle` and `polling()` functions to the poll object.

I think I need a little help with the React and Svelte side. It seems a little more intuitive that there's a `polling` reactive property that gets returned from the `usePoll` hooks.

### DIFF
--- a/packages/core/src/polls.ts
+++ b/packages/core/src/polls.ts
@@ -12,18 +12,12 @@ class Polls {
     interval: number,
     cb: VoidFunction,
     options: PollOptions,
-  ): {
-    stop: VoidFunction
-    start: VoidFunction
-  } {
+  ): Poll {
     const poll = new Poll(interval, cb, options)
 
     this.polls.push(poll)
 
-    return {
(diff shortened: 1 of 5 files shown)


packages/core/src/polls.ts
```diff
@@ -12,18 +12,12 @@ class Polls {
...
+  ): Poll {
     const poll = new Poll(interval, cb, options)
 
     this.polls.push(poll)
 
-    return {
-      stop: () => poll.stop(),
-      start: () => poll.start(),
-    }
+    return poll
```
grantholle, line 20: I wasn't sure why the reference gets broken here, but it's needed to keep the reactivity later down the line

joetannenbaum:
To keep the polling hook simple, I'm going to hold off on these changes for now.

SagarNaliyapara:
@joetannenbaum this is important, especially when you want to prevent multiple requests from being fired by using a conditional loading flag or something

Issue #2072, opened 2024-10-31
V2: href of Link Component does not update on change


NicolasKion:
### Versions:

- `@inertiajs/vue3` version: 2.0.0-beta.2

### Describe the problem:

It appears that the Link component does not react to updates of a <del>computed </del> any source href, computed or not

### Steps to reproduce:

```vue 
<script setup>
import {Link} from "@inertiajs/vue3";
import {computed, onMounted, onUnmounted, ref} from "@vue/runtime-core";

const val = ref('test');

// Some arbitrary computed value as source of the href
const href = computed(() => {
    return `/${val.value}`;
});

const iv = ref(0);

/*
    Updating the source of the computed value will also update the computed value
 */
onMounted(() => {
    iv.value = setInterval(() => {
        val.value = 'test' + Math.random();
        console.log(href.value);
    }, 1000);
});

onUnmounted(() => {
    clearInterval(iv.value);
}  );


// The href of the Link never updates from the initial value
</script>
<template>
    <div>
        <Link :href="href">Home</Link>
    </div>
</template>

```

MacroTom:
Exactly, am having the same issue.

Smef:
PR submitted with fix: https://github.com/inertiajs/inertia/pull/2089

Smef:
I've noticed that for regular <Link> elements everything works fine, but when using slot props the values are not properly updated. the rest of the template in the slot does react and update appropriately when the slot props data changes.

Smef:
AH! I've been able to successfully work around this by using an array syntax instead of accessing attributes directly

As an example:

Instead of 
```vue
            <template #body="{ data }">
<Link :href="'/items/'+ data.id" >
```

You can use 
```vue
            <template #body="{ data }">
<Link :href="'/items/'+ data['id']" >
```

and it seem to update the values correctly.

PR #2089, opened 2024-11-12
[2.x] Fix: make Link href prop reactive

Smef:
href values in <Link> components are not currently reactive. This resolves this issue: https://github.com/inertiajs/inertia/issues/2072


Smef:
Is there something going on with the playwright tests that I need to do? They all pass very quickly when run locally, but it was forced-stopped at 15 minutes when run here.

PR #2097, opened 2024-11-20
Don't prefetch the current page

joetannenbaum:
Here's the scenario:

- User is on `/users`
- There is a `Link` to `/users` that prefetches on hover on the page
- If they hover that `Link` it will prefetch page and refresh the data on the current page
- If there is any changed form state, etc it gets reset back to its original state

My solution is: if we are trying to prefetch the current page, just don't. I can't think of a scenario where this is useful, we're already on the page, so we just abort it before it even kicks off.

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -198,7 +198,10 @@ export class Router {
       prefetch: true,
     })
 
-    if (visit.url.href === window.location.href) {
+    const visitUrl = visit.url.origin + visit.url.pathname + visit.url.search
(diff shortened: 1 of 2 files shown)


Jussi-Mannisto:
Strict href comparison doesn't take fragments into account, e.g. `/about#contact`. These links would still get prefetched after this fix.

I used this comparison in my workaround:

```
function isCurrentPage(href) {
	const url = new URL(href, window.location.origin);
	const current = window.location;

	return url.origin + url.pathname + url.search === current.origin + current.pathname + current.search;
}
```

Issue #2103, opened 2024-11-24
global axios instance breaking with other packages that use it


LeTamanoir:
### Versions:

- `@inertiajs/core` version: 1.2.0
- `@inertiajs/react` version: 1.2.0

### Describe the problem:

If I add a custom error interceptor to the global Axios instance in my project that changes the shape of the error it will break Inertia's error handling (handling of `409 Conflict` etc...).

for example, the following code from `@notabene/pii-sdk` adds an error interceptor to the global axios instance:
https://gitlab.com/notabene/open-source/pii-sdk/-/blob/master/src/utils/create-client.ts?ref_type=heads#L42-63

I've opened an issue on their repo to ask them if they can create their own instance and not add interceptor to the global one.
But I believe this could also be done in Inertia as it also represent a security risk:
if a package is compromised and adds custom interceptor to perform malicious actions (e.g. steal credentials, etc...) inertia won't be protected and will send all of its request through it.

### Steps to reproduce:

I created an example repo: https://github.com/LeTamanoir/inertia-axios-bug

I added interceptors to the global axios instance here: https://github.com/LeTamanoir/inertia-axios-bug/blob/main/resources/js/app.tsx

then you can do the following to reproduce:
- run `npm run build`

- open the website in your browser and open the console and network tab

- add a div in https://github.com/LeTamanoir/inertia-axios-bug/blob/5f3d9542fa521007920e4d9695666fa23718f83a/resources/js/Pages/Dashboard.tsx#L13-L14 (the file is not important, we just want to trigger a new build with new file names)

- run `npm run build`

- go back to the tab with the website **without reloading it** and  click on any inertia link

normally you should see the request error with a `409 Conflict` status code, which will not be handled by inertia to trigger the full page reload as the [added interceptor](https://github.com/LeTamanoir/inertia-axios-bug/blob/5f3d9542fa521007920e4d9695666fa23718f83a/resources/js/Pages/Dashboard.tsx#L13-L14) changes the shape of the error which causes this part of the code https://github.com/inertiajs/inertia/blob/efc3cc85984e23f4924f1977861b8ed89339a56c/packages/core/src/router.ts#L418-L425 to not trigger as the `response` key is not present on the error rejected in the custom interceptor.

To fix this I think creating a local axios instance in inertia's core package should be enough, happy to submit a fix PR if you want 🙏

LeTamanoir:
here is an example PR: #2104

LeTamanoir:
[...]

PR #2104, opened 2024-11-24
feat: allow passing custom axios instance to `createInertiaApp`

LeTamanoir:


reinink:
What's the purpose of this change?

wfern:
@reinink he explains it in detail in the issue: https://github.com/inertiajs/inertia/issues/2103

pascalbaljet:
I get the problem, but changing this would be a breaking change, as I know that people are relying on the global Axios instance, for example, to set additional headers to send with each request.

If you can refactor it to something that you can pass as an option/resolver into `createInertiaApp()`, then I'm fine with it.

LeTamanoir:
Hey @pascalbaljet sorry I missed your reply, I'll update my PR with your suggestions 👍

LeTamanoir:
Hey @pascalbaljet I updated the PR.
I went with a static property as it seemed to need less code to change.
Not sure if it fits the code style of the project though 🫤 what do you think ?

chasegiunta:
@LeTamanoir why was this closed? did you happen to find an alternative?

@pascalbaljet how are people using the global axios instance it's never being exported? I've done it previously by forcing vite to pull axios from the inertia node_modules, but any package manager that uses shallow copying (bun, pnpm) will break it's importing, so it not reliant.

PR #2112, opened 2024-11-30
fix(vue/useForm/defaults): untrack before assign

### LINKED ISSUE #2109: Form isDirty always false after setting defaults
### Versions:

- `@inertiajs/vue3` version: 1.0.14

### Describe the problem:

IsDirty remains false after pushing values to an array.

### Steps to reproduce:

Set up a form:
```
const form = useForm({
    foo: []
})
```

Set the values as default:
```
form.defaults()
```

Push to the array:
```
console.log('before push', form.isDirty) // false
form.foo.push(1)
console.log('after push', form.isDirty) // false?
```

Dsaquel:
Removed the proxy layer before assign as defaults values.

Fixes #2109


joetannenbaum:
Looks good, thanks!

PR #2118, opened 2024-12-07
docs: fix incorrect SSR server start command

jordyvanderhaegen:
Hi, it seems like the contributing docs mention a non-existing command to start the SSR server.
I've replaced it with the ones mentioned in the Inertia [docs](https://inertiajs.com/server-side-rendering#running-the-ssr-server).


reinink:
Thanks!! 🙏

PR #2119, opened 2024-12-07
[2.x] feat: support react 19

jordyvanderhaegen:
Hi, this PR adds support for React 19.
Additionally, I've also included React 19 to be used in the playground environment.


jordyvanderhaegen:
Svelte tests failing seems to be unrelated.

GraxMonzo:
Let's also update this `<Link>` component, as `ref` is now a prop in React 19. https://github.com/jordyvanderhaegen/inertia/blob/2c2cd89b417fb8541028fdbe8a148765f7351eed/packages/react/src/Link.ts#L46-L76

jordyvanderhaegen:
@GraxMonzo that would require all other react versions to be dropped for the 2.x release, right?
Not sure if that's something the maintainers are up to.

Seems like `forwardRef` is indeed deprecated, so migrating away from it will have to happen at some point.
https://react.dev/reference/react/forwardRef

PR #2120, opened 2024-12-07
ci: fix cache deprecation warnings

jordyvanderhaegen:
Hi, the current build action shows a warning for usage of the deprecated `actions/cache@v2`. This PR should fix that.


reinink:
Thanks again @jordyvanderhaegen! 💪

PR #2132, opened 2024-12-12
[1.x] Restore command + click behavior on Inertia links in React

derrickreimer:
In #1910 (d53b2b3), we switched to passing `event.nativeEvent` to `shouldIntercept` in an attempt to consistently always pass native browser events to this framework-agnostic helper. This actually introduced a bug though, because `currentTarget` is _different_ on React's synthetic event vs. the underlying native event. The correct value for `currentTarget` lives on the synthetic event.

Since we are no longer relying on `event.which` (and this is the only property we used to be concerned with that was not included on the React synthetic event), we'll simply move back to passing the React synthetic event to `shouldIntercept`.

### DIFF
--- a/packages/core/src/shouldIntercept.ts
+++ b/packages/core/src/shouldIntercept.ts
@@ -1,4 +1,13 @@
-export default function shouldIntercept(event: MouseEvent | KeyboardEvent): boolean {
+// The actual event passed to this function could be a native JavaScript event
+// or a React synthetic event, so we are picking just the keys needed here (that
+// are present in both types).
+
+export default function shouldIntercept(
+  event: Pick<
+    MouseEvent,
+    'altKey' | 'ctrlKey' | 'defaultPrevented' | 'target' | 'currentTarget' | 'metaKey' | 'shiftKey' | 'button'
+  >,
+): boolean {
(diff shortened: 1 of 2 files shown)


packages/core/src/shouldIntercept.ts
```diff
@@ -1,4 +1,13 @@
-export default function shouldIntercept(event: MouseEvent | KeyboardEvent): boolean {
+// The actual event passed to this function could be a native JavaScript event
+// or a React synthetic event, so we are picking just the keys needed here (that
+// are present in both types).
+
+export default function shouldIntercept(
+  event: Pick<
+    MouseEvent,
+    'altKey' | 'ctrlKey' | 'defaultPrevented' | 'target' | 'currentTarget' | 'metaKey' | 'shiftKey' | 'button'
+  >,
```
derrickreimer, line 9: There might be a more elegant way to do this? I didn't want to introduce a dependency on React type definitions in the core package (to make this a union type), so figured the next best thing was to narrow this down to accommodate both native and synthetic event shapes. 

I confirmed that TypeScript _will_ fail if this is type is just a `MouseEvent`, since `React.MouseEvent` contains some different stuff.

joelstein:
Thanks!

PR #2134, opened 2024-12-13
Support `touchstart`, `focus`, `blur` events in link prefetch

m5r:
Add support for `touchstart`, `focus`, and `blur` events in link prefetch `hover` behavior.

This is hugely inspired from React Router's [link prefetching code](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/dom/ssr/components.tsx#L81-L172). Their event handlers composition is pretty neat too but I don't think it's necessary here.

### DIFF
--- a/packages/react/src/Link.ts
+++ b/packages/react/src/Link.ts
@@ -165,7 +165,7 @@ const Link = forwardRef<unknown, InertiaLinkProps>(
     }, prefetchModes)
 
     const regularEvents = {
-      onClick: (event) => {
+      onClick: (event: Parameters<InertiaLinkProps['onClick']>[0]) => {
         onClick(event)
 
         if (shouldIntercept(event)) {
@@ -174,9 +174,6 @@ const Link = forwardRef<unknown, InertiaLinkProps>(
           router.visit(href, visitParams)
         }
       },
-    }
-
-    const prefetchHoverEvents = {
       onMouseEnter: () => {
         hoverTimeout.current = window.setTimeout(() => {
           doPrefetch()
@@ -185,6 +182,14 @@ const Link = forwardRef<unknown, InertiaLinkProps>(
       onMouseLeave: () => {
         clearTimeout(hoverTimeout.current)
       },
+    }
+
+    const prefetchHoverEvents = {
+      onMouseEnter: regularEvents.onMouseEnter,
+      onMouseLeave: regularEvents.onMouseLeave,
+      onTouchStart: regularEvents.onMouseEnter,
+      onFocus: regularEvents.onMouseEnter,
+      onBlur: regularEvents.onMouseLeave,
       onClick: regularEvents.onClick,
     }
 
--- a/packages/svelte/src/link.ts
+++ b/packages/svelte/src/link.ts
@@ -54,22 +54,31 @@ function link(
   let baseParams: VisitOptions
   let visitParams: VisitOptions
 
-  const regularEvents: ActionEventHandlers = {
+  const regularEvents = {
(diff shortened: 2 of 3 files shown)


packages/react/src/Link.ts
```diff
@@ -165,7 +165,7 @@ const Link = forwardRef<unknown, InertiaLinkProps>(
     }, prefetchModes)
 
     const regularEvents = {
-      onClick: (event) => {
+      onClick: (event: Parameters<InertiaLinkProps['onClick']>[0]) => {
```
m5r, line 168: This looks a bit ugly but it's basically reusing the type of the first parameter of the `onClick` property defined in `InertiaLinkProps` above

packages/svelte/src/link.ts
```diff
@@ -54,22 +54,31 @@ function link(
...
+
+  const prefetchHoverEvents = {
+    mouseenter: regularEvents.mouseenter,
+    mouseleave: regularEvents.mouseleave,
+    touchstart: regularEvents.mouseenter,
+    focus: regularEvents.mouseenter,
+    blur: regularEvents.mouseleave,
     click: regularEvents.click,
-  }
+  } satisfies ActionEventHandlers
```
m5r, line 79: I replaced the type assignment with the [`satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) to make autocomplete smarter. Now typescript knows which properties from `ActionEventHandlers` are defined and which aren't while still validating the types are correct.

tests/prefetch.spec.ts
```diff
@@ -51,11 +51,11 @@ test('can prefetch using link props', async ({ page }) => {
 
   await page.getByRole('link', { name: 'On Mount' }).click()
   await isPrefetchPage(page, 2)
-  await expect(requests.requests.length).toBe(0)
+  expect(requests.requests.length).toBe(0)
```
m5r, line 54: `await`s here were not necessary, `expect().toBe()` doesn't return a promise, it's all synchronous

packages/core/src/shouldIntercept.ts
```diff
@@ -1,9 +1,9 @@
-export default function shouldIntercept(event: MouseEvent | KeyboardEvent): boolean {
+export default function shouldIntercept(event: MouseEvent | KeyboardEvent | import('react').MouseEvent<Element>): boolean {
```
joetannenbaum: Might be a silly question, but doesn't this make `react` a dependency in the core library now?

packages/react/src/Link.ts
```diff
@@ -174,9 +174,6 @@ const Link = forwardRef<unknown, InertiaLinkProps>(
           router.visit(href, visitParams)
         }
       },
-    }
-
-    const prefetchHoverEvents = {
       onMouseEnter: () => {
```
joetannenbaum, line 182: Why did you move these events into the `regularEvents`? These shouldn't be fired if they don't intend to prefetch, correct?

joetannenbaum (changes requested):
This looks cool, I just have a couple of questions here that I need clarification on. Thanks!

pascalbaljet:
@m5r Thanks for this, looks like a great PR! Could you maybe check out Joe's comments about the React import and the `mouseenter`/`mouseleave` in the `regularEvents` object? Thanks!

pascalbaljet:
I'm closing this pull request due to lack of activity. If you're still working on this or have updates to share, feel free to open a new PR. We'd be happy to take another look. Thanks!

PR #2135, opened 2024-12-14
Fix: `page.resolve.then` inline promise never resolving

Smef:
When the `<WhenVisible>` component is used with `preserveUrl: true` the promise in `page.resolve` never actually resolves since history isn't being replaced and the callback is never called.

Example:
```vue
        <WhenVisible
            always
            :params="{
                data: {
                    page: nextPageToLoad,
                },
                only: ['paginatedData'],
                preserveUrl: true,
            }"
        >
            <template #fallback>
                <div>Loading...</div>
            </template>
            <div></div>
        </WhenVisible>
```

This will get stuck forever and never resolve.

This PR adjusts `page.set` when it returns `page.resolve.then` so that the promises it creates resolve without depending on the callback if `preserveUrl` is set to true. The callback is still used if `preserveUrl` is false, since the callback is queued normally in that scenario.

### DIFF
--- a/packages/core/src/page.ts
+++ b/packages/core/src/page.ts
@@ -62,7 +62,8 @@ class CurrentPage {
       replace = replace || isSameUrlWithoutHash(hrefToUrl(page.url), location)
 
       return new Promise((resolve) => {
-        replace ? history.replaceState(page, () => resolve(null)) : history.pushState(page, () => resolve(null))
+        replace ? history.replaceState(page) : history.pushState(page)
+        resolve(null)
       }).then(() => {
         const isNewComponent = !this.isTheSame(page)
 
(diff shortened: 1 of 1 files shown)


CamKem:
@joetannenbaum - Would love it this can be merged and published in a release, can confirm it works. Having to manually work around isn't great, using the old method of hooking into `onSuccess` and updating the history after the promise has resolved... creates a bit of address bar jank.

> (Note: The playground is also sort of broken for vue3 - you will need to replace :once="false" with always in the <WhenVisible> component of the InfiniteScroll.vue page. )

As mentioned in #2140 - it is evident of this feature not working in the current committed version of the InfiniteScroll.vue component in the playground (eg. once & preserveUrl)

joetannenbaum:
Thank you for the PR! This is solving the same issue as #2163, closing this in favor of that solution, will be included in the next release.

PR #2153, opened 2024-12-25
 WhenVisible useEffect function is not recreated when params change. #2152 

### LINKED ISSUE #2152: WhenVisible useEffect function is not recreated when params change.
### Version:

- `@inertiajs/react` version: 2.0.0

### Describe the problem:

The WhenVisible component has a useEffect that makes use of the `params` prop by calling `getReloadParams`. However the prop is not in the dependency array of the useEffect resulting in changes in the params never actually being used. This makes it not possible to implement infinite scroll for instance.

### Steps to reproduce:

The observer is never recreated because the effect isn't being run because of the missing dependencies in the array.
So this won't work. The data will become stale after the first fetch.

```tsx
<WhenVisible
      fallback={"Fallback"}
      data={""}
      always
      params={{
          method: "post",
          data: {
              page: page,
          },
          onSuccess: (data) => {
              setPage((p) => p + 1);
          },
          forceFormData: true,
      }}
  >
      {/** Content */}
</WhenVisible>
```

Loeffeldude:
Closes #2152. Adds the needed dependencies to the useEffect and adds a few refs to keep the same fetching behavior as before.

### DIFF
--- a/packages/react/src/WhenVisible.ts
+++ b/packages/react/src/WhenVisible.ts
@@ -17,7 +17,7 @@ const WhenVisible = ({ children, data, params, buffer, as, always, fallback }: W
   fallback = fallback ?? null
 
   const [loaded, setLoaded] = useState(false)
-  const fetchedOnce = useRef<boolean>(false)
+  const fetchedOnceRef = useRef<boolean>(false)
   const fetchingRef = useRef<boolean>(false)
   const ref = useRef<HTMLDivElement>(null)
 
@@ -46,7 +46,7 @@ const WhenVisible = ({ children, data, params, buffer, as, always, fallback }: W
           return
         }
 
-        if (!always && fetchedOnce) {
+        if (!always && fetchedOnceRef.current) {
           observer.disconnect()
         }
 
@@ -54,7 +54,7 @@ const WhenVisible = ({ children, data, params, buffer, as, always, fallback }: W
           return
         }
 
-        fetchedOnce.current = true
+        fetchedOnceRef.current = true
         fetchingRef.current = true
 
         const reloadParams = getReloadParams()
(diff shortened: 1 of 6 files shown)


packages/react/src/WhenVisible.ts
```diff
@@ -33,40 +33,41 @@ const WhenVisible = ({ children, data, params, buffer, as, always, fallback }: W
...
+    let observer: IntersectionObserver
+    observer = new IntersectionObserver(
       (entries) => {
         if (!entries[0].isIntersecting) {
           return
         }
 
-        if (!always) {
-          observer.current?.disconnect()
+        if (!always && fetchedOnce) {
```
PedroAugustoRamalhoDuarte, line 49: maybe here we need to call .current?
```suggestion
        if (!always && fetchedOnce.current) {
```

PedroAugustoRamalhoDuarte (commented):
Thanks for the PR @Loeffeldude !!

packages/react/src/WhenVisible.ts
Loeffeldude: Oops yeah my bad. Fixed it.

joetannenbaum:
Thanks for the PR @Loeffeldude. I'm a little confused as to when the `fetching` flag was split into two separate flags and converted to `refs`. If I revert it back to a simple `useState` it seems to work as expected, want to make sure I'm not missing something.

Loeffeldude:
We split into two refs (fetchedOnceRef and fetchingRef) because we're handling two distinct problems:

Has this content ever been fetched? (for the always prop logic)
Is there currently a fetch in progress? (to prevent concurrent fetches)

Using a single useState for fetching would create problems because:

State updates are asynchronous and will trigger rerenders
The IntersectionObserver callback can fire multiple times while a fetch is in progress
When props like buffer change, the effect recreates the observer callback, which could capture stale state values

Using refs:
- Avoids unnecessary rerenders
- Provides synchronous access in the observer callback
- Prevents race conditions
- Correctly handles the always prop logic
- Maintains correct behavior when the effect needs to recreate the observer

The ref-based approach is more robust for handling edge cases and async behavior.

PR #2158, opened 2024-12-28
feat(vue3): Add Vue3 KeepAlive Component Support

cnpap:
Description: This PR adds support for Vue's KeepAlive component in Inertia.js Vue3 adapter, allowing for more efficient component caching and state preservation.

Features
- Added keepAliveResolver prop to configure KeepAlive behavior per component
- Implemented KeepAliveOptions interface with the following options:
- enabled: Toggle KeepAlive functionality
- includes: Array of component names to cache
- maxLength: Maximum number of components to keep alive (defaults to 10)
- Enhanced component key management using component names and optional URL-based IDs
- Removed unused layout ref and simplified component rendering logic


pascalbaljet:
Thanks for the PR. I've worked with `KeepAlive` and it's quite cool, but unfortunately I'll close this PR as I see multiple issues with it:

* Most importantly, I think this will cause breaking changes and unexpected behavior in existing codebases. Especially when people are already doing some kind of manual state management.
* The different packages should behave consistently, and this is a very specific feature of Vue, and I'm not sure we can guarantee that we'll find a way to implement it the exact same way in React and Svelte.
* I think the component key generation is a bit hacky, especially with the URL-based IDs.
* The PR lacks new tests and now breaks five existing tests.

PR #2161, opened 2024-12-31
Export `InertiaFormProps` in react

FlorianV85:
Hey! Necessary for typing (especially in related libraries). It was exported in previous, don't know why it isn't now. Have a nice New Year's Eve!


joetannenbaum:
Thank you! (Happy New Year!)

PR #2164, opened 2025-01-03
More resilient logic for stripping the origin from page URLs

derrickreimer:
The previous method for stripping the origin from absolute page URLs produces malformed output when the `host` string happens to appear multiple times in the URL.

```js
// Old logic
"https://example.com/foo?return_to=https%3A%2F%example.com%2Fbar".split("example.com").pop() === "%2Fbar" 

// New logic
const url = new URL("https://example.com/foo?return_to=https%3A%2F%example.com%2Fbar")
url.pathname + url.search + url.hash === "/foo?return_to=https%3A%2F%example.com%2Fbar"
```

I had some trouble getting an automated test case to work in `response.test.ts` (they all seem to failing for me when I try to run `vitest` 🤔 ), but happy to take another crack at it with some guidance.

### DIFF
--- a/packages/core/src/response.ts
+++ b/packages/core/src/response.ts
@@ -233,7 +233,7 @@ export class Response {
 
     setHashIfSameUrl(this.requestParams.all().url, responseUrl)
 
-    return responseUrl.href.split(responseUrl.host).pop()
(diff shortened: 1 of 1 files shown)


joetannenbaum:
Looks good, thank you!

PR #2166, opened 2025-01-05
Export InertiaFormProps for React

joelstein:
This is a simple tweak to export the InertiaFormProps type for React.

Nothing earth-shattering here; just doing the same thing as the Vue integration.

https://github.com/inertiajs/inertia/blob/master/packages/vue3/src/index.ts

This will also help Laravel Precognition with it's useForm() typing.

https://github.com/laravel/precognition/pull/110


joetannenbaum:
Thanks for this! Closing in favor of the merged #2161

joelstein:
Thank you!

PR #2189, opened 2025-01-15
fix: Add optional chaining to prevent history issue.

danmatthews:
We've hit a strange bug where this particular line causes an issue after using `Inertia::location()` - it tries to read the `page` parameter of `window.history.state` which i'm assuming is reset during an Inertia::location() call? 

Adding the optional chaining modifier (as already exists in another reference to this variable further up the page) would fix this for us.

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -101,7 +101,7 @@ class History {
       return Promise.resolve().then(() => {
         this.doReplaceState(
           {
-            page: window.history.state.page,
+            page: window.history.state?.page,
             scrollRegions,
           },
           this.current.url!,
@@ -115,7 +115,7 @@ class History {
       return Promise.resolve().then(() => {
         this.doReplaceState(
           {
-            page: window.history.state.page,
+            page: window.history.state?.page,
             documentScrollPosition: scrollRegion,
           },
           this.current.url!,


danmatthews:
For context, we have an issue with the Svelte adapter, `@inertiajs/svelte` versions `2.0.0` and `2.0.1` that when we use an `Inertia::location()` call, it will trigger the following console error:

image

I can't reproduce this in a very minimal reproduction repo - seems to work fine, but likely because i'm not passing large amounts of props around like we are in our production app.

Hopefully this quick fix solves it for us without impacting any compatibility for any other adapters or the core as it's so small.

victorfritz:
Same error in with our production project, thank you for your fix, hop it will be merge asap 🙏

joetannenbaum:
Thanks for the PR, I'm trying to nail down a reliable reproduction of this to add as a test so I can see what's actually going on/make sure it doesn't happen again, but I'm having trouble replicating this or any of the examples from https://github.com/inertiajs/inertia/issues/2195.

Does anyone have a clean example of producing this error?

joetannenbaum:
Closing this one in favor of #2200 as I added some tests and took a slightly different approach. Thanks @danmatthews!

PR #2191, opened 2025-01-15
Fix for scroll + back bug

joetannenbaum:
v2.0.1 introduced a bug where if you did the following sequence of events the URL wouldn't ultimately resolve to the correct final URL:

- Navigate to a another page
- Scroll down on that page
- Press the back button

This PR ensures that even when handling the pop state, our internal history object is updated with the current page. Thanks to @crynobone for the catch.

### DIFF
--- a/packages/core/src/eventHandler.ts
+++ b/packages/core/src/eventHandler.ts
@@ -77,24 +77,22 @@ class EventHandler {
       return
     }
 
-    if (history.isValidState(state)) {
-      history
-        .decrypt(state.page)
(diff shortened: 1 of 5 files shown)


nicolagianelli:
It seems that with these changes the redirects I have with anchor links don't work anymore..could it be because you removed Scroll.restoreDocument() ?
I'm redirecting user after Socialite authorization to the profile page in the "#accounts" section but now it loads the page and doesn't scroll. @joetannenbaum

PR #2198, opened 2025-01-19
Use preserveUrl with Link Component

thoresuenert:
Add preserveUrl to Link component props.


joetannenbaum:
Can you give me a use case for this?

pascalbaljet:
I'm closing this one due to the lack of response, and I also don't see a use case for this. You can use `router.reload()` if you want to revisit the current page.

Issue #2195, opened 2025-01-16
Bad developer practice causes JS exception


ipeak:
Adding `href=#` to an A tag and then scrolling after clicking triggers an Exception in FireFox (v134.0) with the following error:

```
Uncaught (in promise) TypeError: window.history.state is null
    saveScrollPositions history.ts:106
    promise callback*saveScrollPositions/< history.ts:103
```

Although its bad practice, I think it needs to be handled as it still seems quite a common unfortunately, and it took a lot of head scratching to realise what was causing the site to stop working!

danmatthews:
I have PR for this, we've had the same issue:

https://github.com/inertiajs/inertia/pull/2189

davidwessman:
We are experiencing this when doing `$inertia.reload` as well, some could be replaced by `router.poll` and it fixes the error, but we have reloads when we do e.g. pagination as well.

danmatthews:
@davidwessman are you using `replaceState: true` in those reloads? Just wondering if that causes the state to be initialised to `null` and then causes the error.

davidwessman:
Right now it looks like: 

```javascript
created() {
    this.poll();
  },

  methods: {
    poll() {
      this.$inertia.reload({
        onSuccess: (event) => {
          if (event.component === "PollingShow") {
            setTimeout(this.poll, 1000);
          }
        },
      });
    },
  },
```

joetannenbaum:
Does anyone have a clean example of producing this error? Trying to write a test that replicates the behavior and having trouble. Clean meaning: fire up a fresh Inertia app and write the simplest script possible to re-create the error.

danmatthews:
@ipddev @davidwessman can you both drop your `package.json`'s here please? Trying to see if this might be caused by something in a common dependency or something, or version specific to a certain vite version maybe.

danmatthews:
Found it, i think @joetannenbaum.

Seems to happen (for us) when doing a `router.reload` immediately in the `onMount` of a Svelte component on the page immediately after returning from an `Inertia::location()` call.

We're still using `Inertia::lazy` here, and this would likely be fixed by refactoring to `::defer()` but it's a good one to fix for backwards compatibility. My guess is likely something to do with the new async nature of requests.

Simple repro repo below, just `npm install` and `composer install` 👍🏻 

[inertia-svelte-repro.zip](https://github.com/user-attachments/files/18482254/inertia-svelte-repro.zip)

LucaRed:
[...]

PR #2200, opened 2025-01-21
Fix: Reload on mount

joetannenbaum:
This PR should fix an issue where we try to save the initial scroll state before we even have any history state.

Should fix https://github.com/inertiajs/inertia/issues/2195

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -101,6 +101,10 @@ class History {
   public saveScrollPositions(scrollRegions: ScrollRegion[]): void {
     queue.add(() => {
       return Promise.resolve().then(() => {
+        if (!window.history.state?.page) {
+          return
+        }
+
         this.doReplaceState(
           {
             page: window.history.state.page,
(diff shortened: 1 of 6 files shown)


danmatthews:
Thanks @joetannenbaum

blackjak231:
@joetannenbaum the issue seems to still be relevant on Safari iOS. Here is the line indicated by Sentry : 

Type Error: null is not an object (evaluating 'window.history.state.page')
````
{snip} 
solve().then(()=>{this.doReplaceState({page:window.history.state.page,documentScrollPosition:e},this.current.url)}))}getScrollRegions(){retu {snip}
```

However, I have no idea why it's triggering an error and I'm still investigating.

cgaube:
We are also seeing on firefox  
Screenshot 2025-01-24 at 10 06 35

https://github.com/inertiajs/inertia/blob/1b29f8aec9f07b05a11fb6e6a030b850167db700/packages/core/src/history.ts#L134

  ```
  public getScrollRegions(): ScrollRegion[] {
    return window.history.state.scrollRegions || []
  }
  ```
  
  should be 
  
    ```
  public getScrollRegions(): ScrollRegion[] {
    return window.history.state?.scrollRegions || []
  }
  ```
  
  i guess the entire file needs to be revamp a bit knowing that state can be null

PR #2206, opened 2025-01-23
[2.x] SSR clustering

RobertBoes:
Reattempt of #1909 for 2.x

Currently the Inertia SSR server would only run on a single thread, so on high load this might have implications, since the Node process could only take 100% on a single core/thread.

Node has the cluster module [nodejs.org/api/cluster.html](https://nodejs.org/api/cluster.html), which would make it possible to start multiple Node servers on the same port, then each request would be dealt with by each thread in a round-robin way.

With this PR it's possible to enable clustering by using;

```ts
createServer(
  (page) => createInertiaApp( ... ),
  {
    cluster: true,
  }
)
```

Massive thanks to @christianschoenmakers who discovered the issue and asked about it on the Inertia.js Discord server:

> Hi guys! I'm using Laravel Inertia + Vue with SSR. Sometimes the site is becoming really slow. I figured out this is because of the node SSR process is over 100% of CPU usage. However, this is only a single core, there are 7 cores idle. I want to make the node process capable of using multiple CPU cores to spread the load. What is the best way to do it? Should I spawn multiple processes with the inertia:start-ssr command, or should I implement node clustering somehow? Does anyone have a suggestion?
> [discord.com/channels/592327939920494592/758259460920573992/1255111006384820234](https://discord.com/channels/592327939920494592/758259460920573992/1255111006384820234)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5368,7 +5368,7 @@
       },
       "devDependencies": {
         "@types/deepmerge": "^2.2.0",
-        "@types/node": "^14.0",
+        "@types/node": "^18.4",
         "@types/nprogress": "^0.2.0",
         "@types/qs": "^6.9.0",
         "esbuild": "^0.16.13",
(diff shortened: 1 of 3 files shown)


taylorotwell:
@RobertBoes this is cool! Any idea what kind of performance improvements you see?

lovlyx:
> @RobertBoes this is cool! Any idea what kind of performance improvements you see?

We've tested this on one of our projects, and got a massive performance boost.

Previously requests would constantly timeout, 1 cpu core would always be at 100% usage whilst the other cores remained unused, now that it's utilizing all cores, it's easier for us to determine when we need to upgrade the server.

These are direct results from the update on Google Search Console, we use SSR for bot requests only, our average crawl response time went down by 4-3x, and continues to go down as we scale our server to handle the insane increase in crawl requests. This had immediate positive returns on our SEO.

CleanShot 2025-03-21 at 11 12 30@2x

RobertBoes:
Hey @taylorotwell, sorry for the late reply. I've stopped using Inertia's SSR implementation, because of lack of development and many issues I faced. So I can't easily give some benchmarks. But since it's currently single threaded, if many requests come in it will fully utilize only one core. If the system it's running on has, let's say, 4 core, that means 3 wouldn't be utilized, while the SSR server can't process more requests.
I recently found out PM2 can also wrap a Node script that uses the Node server and add clustering in kind of the same way. With an existing Inertia app you can run the SSR server through PM2, with `exec_mode` set to `cluster`. I believe that will give the same or similar results as this PR. See https://pm2.keymetrics.io/docs/usage/cluster-mode/

joetannenbaum:
This looks great, thanks @RobertBoes 👍

bram-pkg:
Just curious @RobertBoes, what SSR implementation are you using now instead of Inertia's? Are you still using Inertia with SSR but just a custom implementation?

PR #2210, opened 2025-01-24
chore: replace lodash to decrease bundle size

lcdss:
This PR replaces `lodash.isequal` by `@gilbarbara/deep-equal` and `lodash.clonedeep` by `klona` to decrease the bundle size in ~6kB (gzip).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -769,6 +769,12 @@
         "node": ">=18"
       }
     },
+    "node_modules/@gilbarbara/deep-equal": {
+      "version": "0.3.1",
+      "resolved": "https://registry.npmjs.org/@gilbarbara/deep-equal/-/deep-equal-0.3.1.tgz",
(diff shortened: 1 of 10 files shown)


kikyous:
I think replace lodash with es-tookit is a better solution  https://github.com/toss/es-toolkit

lcdss:
> I think replace lodash with es-tookit is a better solution https://github.com/toss/es-toolkit

I didn't know about it. I'll check how  the bundle size compares to the current choices and if it's close, I think it'll be a better alternative.

eznix86:
Thanks @lcdss, pls merge it if possible :)

lcdss:
Summarizing, `lodash` costs 8.77 kB (gzip), `es-toolkit` costs 1.84 kB (gzip) and `@gilbarbara/deep-equal` + `klona` costs 1kB (gzip). As these are generic functions that doesn't require almost any maintenance I would go with the current option, but taking popularity and security into consideration, `es-toolkit` is definitely the best option, unless we add the implementation for both functions directly to this repository.

joetannenbaum:
I think this makes sense, thanks for the PR!

PR #2181, opened 2025-01-10
[2.x]: Support for nested paths in forms

joaopalopes24:
This PR adds support for nested paths in forms: React, Vue and Svelte.

I add support for all methods: setData, reset, setErros, clearErros, defaults.

PR #2224, opened 2025-02-01
Fixed TypeError on providing only a partial subset of form fields in setErrors for useForm

### LINKED ISSUE #2214: useForm setError typing seems wrong
### Version:

- `@inertiajs/react` version: latest

### Describe the problem:

When using `useForm` in a react application, we have access to the function `setError` which has two overrides : one with a single error key and value, and another with an error object. The error object seems wrongly typed, because it assumes we always pass all key / value pair in the errors, which does not reflect the reality when used with validation libraries.

This is the current type 👇 
```ts
errors: Record<keyof TForm, string>
```

It should probably be 👇 
```ts
errors: Partial<Record<keyof TForm, string>>
```

In the same spirit, we currently create a lot of boilerplate code just to prevent sending the form when using validation librairies like zod/yup/joi etc. Would it be possible to consider accepting a resolver in the `useForm` hook configuration to prevent the sending of the form in case of validations error ? It would avoid unnecessary requests to the server and improve the DS similarly to what `react-hook-form` accepts.

Here is my current implementation with Yup as a reference 👇 

```ts
function useValidatedForm<T extends Record<string, FormDataConvertible>>(schema: Schema, initialData?: T) {
  const form = useForm<T>(initialData);

  const isValid = useCallback(
    (data: T, schema: Schema) => {
      try {
        form.clearErrors();
        schema.validateSync(data, { abortEarly: false });
        return true;
      } catch (errors: any) {
        const validationErrors: Partial<Record<keyof T, string>> = {};

        errors.inner.forEach((error: ValidationError) => {
          if (error.path !== undefined) {
            validationErrors[error.path as keyof T] = error.errors[0];
          }
        });

        form.setError(validationErrors as Record<keyof T, string>);
        return false;
      }
    },
    [form],
  );

  const submit = useCallback(
    (method: Method, url: string, options?: VisitOptions) => {
      return form.submit(method, url, {
        ...options,
        onBefore: (visit) => {
          if (!isValid(form.data, schema)) {
            return false;
          }

          if (options?.onBefore) {
            return options.onBefore(visit);
          }
        },
      });
    },
    [form, schema, isValid],
  );

  return {
    ...form,
    submit,
  };
}
```

### Steps to reproduce:

- Call `setError` with an error object containing a single key/value pair -> TS errors

Verox001:
Fixes #2214 by making the setErrors second signature partial, so not all form fields have to provided at the same time when calling `setErrors` on multiple form fields.

### Error Handling Improvements:

* [`packages/react/src/useForm.ts`](diffhunk://#diff-aff3ba12475ecdb346e7706b8abb04020a1adef9850539a8e9f7c98fcfb702b5L29-R29): Modified the `setError` method in the `InertiaFormProps` interface and the `useForm` function to accept partial records for error types. [[1]](diffhunk://#diff-aff3ba12475ecdb346e7706b8abb04020a1adef9850539a8e9f7c98fcfb702b5L29-R29) [[2]](diffhunk://#diff-aff3ba12475ecdb346e7706b8abb04020a1adef9850539a8e9f7c98fcfb702b5L219-R225)
* [`packages/svelte/src/useForm.ts`](diffhunk://#diff-59e9def88f278d4758846a9ada014927e1836d15714daad8ad9d02705bfe9f8dL39-R39): Updated the `setError` method in the `InertiaFormProps` interface and the `useForm` function to handle partial errors. [[1]](diffhunk://#diff-59e9def88f278d4758846a9ada014927e1836d15714daad8ad9d02705bfe9f8dL39-R39) [[2]](diffhunk://#diff-59e9def88f278d4758846a9ada014927e1836d15714daad8ad9d02705bfe9f8dL123-R126)
* [`packages/vue3/src/useForm.ts`](diffhunk://#diff-6a34bba1f476223d7b1c47c1b1d060324601d770f5d6c9b4a723707bd91dc728L25-R25): Changed the `setError` method in the `InertiaFormProps` interface and the `useForm` function to use partial records for error types. [[1]](diffhunk://#diff-6a34bba1f476223d7b1c47c1b1d060324601d770f5d6c9b4a723707bd91dc728L25-R25) [[2]](diffhunk://#diff-6a34bba1f476223d7b1c47c1b1d060324601d770f5d6c9b4a723707bd91dc728L111-R111)

### Example Usage and Test Update (can be reverted if not needed):

* [`playgrounds/react/resources/js/Pages/Form.tsx`](diffhunk://#diff-35feec69fdac866de37692fc93382e29577823cba800684bc36a98392fa16125R14-R18): Added an example usage of the `setError` method to demonstrate setting multiple errors at once.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -26,7 +26,7 @@ export interface InertiaFormProps<TForm extends FormDataType> {
   reset: (...fields: (keyof TForm)[]) => void
   clearErrors: (...fields: (keyof TForm)[]) => void
   setError(field: keyof TForm, value: string): void
-  setError(errors: Record<keyof TForm, string>): void
+  setError(errors: Partial<Record<keyof TForm, string>>): void
   submit: (method: Method, url: string, options?: FormOptions) => void
   get: (url: string, options?: FormOptions) => void
   patch: (url: string, options?: FormOptions) => void
@@ -216,13 +216,13 @@ export default function useForm<TForm extends FormDataType>(
   )
 
   const setError = useCallback(
-    (fieldOrFields: keyof TForm | Record<keyof TForm, string>, maybeValue?: string) => {
+    (fieldOrFields: keyof TForm | Partial<Record<keyof TForm, string>>, maybeValue?: string) => {
       setErrors((errors) => {
         const newErrors = {
           ...errors,
           ...(typeof fieldOrFields === 'string'
             ? { [fieldOrFields]: maybeValue }
-            : (fieldOrFields as Record<keyof TForm, string>)),
+            : (fieldOrFields as Partial<Record<keyof TForm, string>>)),
         }
         setHasErrors(Object.keys(newErrors).length > 0)
         return newErrors
--- a/packages/svelte/src/useForm.ts
+++ b/packages/svelte/src/useForm.ts
@@ -36,7 +36,7 @@ export interface InertiaFormProps<TForm extends FormDataType> {
   reset(...fields: (keyof TForm)[]): this
   clearErrors(...fields: (keyof TForm)[]): this
   setError(field: keyof TForm, value: string): this
-  setError(errors: Errors): this
+  setError(errors: Partial<Errors>): this
   submit(method: Method, url: string, options?: FormOptions): void
   get(url: string, options?: FormOptions): void
   post(url: string, options?: FormOptions): void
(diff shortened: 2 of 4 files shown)


Redleks:
@joetannenbaum please

joetannenbaum:
This is likely now addressed with the more robust typing offered in https://github.com/inertiajs/inertia/pull/2181, closing.

PR #2232, opened 2025-02-12
Add state null checks in history

### LINKED ISSUE #2204: [Vue3] [React] `window.history.state` is null causing errors after upgrading to 2.0.3
### Version:

- `@inertiajs/vue3` version: 2.0.3

### Describe the problem:

Before upgrading to 2.0.2 we had tests failing due to https://github.com/inertiajs/inertia/issues/2195.
Then we waited for 2.0.3 fixing the first issue, but as soon as we deployed to production we started receiving other errors of similar nature.

We might be doing something bad in these views, it seems related to some navigation but I cannot pinpoint it.

```javascript
TypeError
window.history.state is null
```

Image

### Steps to reproduce:

I cannot reproduce it,  but the errors were rolling into Sentry until we reverted.
Will try to make an example if I can :/

RSpeekenbrink:
When switching from a non-inertia handled page to an inertia page in the same application the state object is not set. Thereby the getScrollRegions() and getDocumentScrollPosition() getters would generate the following error;

```
@inertiajs_react.js?v=387df219:3538 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'scrollRegions')
```

This is fixed by adding null checks for history state.

Same goes for `saveScrollPositions` when the page object on state does not exist.

Fixes #2204

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -119,6 +119,10 @@ class History {
   public saveDocumentScrollPosition(scrollRegion: ScrollRegion): void {
     queue.add(() => {
       return Promise.resolve().then(() => {
+        if (!window.history.state?.page) {
+          return
+        }
+
         this.doReplaceState(
           {
             page: window.history.state.page,
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -131,11 +131,11 @@ class History {
   }
 
   public getScrollRegions(): ScrollRegion[] {
-    return window.history.state.scrollRegions || []
+    return window.history.state?.scrollRegions || []
   }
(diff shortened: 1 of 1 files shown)


PNardman:
👍  waiting for this, since this is blocking inertia/nova upgrades in our project.

RSpeekenbrink:
@reinink is inertiaJS development suspended until further notice about the Laravel takeover? Would be nice to get some info about what's going on ❤️

joetannenbaum:
Thank you for this PR! I got to #2265 before this one and merged it, looks like it was handled by the other PR, so I'm going to go ahead and close this one.

PR #2236, opened 2025-02-14
Make isDirty reactive to defaults

mochetts:
# Problem
The isDirty prop is not reactive as it is for the vue package.
This causes several issues such as the form staying as dirty even after a successful submit.

# Reason
The defaults are not being set as the data after a successful submission.

# Solution
Set the defaults as the submitted data upon a sucessful submit.
See vue3 implementation as it's [doing exactly that](https://github.com/inertiajs/inertia/blob/master/packages/vue3/src/useForm.ts#L174).


packages/react/src/useForm.ts
```diff
@@ -69,6 +70,10 @@ export default function useForm<TForm extends FormDataType>(
     }
   }, [])
 
+  useEffect(() => {
+    setIsDirty(!isEqual(data, defaults))
```
mochetts, line 74: "listen" to changes in data and defaults setting the `isDirty` state upon changes in any of these two

packages/react/src/useForm.ts
```diff
@@ -111,6 +116,7 @@ export default function useForm<TForm extends FormDataType>(
             setHasErrors(false)
             setWasSuccessful(true)
             setRecentlySuccessful(true)
+            setDefaults(data)
```
mochetts, line 119: This should reset the isDirty prop back to false, via the added `useEffect `.

joetannenbaum:
Thank you, looks good!

Issue #870, opened 2021-08-18
Add ability to make client-side only visits


reinink:
There are situations where making a visit to the server to set a prop is kind of unnecessary. For example, maybe you have a `/users/create` endpoint that sets a `showCreateModal` prop to `true` on the users index page.

In these situations, it would be handy if you could make a visit without hitting the server. And, it turns out you actually CAN do this today, using the `Inertia.setPage()` method, although it's quite ugly:

```js
Inertia.setPage({
  component: Inertia.page.component,
  url: '/users/create',
  props: {
    ...Inertia.page.props,
    showCreateModal: true,
  },
  version: Inertia.page.version,
}, {
  replace: true,
  preserveScroll: true,
  preserveState: true
})
```

We could make this API nicer, by creating a dedicated method for this.

```js
Inertia.push({
  // The page component name that will be used.
  // Optional, default: the current page component.
  component: 'Users',

  // The URL that the browser will be updated to.
  // Note, no actual request will be made.
  // Optional, default: the current URL.
  url: '/users/create',

  // The props for the page component.
  // The callback will receive the current page props.
  // Optional, default: {}
  props(props) {
    return {
      ...props,
      showCreateModal: true,
    }
  },

  // Indicates if it resets the scroll position.
  // Optional, default: true.
  preserveScroll: true,

  // Indicates if it resets the page state.
  // Optional, default: true.
  preserveState: true,
})
```

Which would simplify our above example to just this:

```js
Inertia.push({
  url: '/users/create',
  props: props => ({ ...props, showCreateModal: true })
})
```

And we could also use the existing deprecated `replace` method for replacing the current history state (breaking change):

```js
Inertia.replace({ ... })
```

There are also situations where you might want to make a visit to a page that uses an entirely different component:

```js
Inertia.push({
  url: '/terms-and-conditions',
  component: 'TermsAndConditions',
})
```

reinink:
Hey! Thanks so much for your interest in Inertia.js and for sharing this issue/suggestion.

In an attempt to get on top of the issues and pull requests on this project I am going through all the older issues and PRs and closing them, as there's a decent chance that they have since been resolved or are simply not relevant any longer. My hope is that with a "clean slate" me and the other project maintainers will be able to better keep on top of issues and PRs moving forward.

[...]

PR #2238, opened 2025-02-17
Improvements to history  management

alexspeller:
Before I start going deep down the rabbithole making sure this is all cleaned up and tested etc, I'd like to get some feedback on if you're likely to accept this PR.

I've had some issues with inertia's history handling during a migration to inertia with a legacy app.

The problem is that inertia assumes that _all_ history manipulation is handled with inertia's router.push / router.replace methods, which is not the case.

There are two ways this causes issues:

1. When handling popstate, it assumes that the current history state should be handled by inertia. However it the state was not pushed by inertia, history.state will not contain a valid page object and should be ignored to prevent errors
2. When preserving scroll positions, the event handler assum
es that the url stored in currentPage is the correct current url, which it might not be if some other code has pushed a new state onto the stack in the meantime.

This causes a bug that means that if you are currently on and inertia page http://example.com/foo, and run `pushState({}, "", "http://example.com/foo?bar=123")`, you end up in a situation that inertia can't handle well
	a. If you scroll the page, the scroll handler will pushState with the latest inertia it knows about, and the query param will disappear
	b. If you hit the back button, inertia will try to handle the pop state event with the empty state with no inertia page in it, and will error out trying to render a component.

I would like to fix these issues to make inertia more interoperable especially during stack migrations, so please let me know if going down this path is likely to align with the project goals and if so I'll improve the handling here to make things a little smoother

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -64,7 +64,15 @@ php artisan inertia:start-ssr
 Inertia.js uses [Playwright](https://playwright.dev/) for testing. To run the tests, use the following command:
 
 ```sh
-cd tests && npx playwright test
+npx playwright install # install playwright with system-wide browser dependencies
+
+# install packages for the test app
+cd tests/app
+npm install
+
+# run tests
+cd ..
+npx playwright test
 ```
 
 ## Publishing
--- a/packages/core/src/eventHandler.ts
+++ b/packages/core/src/eventHandler.ts
@@ -3,7 +3,7 @@ import { fireNavigateEvent } from './events'
 import { history } from './history'
 import { page as currentPage } from './page'
 import { Scroll } from './scroll'
-import { GlobalEvent, GlobalEventNames, GlobalEventResult, InternalEvent } from './types'
+import { EventHandlerInitParams, GlobalEvent, GlobalEventNames, GlobalEventResult, InternalEvent } from './types'
 import { hrefToUrl } from './url'
 
 class EventHandler {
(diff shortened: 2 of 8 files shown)


mochetts:
I had the same issue recently. 👍🏻

pascalbaljet:
Could you give me an example of how/why you are doing manual history manipulation in an Inertia app? In v2 you can use [Client side visits](https://inertiajs.com/manual-visits#client-side-visits) to change just the URL:

```js

router.replace({
    url: '/?foo=bar',
    preserveScroll: true,
    preserveState: true,
});
```

alexspeller:
I am migrating a large app slowly to inertia. It has lots of uses of pushState and replaceState already in the app. I am currently using a forked version of inertia with these changes to support this.



The client side api is great, but it doesn't help with legacy code, or 3rd party code over which we don't have control - it's fine but it would be better to not break `history.pushState` / `history.replaceState` browser apis when it's not necessary to do so. It limits the compatibility of inertia with other ecosystem components and legacy apps when migrating.

joelstein:
@pascalbaljet Here's a use case [I wrote about last year](https://github.com/inertiajs/inertia/issues/870#issuecomment-2004799193):

> We have several big data pages that include all data on the initial page load and are filtered through various form controls. These settings are mirrored to the URL so users can bookmark the link and get the same options.

pascalbaljet:
I'm still not fully sure if we should support this. Up until now, we assume that Inertia is the only package that handles the History state. If we open this up, this might lead to a ton of issues with incompatible third-party packages. For legacy code, I'd strongly suggest migrating to the official way of doing this with the [Client Side Visits](https://inertiajs.com/manual-visits#client-side-visits) feature that was introduced in v2.

But taking a technical look at this PR, could you explain how this problem was solved?

> When handling popstate, it assumes that the current history state should be handled by inertia. However it the state was not pushed by inertia, history.state will not contain a valid page object and should be ignored to prevent errors

If I understand the new code correctly, the scroll handling becomes optional. This would be a breaking change. How would you enable it again?

PR #2258, opened 2025-02-28
Fix anchor links on initial visits

reinink:
This PR fixes a bug where full page visits with anchor links (ie. `/page#someid`) don't work because the scroll restoration interferes with it.

The reason we do scroll restoration on initial page visits is to maintain the scroll position when reloading a page (see #1360). However, we don't want to do this when it's just a standard "navigate" type.

This PR adds a navigation type to see if it's a "reload" visit, and only then does scroll restoration.


taylorotwell:
Thanks!

Issue #2233, opened 2025-02-14
Query params added with history.pushState/replaceState are removed after scrolling


miDeb:
### Version:

- `@inertiajs/react` version: 2.0.0

### Describe the problem:

On one page of my application I have to use a dependencly (`@calcom/atoms`) that, when making a selection in its component, "records" this by pushing a new state through the History API. This means the url in the browser is updated and something like `&slot=2025-02-17T08%3A00%3A00.000Z` is appended
 Inertia naturally does not know about this, and after scrolling the page it itself pushes a new state to the history (to record the scroll offset), which however does not include the previous url change, causing the url to revert to a previous state. The dependency reacts badly to this in some circumstances, e.g. by immediately closing a popup it just opened.
As a workaround I removed the scroll listener Inertia adds, which fixed the issue for me. I think that's a bit of an ugly workaround, so I'm open to suggestions for a better one.

### Steps to reproduce:

Call `history.pushState(history.state, "" , "?query")` from the console on a page with Inertia, then scroll. Observe as the query parameter in the url bar disappears.

The relevant code appears to be this:
https://github.com/inertiajs/inertia/blob/fc8e1d4cccbc90b1b714087e4fbb01611375298c/packages/core/src/history.ts#L113
Here the current url (as known to Inertia) is passed which will then replace the (real) current url. Since an update to the url is not the goal here, maybe it could simply not be passed so the browser doesn't update it?

neluba:
This also happens with vue

mvinnicius22:
also having exacly the same issue, using inertiajs/vue3. Do you recommend any workaround for now? @neluba @miDeb

miDeb:
@mvinnicius22 you can make the changes from #2280 yourself locally using something like [patch-package](https://www.npmjs.com/package/patch-package) (the shipped code is minified, but if you look for the saveScrollPositions function you should be able to find the correct spot).
There might be an easier way, but that's the workaround I used.

PR #2280, opened 2025-03-21
do not pass url when storing scroll state to history

miDeb:
If there was a change to the url that interia was not aware of, such as the modification of a query parameter, this would cause the current url to be overwritten and reset to a previous value.

Since passing the url is optional, we can simply not pass it to avoid overriding the url in any case.

Closes https://github.com/inertiajs/inertia/issues/2233

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -109,8 +109,7 @@ class History {
           {
             page: window.history.state.page,
             scrollRegions,
-          },
-          this.current.url!,
+          }
         )
       })
     })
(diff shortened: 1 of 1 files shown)


jaredhobbs (approved):
LGTM, hopefully this can get approved and merged soon! We're currently relying on a yarn patch to do this but it would be much nicer to not have to do that!

mvinnicius22:
@jaredhobbs I'm also having the same problem, but using inertiajs/vue3. We need this PR :(

joetannenbaum:
Thank you, looks good!

Issue #1853, opened 2024-04-13
 "undefined" is not valid JSON at JSON.parse in createInertiaApp()


whatnotery:
### Version:

- `@inertiajs/svelte` version: 1.0.15

### Describe the problem:

I am trying to setup a rails application with inertia and svelte

I've got  the following files set-up as  per the documentation however it will not render the svelte component it errors in the console with this message:

```
Uncaught (in promise) SyntaxError: "undefined" is not valid JSON
    at JSON.parse (<anonymous>)
    at createInertiaApp (@inertiajs_svelte.js?v=5b546d2f:5036:37)
    at application.js:3:1
```

I saw this error referenced in #1358 as well although I'm not sure they're directly related if anyone has insight on how to resolve this I'd appreciate it

*app/frontend/entrypoints/application.js*
```js
import { createInertiaApp } from '@inertiajs/svelte'

createInertiaApp({
  resolve: name => {
    const pages = import.meta.glob('./Pages/**/*.svelte', { eager: true })
    return pages[`./Pages/${name}.svelte`]
  },
  setup({ el, app, props }) {
    new app({ target: el, props })
  },
})
```
*config/routes.rb*
```ruby
Rails.application.routes.draw do  
  root "static#home"
end
```

*app/controllers/static_controller.rb*
```ruby
class StaticController < ApplicationController
    def home
      render inertia: 'app'
    end
  end
```
*app/views/layouts/application.html.erb*
```erb
<!DOCTYPE html>
<head>
    <%= vite_client_tag %>
    <%= vite_javascript_tag "application", defer: true %>
</head>
<body>
    <div id="app">
    </div>
</body>
```

*app/frontend/pages/app.svelte*
```html
<h1>Hello World!</h1>
<p>This is the App component</p>
```

### Steps to reproduce:

setup a rails/inertia/svelte app
with the preceding set up

whatnotery:
This was error on my part and I have resolved this

alexmillman23:
> This was error on my part and I have resolved this

thanks for the detailed response

whatnotery:
I'm sorry I've upset so many people, I just kept trying to things till it worked out and don't fully understand what I did to resolve things I can link to the repo where I resolved the issue so people can dig into it themselves

whatnotery:
https://github.com/whatnotery/MovieFinder.club/blob/main/app/frontend/entrypoints/application.js

alexmillman23:
[...]

PR #2289, opened 2025-04-01
Allow better missed root handling

skryukov:
This PR ports existing logic from svelte adapter to the react and vue adapter to allow graceful handling for non-inertia pages:

before:
```js
const rootId = 'app'

if (document.getElementById(rootId)) {
  void createInertiaApp({
    id: rootId,
   //...
  })
}
```

after:
```js
void createInertiaApp({
  setup({ el, App, props }) {
    if (el) {...}
  }
})
```

### DIFF
--- a/packages/react/src/createInertiaApp.ts
+++ b/packages/react/src/createInertiaApp.ts
@@ -45,7 +45,7 @@ type InertiaAppOptionsForCSR<SharedProps extends PageProps> = BaseInertiaAppOpti
         includeCSS?: boolean
         showSpinner?: boolean
       }
-  setup(options: SetupOptions<HTMLElement, SharedProps>): CreateInertiaAppSetupReturnType
+  setup(options: SetupOptions<HTMLElement | null, SharedProps>): CreateInertiaAppSetupReturnType
 }
 
 type CreateInertiaAppSSRContent = { head: string[]; body: string }
@@ -76,7 +76,7 @@ export default async function createInertiaApp<SharedProps extends PageProps = P
 > {
   const isServer = typeof window === 'undefined'
   const el = isServer ? null : document.getElementById(id)
-  const initialPage = page || JSON.parse(el.dataset.page)
+  const initialPage = page || JSON.parse(el?.dataset.page || '{}')
   // @ts-expect-error
   const resolveComponent = (name) => Promise.resolve(resolve(name)).then((module) => module.default || module)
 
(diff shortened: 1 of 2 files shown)


joetannenbaum:
Looks good, thank you!

RobertBoes:
@joetannenbaum Not sure if I'm reading this change correctly, but does this mean when someone makes a mistake in their setup, it will now just silently fail as if everything works correctly? Seems like, when the element doesn't exist, or the data prop isn't present it would fallback to an empty object. This means the page would render without issues, except now the props are empty.

I've helped countless people on Discord, a fairly common issue was due to the div not rendering properly. For example by overriding the root element (https://inertiajs.com/client-side-setup#defining-a-root-element), while only being done at either the frontend or the backend. Now, with this change, there wouldn't be an error, making it extremely difficult to debug, both as an outsider helping someone or someone being new to Inertia.

skryukov:
@RobertBoes Thanks for your guidance, much appreciated!

For React, users will now see a much more descriptive error: `Target container is not a DOM element.` However, for Vue — you're absolutely right, this change causes it to silently ignore `mount(undefined)`. Instead of reverting, we can add a descriptive error or warning log when the root element is not found. wdyt?

RobertBoes:
> @RobertBoes Thanks for your guidance, much appreciated!
> 
> For React, users will now see a much more descriptive error: `Target container is not a DOM element.` However, for Vue — you're absolutely right, this change causes it to silently ignore `mount(undefined)`. Instead of reverting, we can add a descriptive error or warning log when the root element is not found. wdyt?

It would also be pretty much impossible to debug mistakes like this: https://github.com/inertiajs/inertia/issues/1853#issuecomment-2383753846
With this change it will fail silently, as the element exists. The JSON can't be parsed (since it's not present), but it falls back to an empty object. Thus the page will render without any noticeable issues, while the page props aren't populated.

skryukov:
I'd say that it's a different case—I would throw an error right away, smth like: `'Element with ID "app" has been found, but it doesn't contain the Inertia data prop. Make sure you don't have multiple DOM elements with the same ID.'`.

joetannenbaum:
Thought about this a lot, I think I didn't call this one right and I'm going to side with @RobertBoes here. The biggest gain we're getting is just moving an `if` statement from outside to inside, and I think it creates more accidental footguns for folks. 

At some point I'll remove it from the Svelte adapter as well, but that would be a breaking change, so holding off for now.

Thank you for your work on this PR @skryukov, I appreciate it.

PR #2292, opened 2025-04-04
Allow Object type for href prop

nckrtl:
The Link component has support for the href prop to be an object. However, Object type is missing from the allowed types for the href prop. Resulting in route objects not being allowed to be passed to the href prop.

For example, with wayfinder a controller action or route can not be passed directly to the href prop of Link:

```vue
<script setup lang="ts">
import { about } from "@/routes/about";
</script>

<template>
    <div>
        <Link :href="about()"> About </Link>
    </div>
</template>
```

But this does not currently work as the only allowed type for href is String. The typescript definition allows the Object type so the IDE doesn't throw an error. The error is thrown in the browser:
Screenshot 2025-04-04 at 10 04 31

### DIFF
--- a/packages/vue3/src/link.ts
+++ b/packages/vue3/src/link.ts
@@ -53,7 +53,7 @@ const Link: InertiaLink = defineComponent({
       default: () => ({}),
     },
     href: {
-      type: String,
+      type: [String, Object],
       required: true,
     },
     method: {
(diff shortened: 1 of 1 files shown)


joetannenbaum:
If you import like this:

`import about from "@/routes/about";`

Does that fix your issue? We recently adjusted the way methods were exported from Wayfinder.

joetannenbaum:
Ah, sorry, was thinking about this from a Wayfinder perspective and not an Inertia perspective. Looks good, thanks!

PR #2306, opened 2025-04-10
React component names must start with an uppercase letter.

mohamadmurad:
React component names must start with an uppercase letter.


HichemTab-tech:
Hello @mohamadmurad , usePage is considered a Hook not really a component, and hooks start usually with a lowercase i think.

mohamadmurad:
@HichemTab-tech  
thanks , but when i'm working the IDE highlight it with red line and this is the message from it

`ESLint: React Hook "usePage" is called in function "index" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use".(react-hooks/ rules-of-hooks)`

HichemTab-tech:
@mohamadmurad Actually no, this is not supposed to happen unless you are using the `usePage` where it's not supposed to be used, hooks should always be used inside components, can you share the code and the eslint config you are using please ?

Because as far as I know, the `react-hooks/rules-of-hooks` allow `usePage` to be as it is inside components. And the only times i see an error similar to yours are when i attempt to call a hook outside a component :)

mohamadmurad:
@HichemTab-tech  
Thank you again , i have use it in component but  the component name is in lowercase (index) , it should by (Index)
i have fix the problem , thanks

PR #2315, opened 2025-04-17
Update to Vite 6

SuperDJ:
- Update to Vite v6;
- Update other packages to fix several vulnerabilities;

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,17 +60,10 @@
         "node": ">=6.9.0"
       }
     },
-    "node_modules/@babel/code-frame/node_modules/js-tokens": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/js-tokens/-/js-tokens-4.0.0.tgz",
-      "integrity": "sha512-RdJUflcE3cUzKiMqQgsCu06FPu9UdIJO0beYbPhHN4k6apgJtifcoCtT9bcxOpYBtpD2kCM6Sbzg4CausW/PKQ==",
-      "dev": true,
-      "license": "MIT"
-    },
     "node_modules/@babel/compat-data": {
-      "version": "7.26.3",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.26.3.tgz",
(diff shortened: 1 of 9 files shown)


packages/svelte/test-app/Pages/Events.svelte
joetannenbaum, line 1: Why change all of the "#"?

SuperDJ: Not required for this PR but it solved a lot of [ally](https://svelte.dev/docs/svelte/compiler-warnings#a11y_invalid_attribute) issues. Since I have no experience with Svelte and the failing Svelte tests I hoped that would help fix the tests

pascalbaljet:
I'm currently working on this, but it's not as easy as the Svelte test app in v4, and v5 [requires quite some migration work](https://svelte.dev/docs/svelte/v5-migration-guide). Maybe we even need separate test apps to keep supporting v4 and v5.

Also, I think Vite is not a requirement to build the `@inertiajs/svelte` package, as `@sveltejs/package` is used for that. Not sure why that is a dev dependency as it builds perfectly fine without it.

pascalbaljet:
I've reverted the changes to the Svelte package so that we can keep building against Svelte 4 + Vite 5. Nonetheless, the Svelte test keeps failing in CI. I'm not sure why, as there are no changes now for Svelte.

pascalbaljet:
OK, it's unrelated. Running `npm upgrade` without the Vite 6 upgrade already breaks Svelte, so I'm fixing that first, and then I'll come back to this PR.

SuperDJ:
@pascalbaljet can that be related to the used node version?

pascalbaljet:
> @pascalbaljet can that be related to the used node version?

That's what I thought as well, but even running `npm upgrade` on the master branch (with no other changes) fails.

pascalbaljet:
It's a recent update of Playwright. I'll open another PR to fix that first.

pascalbaljet:
Thanks for the PR and your patience, @SuperDJ! We'll migrate the Svelte adapter later on as we don't want to drop Svelte 4 support yet.

Issue #2261, opened 2025-03-03
Deferred Props injecting styles breaks CSP


cwilsn:
### Versions:

- `@inertiajs/vue3` version: 2.0.3

### Describe the problem:

Deferred props are injecting `<style>#nprogress { display: none; }</style>` in <head> which breaks our CSP.  I've been able to work our the nprogress CSP issue but setting `includeCSS: false`[1].. but that does not prevent the CSS injection when using deferred props.

[1] https://github.com/inertiajs/progress/issues/21#issuecomment-1176231789

### Steps to reproduce:

Create a CSP with `style-src 'self'` and use a deferred prop after setting up nprogress to `includeCSS:false`

pascalbaljet:
Thanks for sharing! This has been fixed in v2.0.9 with PR #2316.

PR #2316, opened 2025-04-17
Progress: Make hide and reveal CSP-compatible

flexponsive:
Removes the use of dynamically injected <style> tags to hide and reveal the progress bar. Instead, it toggles visibility by directly modifying the display property on the element.

This avoids triggering CSP violations in environments with a strict style-src policy that disallows 'unsafe-inline'. It also slightly simplifies the code.

Related issue: https://github.com/inertiajs/inertia/issues/2261


joetannenbaum:
Looks good, thanks 👍

Issue #2310, opened 2025-04-14
Behavior of parameters with undefined changed in version 2.0.8


GulomovCreative:
### Version:

- `@inertiajs/vue3` version: 2.0.8

### Describe the problem:

In version 2.0.6 if you send undefined to data, the parameter was removed from get, but in version 2.0.8 the behavior has changed

### Steps to reproduce:

In version 2.0.6 if undefined was sent to data, the parameter was removed from get:

```ts
const data = reactive({ foo: 'bar', test: 1 })

router.reload({
  data: {
    ...data,
    foo: undefined,
  }
}) // /?test=1
```

And in version 2.0. 8 there was a change that breaks the application behavior, now the same code outputs

```ts
const data = reactive({ foo: 'bar', test: 1 })

router.reload({
  data: {
    ...data,
    foo: undefined,
  }
}) // /?foo=bar&test=1
```

CTOJoe:
As seen in [this diff](https://github.com/inertiajs/inertia/commit/2ed6671cd7288a8a0987a51a6f4bfc198241fc78#diff-5ccc166b9dbc7c5cb9816b15b8296bed6bc3363ff434a96da345e2e99d02779eL1), the method `deepmerge` was replaced with the `merge` method from `@esbuild-kit/esm-utils` in `packages/core/src/url.ts`.

Unfortunately, this new `merge` implementation does not handle explicitly undefined values in objects the same way `deepmerge` did. As noted in the second bullet point of its [documentation](https://es-toolkit.slash.page/reference/object/merge.html#merge), properties with undefined values are not overwritten (or removed in this case) during merging.

This behavioral difference results in a breaking change, particularly for cases where developers relied on explicitly undefined to remove query parameters (e.g., using router.reload({ data: { foo: undefined } }) to remove `foo` from the query string).

I've included a [StackBlitz example](https://stackblitz.com/edit/vitejs-vite-ia1xw8w8?file=src%2FApp.vue) demonstrating the issue, along with a potential fix using `withMerge` instead of `merge` to preserve the previous behavior.

I would like to propose replacing:
```js
url.search = qs.stringify(merge(qs.parse(url.search, { ignoreQueryPrefix: true }), data), {
 encodeValuesOnly: true,
 arrayFormat: qsArrayFormat,
})
```

with:
```js
url.search = qs.stringify(withMerge(qs.parse(url.search, { ignoreQueryPrefix: true }), data,
  (targetValue, sourceValue, key, target) => {
[...]

PR #2320, opened 2025-04-21
Corrects url search parameter merge logic to match behavior prior to v2.0.8

CTOJoe:
In versions prior to `v2.0.8`, `deepmerge` treated `undefined` values as a signal to remove keys from url search parameters.

So:
```js
router.reload({ data: { foo: undefined } });
```
would remove `?foo=...` from the URL entirely.

In [this commit](https://github.com/inertiajs/inertia/commit/2ed6671cd7288a8a0987a51a6f4bfc198241fc78), `deepmerge` was removed and replaced with an `es-toolkit` method of `merge` in [packages/core/src/url.ts](https://github.com/inertiajs/inertia/commit/2ed6671cd7288a8a0987a51a6f4bfc198241fc78#diff-5ccc166b9dbc7c5cb9816b15b8296bed6bc3363ff434a96da345e2e99d02779e). This new method **does not omit** `undefined` values and instead serializes them as empty strings.

So:
```js
{ foo: undefined } 
```
becomes

```js
?foo=
```

This is a subtle but breaking change in behavior and arguably a regression. 

Fixes issue #2310

### DIFF
--- a/packages/core/src/url.ts
+++ b/packages/core/src/url.ts
@@ -1,4 +1,4 @@
-import { merge } from 'es-toolkit'
+import { mergeWith } from 'es-toolkit'
 import * as qs from 'qs'
 import { hasFiles } from './files'
 import { isFormData, objectToFormData } from './formData'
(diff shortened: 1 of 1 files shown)


joetannenbaum:
@CTOJoe thanks for this PR! I'm trying to push some adjustments up to the PR and am running into permission issues:

```
ERROR: Permission to laborsync/inertia.git denied to joetannenbaum.
fatal: Could not read from remote repository.
```

Not sure if there is a configuration issue on your side?

CTOJoe:
@joetannenbaum I’ve invited you as a contributor to the repo. Since this is a public fork that will eventually be deleted once (hopefully) merged, we're fine with this approach. Hopefully, the invitation resolves the access issue.

We've never encountered this problem with a pull request before, so we'll investigate it on our end to avoid taking up more of your time. Thanks again for your consideration!

PR #2321, opened 2025-04-21
Fix for deferred props + prefetch links

joetannenbaum:
A bug surfaced when https://github.com/inertiajs/inertia/pull/2223 was merged where deferred props were reverting to their fallback state when prefetch links were hovered.

This PR fixes this behavior to only revert to fallback state if we are fetching the same URL.

### DIFF
--- a/packages/react/src/Deferred.ts
+++ b/packages/react/src/Deferred.ts
@@ -2,6 +2,17 @@ import { ReactElement, useEffect, useMemo, useState } from 'react'
 import { router } from '.'
 import usePage from './usePage'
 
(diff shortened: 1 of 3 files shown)


bradleybernard:
Not sure if it's related to this PR since I was looking at changelog, but now prefetch links (on hover for me) will show the page loading bar at the top, is that expected?

joetannenbaum:
@bradleybernard wasn't related to this, but just fixed in v2.0.10

bradleybernard:
Ah got it, thanks Joe! 🙏

PR #2324, opened 2025-04-24
feat: support form submission via FormData

josiahwiebe:
This is a proposed improvement to simplify form input handling with minimal abstraction.

I'm creating this as a draft because I don't think it's necessarily dogmatic but I see it as an improvement. 

After attending Leah's Inertia.js talk at React Miami, I found myself thinking the form input handling was a bit too verbose for some use cases. Currently, forms require individual `onChange` events for every input (updating the component state), which feels like it can be unnecessarily complex for some use-cases.

This PR adds the ability to pass the form element (target) as an option to the `form.post` method, which will then use native `FormData` to handle the payload.

For example:

```ts
function submit(e: React.FormEvent<HTMLFormElement>) {
  e.preventDefault();
  form.post('/user', {
    target: e.currentTarget,
  })
}
```

While this pattern might not be used everywhere, I believe it's valuable to offer a more native, standards-focused API for use-cases that require minimal abstraction.


chasegiunta:
I wish this could be taken one step further. Why a submit handler to begin with?

What if there was a `Form` component?
```
<Form action="/user">
  <input type="text" name="fullName">
  <input type="text" name="fields[nestedField]">
  <input type="text" name="fields[nestedArrayField][]" value="new item">
  
</Form>
```

josiahwiebe:
I think a `Form` component is probably a better approach overall. 

- Keep the submit handler for more complex scenarios
- Create a simpler `Form` component that handles `FormData` by default

pascalbaljet:
I'm closing this one as Inertia will soon come with a `Form` component ([announced at Laracon US](https://blog.laravel.com/everything-we-announced-at-laracon-us-2025)) and you can already submit a `FormData` instance with the `router`:

```js
router.post('/user', new FormData(formElement))
```

PR #2328, opened 2025-04-28
fix: React StrictMode breaking Inertia Head

### LINKED ISSUE #2253: React StrictMode breaks Inertia Head functionality.
### Version:

- `@inertiajs/react` version: 2.0.3

### Describe the problem:

When using react in `<StrictMode>...</StrictMode>`, the inertia `<Head>` component does not work correctly.

E.g. on a simple setup where we are rendering one page with a few useEffects and a single `<Head title="Profile Update" />` we are doing a `console.log` of the passed in `title` parameter to the title callback of `createInertiaApp`.

The logs are 

Image

So a few rerenders, but we ultimately end up with the title set correctly, and this is what we expect as the inertia head manager first sets it to an empty string, and the the `<Head` call sets it to our actual value.

However as soon as we enable `<StrictMode>` mode our logs look like this

Image

Resulting in us ultimately having an empty `title` parameter.

### Steps to reproduce:

In the `setup` method of createInertiaApp wrap `<App />` in `<StrictMode>` I.e.
```jsx
setup({ el, App, props }) {
  const root = createRoot(el);
  root.render(
    <StrictMode>
      <App {...props} />
    </StrictMode>
  );
},
```
Then use the `<Head title="Whatever" />` component in a layout or page.

jordanhavard:
This PR addresses an issue where the page title would temporarily disappear when using React's Strict Mode, which caused components to mount/unmount/remount during development.
The issue occurred because the provider was being disconnected during Strict Mode's temporary unmount phase. We've fixed this by:

1. Adding a reconnect() function to the head manager that allows providers to be restored
2. Calling reconnect() in the component's effect
3. Updating the dependency array to include the children prop, ensuring content updates properly

___

Fixes #2253


pascalbaljet:
Thanks!

PR #2329, opened 2025-04-29
Allow custom URL protocols

mpociot:
This PR slightly changes the host detection for URLs in order to properly support URLs that have a custom protocol.

**Why a custom protocol?**  ...you might ask.

Great question!

This will allow us to make use of Inertia within a NativePHP iOS application, where the internal URL protocol is `php://`.

With the current hardcoded check for `https?`, the URL generation ends up to become wrong when using a custom protocol.

As the regular expression still checks for the beginning of the string, I don't really see an issue with this.

### DIFF
--- a/packages/core/src/url.ts
+++ b/packages/core/src/url.ts
@@ -36,7 +36,7 @@ export function mergeDataIntoQueryString(
   data: Record<string, FormDataConvertible>,
   qsArrayFormat: 'indices' | 'brackets' = 'brackets',
 ): [string, Record<string, FormDataConvertible>] {
-  const hasHost = /^[a-zA-Z]+:\/\//.test(href.toString())
+  const hasHost = /^[a-z][a-z0-9+.-]*:\/\//i.test(href.toString())
   const hasAbsolutePath = hasHost || href.toString().startsWith('/')
   const hasRelativePath = !hasAbsolutePath && !href.toString().startsWith('#') && !href.toString().startsWith('?')
   const hasSearch = href.toString().includes('?') || (method === 'get' && Object.keys(data).length)
--- a/packages/core/src/url.ts
+++ b/packages/core/src/url.ts
@@ -36,7 +36,7 @@ export function mergeDataIntoQueryString(
   data: Record<string, FormDataConvertible>,
   qsArrayFormat: 'indices' | 'brackets' = 'brackets',
 ): [string, Record<string, FormDataConvertible>] {
-  const hasHost = /^https?:\/\//.test(href.toString())
+  const hasHost = /^[a-zA-Z]+:\/\//.test(href.toString())
   const hasAbsolutePath = hasHost || href.toString().startsWith('/')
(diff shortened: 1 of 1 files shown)


RobertBoes:
Wouldn't this cause all kinds of side-effects and potentially security issues? With this change it would allow for any kind of URL, especially concerning if you're dealing with user-generated URLs. For example it would allow all sorts of potentially really dangerous schemes:
- `resource://` - https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/resource
- `file://`
- `ftp://`
- `ws://` / `wss://`
- Along with opening any sort of app such as `slack://`

These are all schemes Inertia cannot deal with to begin with.

simonhamp:
@RobertBoes this shouldn't impact security at all. This simply determines whether the given URL has a protocol+host part to decide whether it's an absolute URL or relative one.

It doesn't change anything about how Inertia handles such URLs. Indeed, Inertia doesn't _prevent_ them being used at all, it just makes them impractical because the URLs would not be translated correctly, which could result in unexpected behaviour (which could even be considered a bug!)

With the current implementation:

```
mergeDataIntoQueryString('GET', 'ftp://mydomain.com/', {'param' => 'test'})
 => /?param=test
```

With this fix:

```
mergeDataIntoQueryString('GET', 'ftp://mydomain.com/', {'param' => 'test'})
 => ftp://mydomain.com/?param=test
```

In practice, the schemes that it's allowed to support also depend on those supported by Axios, so I think if the goal is to protect developers against themselves, then Axios (or at least how it's configured by Inertia) is the place to do that work, not here

RobertBoes:
> It doesn't change anything about how Inertia handles such URLs.
It does, as you've just now shown.

> URLs would not be translated correctly, which could result in unexpected behaviour (which could even be considered a bug!)
No, because Inertia cannot even deal with those URLs, it can't process file URLs. Without the change it would break, or at least attempt to do something sensible. With this change it will in fact attempt to process a file URL

simonhamp:
A URL is just a URL, a string that points to a resource. How the resource that it points to is handled has nothing to do with whether the URL string is absolute or relative.

In my experience, Inertia is quite capable of working with URLs with schemes other than http/s. I'm sure it would quite happily handle the _resource_ referenced by that URL if it is in the expected format

ClaraLeigh:
From a security perspective, you'd have to have a problem with xss already to have a user controlled link on the page. In which case, they could literally just do the same thing with an <a> 

Also axios itself, I don't believe it supports file:// in browsers, only in nodejs, so it's even more of a moot point. 

In terms of the code, I think the regex likely needs to support `+` symbols too. While not immediately relevant, it is used in deep links. Not something I see being needed, but might be worth adding?

Although at that point, we might as well support the full RFC 3986 spec with something like 
```
const hasProtocol = /^[a-z][a-z0-9+.-]*:\/\//i.test(url)
```

mikelmao:
I too would like to see this merged in!

joetannenbaum:
If the main concern here is user-generated content, that is not Inertia's job. That level of security is the app developer's responsibility.

I don't see an inherent issue here, if one arises I will promptly revert.

PR #2330, opened 2025-04-29
pass props as second argument to resolve() in createInertiaApp

HichemTab-tech:
# Description

This PR adds `props` as a second argument to the `resolve` function in `createInertiaApp`.

The goal is to provide more context when resolving pages which is useful for setups where you want custom loading behavior based on the props.

##  Use case

In a Laravel app using modules (like `nwidart/laravel-modules`), pages can live inside each module. With this change, you can dynamically load the correct page component based on a prop like `module`:

```js
resolve: (name, props) => {
        if (props.module) {
            return resolvePageComponent(`../../Modules/${props.module}/resources/js/pages/${name}.tsx`, {
                ...import.meta.glob('./pages/**/*.tsx'),
                ...import.meta.glob('./../../Modules/*/resources/js/pages/**/*.tsx')
            })
        }
        return resolvePageComponent(`./pages/${name}.tsx`, import.meta.glob('./pages/**/*.tsx'))
    },
```

> This is just an example of my implementation, there are a lot of use cases where this can be useful.

This way, you're not limited to a single flat `Pages` directory , you can load pages from multiple modules depending on the current route's props.

## Impact
**No impact** on existing setups

### DIFF
--- a/packages/core/src/page.ts
+++ b/packages/core/src/page.ts
@@ -6,7 +6,7 @@ import {
   Component,
   Page,
   PageEvent,
-  PageHandler,
+  PageHandler, PageProps,
   PageResolver,
   PreserveStateOption,
   RouterInitParams,
@@ -50,7 +50,7 @@ class CurrentPage {
       history.clear()
     }
 
-    return this.resolve(page.component).then((component) => {
+    return this.resolve(page.component, page.props).then((component) => {
       if (componentId !== this.componentId) {
         // Component has changed since we started resolving this component, bail
         return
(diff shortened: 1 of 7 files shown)


packages/core/src/page.ts
```diff
@@ -6,7 +6,7 @@ import {
   Component,
   Page,
   PageEvent,
-  PageHandler,
+  PageHandler, PageProps,
```
Andy9822, line 9:
```suggestion
  PageHandler, 
  PageProps,
```

Andy9822:
+1 it'd be nice to have access to the `props` in the `resolve ` function

pascalbaljet:
I understand the use case and I think I'd be a good addition. I'm going to discuss this internally and then I'll get back to you.

HichemTab-tech:
This PR was closed due to mistakenly deleted fork, here's the re-attempt #2485.

PR #2332, opened 2025-04-30
Add router configuration for modal display on non-Inertia responses

Andy9822:
## Description

This PR adds a new configuration option `showModalOnNonInertiaResponse` to the `createInertiaApp` function in all framework adapters (React, Vue, Svelte). This option allows developers to disable the automatic modal display when receiving non-Inertia responses and handle them in a custom way.

## Problem

Currently, when a non-Inertia response is received (e.g., a regular JSON response instead of an Inertia response), Inertia automatically shows a modal with the response data. While this is helpful during development for debugging, it can create a poor user experience in production if not properly handled.

## Proposed Solution

I've implemented this feature by adding a configuration option to the router. This was one approach among several possibilities, and I chose it because it seemed to fit well with the existing architecture. However, I'm open to feedback if there's a better way to implement this feature.

The implementation:
1. Adds a new `RouterConfig` type in `packages/core/src/types.ts`
2. Adds `configure` and `getConfig` methods to the `Router` class
3. Modifies the `handleNonInertiaResponse` method to check the router configuration
4. Adds the new option to all framework adapters' `createInertiaApp` functions

## Usage Example

```typescript
createInertiaApp({
  setup({ el, App, props }) {
    render(<App {...props} />, el)
  },
  showModalOnNonInertiaResponse: false, // Disable the modal
})
```

With this option disabled, developers can handle non-Inertia responses by listening to the `inertia:invalid` event:

```typescript
document.addEventListener("inertia:invalid", (event) => {
  // Custom error handling logic
  console.error('Received non-Inertia response:', event.detail.response)
})
```

## Benefits

1. **Better Production UX**: The automatic modal can be disruptive in production environments. This option allows developers to handle non-Inertia responses in a more user-friendly way.

2. **Custom Error Handling**: Developers can implement their own error handling logic based on their application's needs.

3. **Maintains Development Experience**: The modal is still shown by default, making it easy to debug during development.

## Additional Context

The modal error handling is very helpful during local development as it immediately shows what's wrong with the response. However, if this behavior accidentally makes it into production, it creates a very poor user experience. Users would see a technical error modal instead of a graceful error page or notification.

By providing this configuration option, we give developers the flexibility to handle these responses appropriately in their production environments while maintaining the helpful debugging experience during development.


packages/core/src/response.ts
```diff
@@ -90,7 +90,9 @@ export class Response {
     }
 
     if (fireInvalidEvent(response)) {
-      return modal.show(response.data)
+      if (modal.getConfig().showOnNonInertiaResponse) {
```
Andy9822, line 93: maybe this logic could be moved inside the `modal.show` so the caller doesn't need to know about this config or any business rule

packages/react/src/createInertiaApp.ts
```diff
@@ -82,6 +84,10 @@ export default async function createInertiaApp<SharedProps extends PageProps = P
 
   let head = []
 
+  if (!isServer) {
+    setupModal({ showOnNonInertiaResponse: showModalOnNonInertiaResponse })
```
Andy9822, line 88: also, if we wanted to follow the same "API" used for the progress configs, we could instead have a `modal = {}` props and then call `setupModal(modal)`

then

Andy9822: A usage example would then look like:
 
 ```jsx
 createInertiaApp({
  modal: {
    showOnNonInertiaResponse: false
  },
  progress: { 
    delay: 300,
    color: '#29d'
  },
  setup({ el, App, props }) {
    render(<App {...props} />, el)
  }
})
```

**Comment by Andy9822** [comment]
hey @joetannenbaum , do you have interest in this contribution?
I wonder if there's anything else I could add or adapt - if this looks like something you'd okay in merging

**Comment by pascalbaljet** [comment]
I think it's a good thing to have this option as this is the only place where Inertia is quite opinionated when it comes to UI. I'm not sure about the `showModalOnNonInertiaResponse` though. I opt for an `errorModal` key that defaults to `true`. That leaves the door open for a config object in the future.

Have you seen the documentation on handling errors in production? You can also redirect back (with a message) or render a dedicated error page: https://inertiajs.com/error-handling#production

**Comment by laserhybiz** [comment]
https://inertiajs.com/events#invalid:~:text=You%20may%20cancel%20the%20invalid%20event%20to%20prevent%20Inertia%20from%20showing%20the%20non%2DInertia%20response%20modal.

**Comment by pascalbaljet** [comment]
As @laserhybiz noted, this is the way to do it!

```js
import { router } from '@inertiajs/vue3'

router.on('invalid', (event) => {
  event.preventDefault()

  // Handle the invalid response yourself...
})
```

Issue #2188, opened 2025-01-15
[2.0.1] Typehinting useForm with interface gives error.


james-atlasopen:
### Version:

- `@inertiajs/react` version: 2.0.1

### Describe the problem:

Using an interface to typehint `useForm` throws a Typescript error in version 2.0.1, whereas in 2.0.0 it worked.

### Steps to reproduce:

1. Create an interface describing an object.
2. Pass it into `useForm` for the type parameter.

### Example:

```ts
  interface ConnectionData {
    url: string | null;
    host: string;
    port: string;
  }
  
  const form = useForm<ConnectionData>({
    url: connection?.url ?? "",
    host: connection?.host ?? "",
    port: connection?.port ?? "",
  });
```

Gives the following error:

TS2344: Type `ConnectionData` does not satisfy the constraint `FormDataType`. Index signature for type string is missing in type `ConnectionData`

I believe it is related to the changes made in https://github.com/inertiajs/inertia/pull/2060

selcuksarikoz:
same issue

<img width="1791" alt="image" src="https://github.com/user-attachments/assets/209088a7-ded9-4eb9-adb3-c3018ba0103a" />

Spice-King:
Bisected to 93af1711d606a0ae12fa5a0f4b74ba88ac4869f6, which is indeed part of #2060, no other adapter has changed from object to Record for the same reason.

Spice-King:
Bisected to 93af1711d606a0ae12fa5a0f4b74ba88ac4869f6, which is indeed part of #2060. The Svelte adapter has the same issue, where as the Vue 3 one does not as it still uses object.

Short term fix, define `ConnectionData` as a type, rather than an interface.  This is due to how TypeScript gives a `type` a free index key that satisfies Record. The one thing you lose is interface merging, and you are probably not using that here.
```ts
import { useForm } from '@inertiajs/react'

type ConnectionData = {
  url: string | null
  host: string
  port: string
}

declare const connection: Partial<ConnectionData> | undefined | null

const form = useForm<ConnectionData>({
  url: connection?.url ?? '',
  host: connection?.host ?? '',
  port: connection?.port ?? '',
})
```

mho22:
This worked for me : 

`packages/vue3/src/useForm.ts`

```diff
- type FormDataType = Record<string, FormDataConvertible>;
+ type FormDataType = Record<[ key : string ], FormDataConvertible>;

lcdss:
[...]

PR #1649, opened 2023-08-16
Update `useForm` to accept interfaces

Spice-King:
A differing attempt to fix #1624. Tested Vue2, Vue3 and React in the included playgrounds, worked fine (after setting the language to TS in the tested Vue2 page). `FormDataType` exists so if someone wants to make a more restrictive type, it's easier for refactors.

PR #2335, opened 2025-05-02
Revamp useForm's generic types across adaptors

Spice-King:
What started as a rework to solve allowing otherwise valid interfaces being provided directly via `useForm<SomeInterface>({...})` (fixing #2188), ended up as a total rework of the generic types they use and their uses and expanding on the work I did in #1649. Hell of a nerd snipe on myself thinking it could be solved.

All adaptors no longer have their own bespoke but functionally identical `FormDataType`, and now use one that I added to core. It should reject things that are invalid to be in the form (Functions come to mind on the typings, but anything that's not basic JSON data), which might cause some type issues for end users if they were doing something invalid to start with. If this is too tight, it can be made more permissive if that's something people want later.

Updated FormDataValues and FormDataKeys to dive down into arrays and TypeScript tuples correctly. I did quickly test all the playgrounds to make sure the adaptors use it for things like `defaults`/`setDefault`, `setData`/`setStore`, or `reset`. All of them work correctly even with arrays, with the *one* rub being that `reset` has a footgun with keys nested inside an array (if a default for the index on the array does not have a value set, it's behaviour is not well defined, ranging from setting the index to undefined or not resetting a value at all).

A type for form errors was extracted to core as `Partial<Record<FormDataKeys<TForm>, string>>` everywhere was verbose.

Relevant types have been exported on the core package for any third party client adaptors to use.

<details>
<summary>Demo of types</summary>

```ts
import { useForm } from '@inertiajs/vue3'

interface NestedType {
  id: number
  name: string
}
interface TestType {
  string: string
  number: number
  nested: NestedType
  array: NestedType[]
  tuple: [string, number, NestedType]
  typeObject: { union: 'foo' | 'bar' }
}
const form = useForm<TestType>({[...]})

form.string
//   ^? (property) TestType.string: string
form.number
//   ^? (property) TestType.number: number
form.nested
//   ^? (property) TestType.nested: NestedType
form.nested.id
//          ^? (property) NestedType.id: number
form.array[1].name
//            ^? (property) NestedType.name: string
form.tuple[0]
//         ^? (property) 0: string
form.tuple[1]
//         ^? (property) 1: number
form.tuple[2]
//         ^? (property) 2: NestedType
form.typeObject.union
//              ^? (property) union: "foo" | "bar"
form.errors['nested.name']
//           ^? (property) "nested.name"?: string
form.errors['tuple.2.name']
//          ^? (property) "tuple.2.name"?: string
form.errors['typeObject']
//          ^? (property) "typeObject"?: string
form.errors['typeObject.union']
//          ^? (property) "typeObject.union"?: string
form.reset('tuple.2.name')
form.defaults('tuple.0', 'newString')
form.defaults('tuple.1', 2)
form.defaults('tuple.2', { id: 1, name: 'Types!' })
form.defaults('typeObject.union', 'foo')
```
</details>

I expect this is not a breaking type change for either users or third party client adaptors, unless they were technically invalid to start with. I could be wrong though. The form error type might be wrong if a server side adaptor does not follow what Laravel's error bag does for nested data structures (IE: `array.0.key`), but it's not like it was being handled correctly before anyway. This is about two days of head bashing and checking done, but open to people poking holes where they can.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -10,16 +10,21 @@ declare module 'axios' {
 export type Errors = Record<string, string>
 export type ErrorBag = Record<string, Errors>
 
+export type FormDataConvertibleValue = Blob | FormDataEntryValue | Date | boolean | number | null | undefined
 export type FormDataConvertible =
   | Array<FormDataConvertible>
   | { [key: string]: FormDataConvertible }
-  | Blob
-  | FormDataEntryValue
-  | Date
-  | boolean
-  | number
-  | null
-  | undefined
+  | FormDataConvertibleValue
+
+export type FormDataType<T extends object> = {
+  [K in keyof T]: T[K] extends FormDataConvertibleValue
+    ? T[K]
+    : T[K] extends (...args: unknown[]) => unknown
+      ? never
+      : T[K] extends object | Array<unknown>
+        ? FormDataType<T[K]>
+        : never
+}
 
 export type FormDataKeys<T extends Record<any, any>> = T extends T
   ? keyof T extends infer Key extends Extract<keyof T, string>
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -60,6 +60,8 @@ export type FormDataValues<T, K extends FormDataKeys<T>> = K extends `${infer P}
       ? T[K & number]
       : never
 
+export type FormDataError<T> = Partial<Record<FormDataKeys<T>, string>>
+
 export type Method = 'get' | 'post' | 'put' | 'patch' | 'delete'
 
 export type RequestPayload = Record<string, FormDataConvertible> | FormData
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -7,7 +7,30 @@ declare module 'axios' {
   }
 }
 
-export type Errors = Record<string, string>
+export type DefaultInertiaConfig = {
+  errorValueType: string
+}
+/** 
+ * Designed to allow overriding of some core types using TypeScript
+ * interface declaration merging.
+ * 
+ * @see {@link DefaultInertiaConfig} for keys to override
+ * @example
+ * ```ts
+ * declare module '@inertiajs/core' {
+ *   export interface InertiaConfig {
+ *     errorValueType: string[]
+ *   }
+ * }
+ * ```
+ */
+export interface InertiaConfig {}
+export type InertiaConfigFor<Key extends keyof DefaultInertiaConfig> = Key extends keyof InertiaConfig
+  ? InertiaConfig[Key]
+  : DefaultInertiaConfig[Key]
+export type ErrorValue = InertiaConfigFor<'errorValueType'>
+
+export type Errors = Record<string, ErrorValue>
 export type ErrorBag = Record<string, Errors>
 
 export type FormDataConvertibleValue = Blob | FormDataEntryValue | Date | boolean | number | null | undefined
@@ -60,7 +83,7 @@ export type FormDataValues<T, K extends FormDataKeys<T>> = K extends `${infer P}
       ? T[K & number]
       : never
 
-export type FormDataError<T> = Partial<Record<FormDataKeys<T>, string>>
+export type FormDataError<T> = Partial<Record<FormDataKeys<T>, ErrorValue>>
 
 export type Method = 'get' | 'post' | 'put' | 'patch' | 'delete'
 
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -26,25 +26,39 @@ export type FormDataType<T extends object> = {
         : never
 }
 
-export type FormDataKeys<T extends Record<any, any>> = T extends T
-  ? keyof T extends infer Key extends Extract<keyof T, string>
-    ? Key extends Key
-      ? T[Key] extends Record<any, any>
-        ? `${Key}.${FormDataKeys<T[Key]>}` | Key
-        : Key
-      : never
-    : never
-  : never
+export type FormDataKeys<T> = T extends Function | FormDataConvertibleValue
+  ? never
+  : T extends Array<unknown>
+    ? number extends T['length']
+      ? `${number}` | `${number}.${FormDataKeys<T[number]>}`
+      :
+          | Extract<keyof T, `${number}`>
+          | {
+              [Key in Extract<keyof T, `${number}`>]: `${Key & string}.${FormDataKeys<T[Key & string]> & string}`
+            }[Extract<keyof T, `${number}`>]
+    :
+        | Extract<keyof T, string>
+        | {
+            [Key in Extract<keyof T, string>]: `${Key}.${FormDataKeys<T[Key]> & string}`
+          }[Extract<keyof T, string>]
 
-export type FormDataValues<T extends Record<any, any>, K extends FormDataKeys<T>> = K extends `${infer P}.${infer Rest}`
-  ? P extends keyof T
-    ? Rest extends FormDataKeys<T[P]>
-      ? FormDataValues<T[P], Rest>
+export type FormDataValues<T, K extends FormDataKeys<T>> = K extends `${infer P}.${infer Rest}`
+  ? T extends unknown[]
+    ? P extends `${infer I extends number}`
+      ? Rest extends FormDataKeys<T[I]>
+        ? FormDataValues<T[I], Rest>
+        : never
       : never
-    : never
+    : P extends keyof T
+      ? Rest extends FormDataKeys<T[P]>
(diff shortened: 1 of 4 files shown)


Spice-King:
[42fbe8c](https://github.com/inertiajs/inertia/pull/2335/commits/42fbe8c1d0527f39eb5bcb3728dd06930de8ae70) adds the ability to override the error type. This was sparked by a request from @RobertBoes to allow an array of strings from the backend, rather than just one, or an mapping of named rules to human descriptions. This allows for greater flexibility for backends.

Spice-King:
> Taking a quick look at this, shouldn't the return type of transform be a FormDataType instead of any object as it must be valid at that point to send to the server?

To be blunt, I kinda considered it out of scope, since there is no good way to deal with the non-sense people can get up to with `transform`.

> ```
>     const form = useForm({
>         bornOn: undefined as Optional<Dayjs>,
>     });
> 
>     form.transform(({ bornOn }) => ({
>         bornOn: bornOn?.format('YYYY-MM-DD'),
>     }));
> ```

This above largely why, *though*, does not full show *why*. There is nothing stopping someone from changing the shape of the object inside `transform`.

```ts
const form = useForm({ bornOn: undefined as Optional<Dayjs> });
form.transform(({ bornOn }) => ({
       myDate: bornOn?.format('YYYY-MM-DD'),
}));

form;
// ^? const form: InertiaFormProps<{ bornOn: Optional<Dayjs> }>
form.errors;
//   ^? (property) InertiaFormProps<{ bornOn: Optional<Dayjs> }>.errors: Partial<Record<"bornOn", string>>

// Validation errors no longer align with the form's own data
console.log(form.errors)
//  We end up with errors in the shape of Partial<Record<"myDate", string>>
```

`transform` can explicitly alter the data before it's sent to the server, but type wise, it can implicitly alter the data *shape* sent to the server, which has the knock on effect of the backend probably offering errors in that data shape's footprint. Personally, I think this is a valuable *escape hatch*, but otherwise something that's footgun that's impossible to handle correctly.  You are ending up with overriding stuff on the front end and the back end to keep stuff in-line, which is just messy.

If there was a way to force the `transform` method to only be valid when chained directly off of `useForm` (IE `const form = useForm({val: 1}).transform((d) => ({...d, random: Math.random()})` where `random` is replaced with something more meaningful), then I'd be able to properly type errors based off the return type of the `transform` call back. But that's just not something that Typescript offers, and neither does it allow changing a type because a function was called (totally ignoring the fact that one could call it multiple times or provide a random function out of a set to it).

> Basically making my working data 'invalid' but turning it into something valid for submission. I'm hoping there is an elegant way I can do this without sacrificing all other protections.
> 
> My first thought would be to allow module augmentation on FormDataConvertible like this:
> 
> ```
> declare namespace FormData {
>   interface ConvertibleValueMap {
>     Blob: Blob;
>     FormDataEntryValue: FormDataEntryValue;
>     Date: Date;
>     boolean: boolean;
>     number: number;
>     null: null;
>     undefined: undefined;
>   }
> }
> 
> export type FormDataConvertibleValue = FormData.ConvertibleValueMap[keyof FormData.ConvertibleValueMap];
> ```
> 
> So consumers can add their own data types that will correctly serialize (or intend to always transform) themselves
> 
> ```
> declare namespace FormData {
>   interface ConvertibleValueMap {
>     // Add custom types
>     Dayjs: Dayjs;
>   }
> }
> ```
> 
> This looks like it could be doable with this approach. There might be some better ways, but let me know what you think

That's a fair thought and all, but disagree somewhat. I don't think there should be a global way to register that a complex type is valid in a form without providing a way to register a sort of global handler to guarantee it will serialize correctly without manual intervention every time it's used with transform or something of the like. As this PR was aimed only at pure type level improvements, such a thing is out of scope for this PR in my eyes.

As an aside, is there a reason you are so fixed on keeping a DateJS instance inside a form, instead say of a Date or raw string? If I needed a enhanced Date instance (Luxon is my pick for that at least) for display or added validation reasons, I'd be using Vue's `computed` ref type to compute a clean instance of said enhanced Date, like `const bornOnValidationHelper = computed(() => dayjs(form.bornOn))` which would only be recreated when `bornOn` changes. I believe that React should offer something similar in functionality.

pascalbaljet:
> I expect this is not a breaking type change for either users or third party client adaptors, unless they were technically invalid to start with. I could be wrong though. The form error type might be wrong if a server side adaptor does not follow what Laravel's error bag does for nested data structures (IE: `array.0.key`), but it's not like it was being handled correctly before anyway. This is about two days of head bashing and checking done, but open to people poking holes where they can.

Thanks for this, @Spice-King! Just wanted to let you know that I'm reviewing it now, but it might take some time as it's quite complicated 😬

Spice-King:
> Thanks for this, @Spice-King! Just wanted to let you know that I'm reviewing it now, but it might take some time as it's quite complicated 😬

Yea, sorry about that @pascalbaljet, TS magic is like that. Will recommend going commit by commit, though a high level summery of the black magic in `FormDataValues<T>` is that it either returns `T[key]` for each key of `T` if it matches the structure what can be sent or `never` if it does not. More or less counts on `T extends never` causing TS to throw an error, making a type nested in the `TForm` stick out if invalid. `interface`s work because, when valid, it's checking against the exact same data shape, just in `type` form.

If you want to pick my brain on any given bit, I am on the Discord, rather than spamming the emails of those who are following.

nickpoulos:
Thank you for this PR, this would be a great update to the core type system. 

Just to add some extra context and comments from the peanut gallery:

I was just asking the Discord channel how they are solving issues like this.  

I solved the FormError types in a very similar way, but still the FormDataConvertible causes mismatches. 

```typescript
let { events = [] }: { events: Events; } = $props();

type FormFieldKeys = Exclude<keyof FormFields, 'errors'>;

type FormErrors = Partial<Record<FormFieldKeys, string | string[]>>;

type FormFields = {
    title: string;
    event: Event | number;
};

// Event Form Submission Request
let form = useForm<FormFields>('createEvent', {
    title: null,
    event: events[0],
    errors?: FormErrors;
});

const handleSubmit = (e: SubmitEvent) => {
        e.preventDefault();
        $form.transform((data) => {
            data.event = data.event.id;
            return data;
        }).post(route('event.save'));
    };
```

```text
Svelte: Type FormFields does not satisfy the constraint FormDataType
Property event is incompatible with index signature.
Type number | Event is not assignable to type FormDataConvertible
Type Event is not assignable to type FormDataConvertible
Type Event is not assignable to type { [key: string]: FormDataConvertible; }
Index signature for type string is missing in type Event
```

The only suitable options currently, are to accept the invalid types, not use types for useForm(), and rewrite/duplicate the FormFields type to have two versions.

pascalbaljet:
I've reviewed and tested the changes, and it's really cool how much it now supports! The only real breaking change I see is using a function as form data, but that didn't make sense in the first place, so I think that's fine.

I'll test it against real-world codebases, just to be sure, and then I think we're good to go.

Spice-King:
> The only real breaking change I see is using a function as form data, [...]

I don't believe I broke typing on that. Both Vue and Svelte's work with functions for initial data, while React never seemed to support it it typewise. Best I can see as a non-React user, the React typing on useForm was wrong to start with, since internally `useState` accepts a function for the initial state. I'll write up a quick fix for that since it's like minor bit.

> [...]  but that didn't make sense in the first place, so I think that's fine.

AFAIK, the point in using the function was for making sure page data that made it's way into the form defaults via initial data would reflect the proper data if the page data changes without out a forced recreation of the page component. It is undocumented though. Probably should fix that.

pascalbaljet:
> > The only real breaking change I see is using a function as form data, [...]
> 
> I don't believe I broke typing on that. Both Vue and Svelte's work with functions for initial data, while React never seemed to support it it typewise. Best I can see as a non-React user, the React typing on useForm was wrong to start with, since internally `useState` accepts a function for the initial state. I'll write up a quick fix for that since it's like minor bit.
> 
> > [...]  but that didn't make sense in the first place, so I think that's fine.
> 
> AFAIK, the point in using the function was for making sure page data that made it's way into the form defaults via initial data would reflect the proper data if the page data changes without out a forced recreation of the page component.

Ah sorry, I think I explained it poorly. What I meant was:

```js
const form = useForm({
  name: () => '',
})
```

Spice-King:
Ah, yes. _That_ is intentionally blocked on a type level. It was something not supported before based on the types used, with the old `FormDataType`s being `Record<string, FormDataConvertible>`. Would make typing _much_ harder in supporting it, assuming that the backing state would even accept that and convert it out.

pascalbaljet:
Thanks again for this, and for your patience! I've merged it into the `typescript-improvements` branch where I'm gathering more TypeScript PRs, which will then be merged into `master`.

PR #2344, opened 2025-05-13
Allow deepMerge on custom properties

mpociot:
This PR is the JS counterpart of https://github.com/inertiajs/inertia-laravel/pull/732

For whatever reason, the tests did time out when I tried to run them via npx for my newly created API route. I did test it manually by running the server though.
Not really sure what I did wrong.

### DIFF
--- a/packages/core/src/response.ts
+++ b/packages/core/src/response.ts
@@ -211,12 +211,18 @@ export class Response {
 
     const propsToMerge = pageResponse.mergeProps || []
     const propsToDeepMerge = pageResponse.deepMergeProps || []
+    const mergeStrategies = pageResponse.mergeStrategies || []
 
     propsToMerge.forEach((prop) => {
       const incomingProp = pageResponse.props[prop]
 
       if (Array.isArray(incomingProp)) {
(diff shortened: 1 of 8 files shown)


HichemTab-tech:
> For whatever reason, the tests did time out when I tried to run them via npx for my newly created API route. I did test it manually by running the server though.
> Not really sure what I did wrong.

Happened to me too, i just kept running the tests until it worked, i think we should increase the timeout a bit.

pascalbaljet:
Thanks @mpociot, this looks great! I'll add another test for props with multiple strategies. I expect you added support for this if you have resources like this:

```php
$data = [
    'users' => [
        ['id' => 1, 'name' => '...'],
        ['id' => 2, 'name' => '...'],
    ],
    'companies' => [
        ['id' => 1, 'name' => '...'],
        ['id' => 2, 'name' => '...'],
    ],
];

Inertia::render('Dashboard', [
    'results' => Inertia::deepMerge($data, ['users.id', 'companies.id']),
]);
```

pascalbaljet:
Updated the test in bdc5ef6 to test multiple arrays with different keys and unique keys. Thanks again!

PR #2364, opened 2025-06-04
Update default state when `setDefault()` is called right after `setData()` is called

pascalbaljet:
Fixes a state issue in React when you call `setDefault()` right after `setData()`:

```jsx
const dataAndDefaults = () => {
  form.setData('foo', [...form.data.foo, 'bar'])
  form.setDefaults()
}
```

Draft for now as it needs some more testing. Also `pendingDefaultsToDataUpdate` could use a better name perhaps 😅


pascalbaljet:
* Renamed `pendingDefaultsToDataUpdate` to `dataAsDefaults`.
* The `isDirty` prop is now memoized.
* Using `useLayoutEffect` instead of `useEffect` to prevent UI flickering of `isDirty`. I think this is fine as there are no expensive calls in the callback (no `cloneDeep`, memoization of `isDirty`).

PR #2368, opened 2025-06-06
Don't overwrite Vite class in Svelte playgrounds

jrmajor:
Currently if you run `dev` without running `build` first, you will get an error, because this version tries to read manifest even when you're in hot mode.

Looking at the `git blame`, I couldn't figure out why this was done, but removing it doesn't seem to break anything.


pascalbaljet:
@pedroborges Do you know the reason why the `Vite` class was extended? Everything seems to work fine when it's all removed from the Service Provider.

pedroborges:
@pascalbaljet that was copied from the Vue playground but I see that [it was removed](https://github.com/inertiajs/inertia/commit/2f9aa83c96acd7905b364cfd21b7600844785753) recently. I didn't run into any issues last time I ran the Svelte playgrounds but it's seems safe to remove it too.

pascalbaljet:
@pedroborges Alright, thanks for clarifying!

PR #2378, opened 2025-06-13
Reduce dependencies from qs

benmccann:
https://npmgraph.js.org/?q=qs - 18 dependencies
https://npmgraph.js.org/?q=neoqs - 0 dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1399,13 +1399,6 @@
       "dev": true,
       "license": "MIT"
     },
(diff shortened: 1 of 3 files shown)


pascalbaljet:
Thanks for the PR, but I'll close it unfortunately. I get almost the exact same build size when using neoqs, but I'm also worried about backwards compatibility, and with less than 200 stars and no commits for 10 months, I don't think it's a huge improvement over the qs repository.

Issue #2340, opened 2025-05-08
Calling router.reload at mount causes an error


DhiaBen20:
### `inertiajs/inertia-laravel` Version: "^2.0"

- `@inertiajs/react` version: "^2.0.0",

### Describe the problem:
I want to partially reload some data from the server only when the component mounts.

### Steps to reproduce:
```ts
useEffect(() => {
  router.reload({only: [key]});
}, []);
```
When I navigate to this page using client-side routing (from another page), everything works fine.
However, when I do a full page refresh, I get an error and the request is not sent to the server.

Here’s the error message:
```
@inertiajs_react.js?v=8ed0a2f8:6373 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'component')
```

This is the line of code that triggers the error:

```js
this.isPartial() && (e["X-Inertia-Partial-Component"] = s.get().component);
```

chack1172:
Fixed in #2379 @pascalbaljet

pascalbaljet:
Fixed in v2.0.13

PR #2383, opened 2025-06-18
Remove `setSwapComponent` method and cleanup after PR #2379

pascalbaljet:
This is a cleanup of PR #2379.

* `isRouterInitialized` has been renamed to `routerIsInitialized`.
* ~~The `routerIsInitialized` bool was enough to fix the problem. The additional `setSwapComponent` logic doesn't seem to provide any benefits. Perhaps this was needed before `currentIsInitialPage` [was introduced](https://github.com/inertiajs/inertia/pull/2377), but I can't reproduce it.~~
* The `setSwapComponent` has been removed to prevent introducing another public method on the core router. The swapping of the `swapComponent` method is now handled internally in `App.ts`. `swapComponent` is now simply a variable that gets replaced in the `useEffect` hook.
* There was already a test for this (*can reload on mount* in `manual-visits.spec.ts`) but in the React implementation it was using a `setTimeout` to overcome the [initial problem](https://github.com/inertiajs/inertia/issues/2340). The `setTimeout` is removed now.

### DIFF
--- a/packages/core/src/page.ts
+++ b/packages/core/src/page.ts
@@ -34,10 +34,6 @@ class CurrentPage {
     return this
   }
 
-  public setSwapComponent(swapComponent: PageHandler): void {
-    this.swapComponent = swapComponent
-  }
-
   public set(
     page: Page,
     {
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -18,7 +18,6 @@ import {
   GlobalEventResult,
   InFlightPrefetch,
   Page,
-  PageHandler,
   PendingVisit,
   PendingVisitOptions,
   PollOptions,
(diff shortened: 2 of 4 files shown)


chack1172:
@pascalbaljet it was because react was complaining about using setCurrent outside of useEffect caused by the swapComponent called at first render, so checking the variable inside fixed the problema and the setSwapComponent can be safely removed

pascalbaljet:
@chack1172 Thanks for clarifying!

pascalbaljet:
Though the tests pass, navigation is broken in the Playground with this change, so let's see if there's a solution by using `useEffect` but without `setSwapComponent`.

> Can't perform a React state update on a component that hasn't mounted yet. This indicates that you have a side-effect in your render function that asynchronously later calls tries to update the component. Move this work to useEffect instead.

pascalbaljet:
I seem to have found a solution by moving the `swapComponent` method into a `ref` and then replacing it in the `useEffect` hook 👌

pascalbaljet:
Closing this PR as it leads to other problems. Gonna have a fresh look at it.

pascalbaljet:
Moved the variables back outside of the App component again, as well as the `swapComponent` method, which now gets replaced in the `useEffect` hook.

chack1172:
It seems working fine!

PR #2385, opened 2025-06-18
Use ReactNode type for children props

chack1172:
ReactNode includes all the possible renderable types. The current definitions misses some types, this will throw type error for example:

```
<Deferred data="options" fallback={<span>Fallback</span>}
<div>first line</div>
<div>second line</div>
</Deferred>
```


pascalbaljet:
Thanks!

PR #2392, opened 2025-06-20
WIP Reload lazy load props in WhenVisible when they become undefined

### LINKED ISSUE #2388: WhenVisible broke on reload
Reloading the page or on form errors, WhenVisible breaks because the state loaded remains to true but the page prop is missing.

Example:
```
const App = () => {
    const { lazyData } = usePage().props;

    useEffect(() => {
        setTimeout(() => {
            router.reload();
        }, 5000);
    });

    return <div>
        <WhenVisible data="lazyData" fallback="Loading...">
            {lazyData.text}
        </WhenVisible>
    </div>;
}

### PR DESCRIPTION
**Pull request by chack1172**
This is an attempt to fix #2388 but it's not finished, it has some bugs.

It works when using the prop without the always attribute, but I'm struggling to make it work when it is true.

Since we cannot rely on useEffect to check if props are changed, because it will throw error before calling them, I switched the **loading** state to a ref and using an useMemo to check if props are changed checking that they are not undefined.

I then added the loading value to the useEffect so that when it change from true to false, it reload the props, but this will make it run twice, it make another request when it pass from false to true. I tried blocking it if props are loaded but this will make the always prop not work anymore.

I'm also having problem with current tests on WhenVisible component:
- The server does not send the prop back, so loading will always be false
- When using the params prop instead of passing data, the **only** key is not required so it will make all the lazy props to reset and so the loading ref will always be false. For example it is not passed in test on the fifth component:

```
<div style={{ marginTop: '6000px' }}>
        <WhenVisible
          fallback={<div>Loading fifth one...</div>}
          always
          params={{
            data: {
              count,
            },
            onSuccess: () => {
              setCount((c) => c + 1)
            },
          }}
        >
          <Foo label={`Count is now ${count}`} />
        </WhenVisible>
      </div>```


pascalbaljet:
I'm closing this one in favor of #2699.

Issue #2393, opened 2025-06-23
@inertiajs/react unable to build due to type error


fxnm:
### Version:

- `@inertiajs/react` version: 2.0.13

### Describe the problem:

If i try to build my typescript project with `strict: true` and `skipLibCheck: false` the build fails because if library internal type errors
The bugs was introduces via https://github.com/inertiajs/inertia/pull/2356

### Steps to reproduce:

1. create a typescript project with inertia
2. verify that `strict: true` and `skipLibCheck: false` are set in the `tsconfig.json`
3. build the project

Error Log:

```
$ run build

> tsc && vite build

node_modules/.pnpm/@inertiajs+react@2.0.13_react@19.1.0/node_modules/@inertiajs/react/types/useForm.d.ts:5:108 - error TS2344: Type 'TForm' does not satisfy the constraint 'Record<any, any>'.

5 export type SetDataAction<TForm> = SetDataByObject<TForm> & SetDataByMethod<TForm> & SetDataByKeyValuePair<TForm>;
                                                                                                             ~~~~~

  node_modules/.pnpm/@inertiajs+react@2.0.13_react@19.1.0/node_modules/@inertiajs/react/types/useForm.d.ts:5:27
    5 export type SetDataAction<TForm> = SetDataByObject<TForm> & SetDataByMethod<TForm> & SetDataByKeyValuePair<TForm>;
                                ~~~~~
    This type parameter might need an `extends Record<any, any>` constraint.


Found 1 error in node_modules/.pnpm/@inertiajs+react@2.0.13_react@19.1.0/node_modules/@inertiajs/react/types/useForm.d.ts:5

 ELIFECYCLE  Command failed with exit code 2.

Process finished with exit code 2
```

pascalbaljet:
I tried to reproduce it by installing the [React Starter Kit](https://github.com/laravel/react-starter-kit) with Inertia v2.0.13 and set `skipLibCheck` to `false` (`strict` is already set to `true` in the Starter Kit). Both `tsc` and `vite build` work without an error.

Could you share your `tsconfig.json`?

fxnm:
> I tried to reproduce it by installing the [React Starter Kit](https://github.com/laravel/react-starter-kit?rgh-link-date=2025-06-23T13%3A11%3A48.000Z) with Inertia v2.0.13 and set `skipLibCheck` to `false` (`strict` is already set to `true` in the Starter Kit). Both `tsc` and `vite build` work without an error.
> 
[...]

PR #2395, opened 2025-06-23
fix: fixed type error in `useForm SetDataAction` type

fxnm:
Closes: https://github.com/inertiajs/inertia/issues/2393


pascalbaljet:
Thanks!

PR #2396, opened 2025-06-24
Make Svelte's <Deferred> not crash in an SSR environment

dkulchenko:
The Deferred Svelte component calls `window.queueMicrotask`. In SSR, `window` is unavailable, leading to SSR throwing: 

```
ReferenceError: window is not defined
```

In SSR, `<Deferred>` props don't need to be rendered at all by definition, so the component should check if `window` is available before scheduling the microtask.


pascalbaljet:
Thanks!

PR #2397, opened 2025-06-24
Call `provider.update` outside useEffect block to support SSR

### LINKED ISSUE #2394: Title tag is missing with latest release
### Version:

- `@inertiajs/react` version: 2.0.13

### Describe the problem:

The `<title>` tag in my site is not producing the tags from the `<Head>` element in SSR.

It worked before this release.

When serving from SSR, and inspecting the source code, the generated HTML has a title tag like the following:

```
<title inertia=""> - Laravel</title>
```

I'm guessing that it's related to these:

https://github.com/inertiajs/inertia/issues/2253
https://github.com/inertiajs/inertia/pull/2328

pascalbaljet:
This restores the call to `provider.update` that was moved to a useEffect block in #2328.

It's no problem to call the method both inside and outside the useEffect block as the update method is debounced, so it's only executed once.

Fixes #2394

### DIFF
--- a/packages/react/src/Head.ts
+++ b/packages/react/src/Head.ts
@@ -11,6 +11,7 @@ type InertiaHead = FunctionComponent<InertiaHeadProps>
 const Head: InertiaHead = function ({ children, title }) {
   const headManager = useContext(HeadContext)
   const provider = useMemo(() => headManager.createProvider(), [headManager])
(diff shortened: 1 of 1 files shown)


pascalbaljet:
> It's no problem to call the method both inside and outside the useEffect block as the update method is debounced, so it's only executed once.

Still added the server check in be29b66

PR #2399, opened 2025-06-24
Introduce `skipNullsInQueryString` option

pascalbaljet:
This PR introduces a `skipNullsInQueryString` visit option to exclude `null` values from the query string:

```js
import { router } from '@inertiajs/react'

router.reload({ data: { foo: null } })
```

With the new option, it would make a request to `/` instead of `/?foo`. Still WIP because it misses tests, and I want to investigate if we could introduce a way to set this globally (same for `queryStringArrayFormat`).

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -340,6 +340,7 @@ export class Router {
       errorBag: '',
       forceFormData: false,
       queryStringArrayFormat: 'brackets',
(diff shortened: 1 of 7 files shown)


pascalbaljet:
Closing this one. If you want to exclude it from the query string, you should use `undefined`:

```js
router.reload({ data: { foo: undefined } })
```

PR #2401, opened 2025-06-24
Fix for `useRemember` in React

### LINKED ISSUE #2226: Issues with history navigation / useForm remember functionality
- `@inertiajs/vue3` version: 2.0.x
- `laravel/framework` version: 11.41.3
- `inertiajs/inertia-laravel` version 2.0.0

Ever since we've migrated our apps to Inertia version 2.x, we've noticed / had the following issues:

- History navigation (back button) no longer behaves like it did in Inertia ver. 1.x

For example, in a web app with a customers.index and customers.{id}.edit page, you make the following visits:
customers.index -> customers.edit (ID 2 so customers.2.edit) -> Back -> customers.edit (ID 3 so customers.3.edit)

When using Inertia 1.x, when you navigated back from the final page (customers.3.edit), you landed back on the customers.index page as expected
When using Inertia 2.x when you navigate back from the final page, you instead land on the customers.2.edit page (and upon further pressing back, it just completely skips over ever navigating you back to the customers.index page).

We've also had issues with the useForm remember functionality, because sometimes when you refresh / rebuild your app, the form no longer gets populated with the prop data, but always just populates with the data that it remembered at some point (which is then fixed with a hard refresh or multiple refreshes of the website)

Example form on our customers.edit page:

```
props: {
  customer: Object
},
data() {
  return {
    form: useForm('UpdateCustomerForm', {
        customer_type_id: this.customer.customer_type_id,
        name: this.customer.name,
        phone_number: this.customer.phone_number,
        email: this.customer.email,
        tax_payer_type_id: this.customer.tax_payer_type_id,
        tax_number: this.customer.tax_number,
        registration_number: this.customer.registration_number,
        country_id: this.customer.country_id,
        address: this.customer.address,
        postal_code: this.customer.postal_code,
        city: this.customer.city,
        notes: this.customer.notes
    })
  }
}
```

So upon refreshing / rebuilding the app, on navigation to the customers.edit page, you'd expect the form to be filled with the data from the prop, but instead of that it's filled with the data it remembered at some point (image from Vue devtools below).

Image

### Steps to reproduce:

Deploy an app with a customers.index and a customers.edit page (or similar), and then compare the navigation (via browser back button) between Inertia 1 and 2. When doing navigation from customers.index -> customers.2.edit -> back -> customers.3.edit -> back, it will skip the customers.index page, go directly to customers.2.edit page, and upon refreshing the page, the remember bug will most likely occur.

pascalbaljet:
This PR aims to fix `useRemember` in React. The following scenario is currently broken:

1. Navigate to a page which has `const [form, setForm] = useRemember({ name: '', remember: false })`
2. Call `setForm({ ...form, name: 'change-1' })` to remember some data
3. Navigate to another page with the Inertia Router and hit the browser's back button
4. The form data is successfully restored, `form.name` is still `change-1` ✅
5. Now change the data again with `setForm({ ...form, name: 'change-2' })`
6. Navigate again to another page with the Inertia Router and hit the browser's back button
7. Now `form.name` is still `change-1` instead of `change-2` ❌

Draft because I want to make sure I understand the original intention of `initialState` correctly.

Fixes #2226.

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -28,7 +28,9 @@ class History {
 
   public restore(key: string): unknown {
     if (!isServer) {
-      return this.initialState?.[this.rememberedState]?.[key]
+      return this.current[this.rememberedState]
+        ? this.current[this.rememberedState]?.[key]
+        : this.initialState?.[this.rememberedState]?.[key]
     }
   }
 
--- a/tests/remember.spec.ts
+++ b/tests/remember.spec.ts
@@ -73,6 +73,18 @@ test.describe('Remember (local state caching)', () => {
     await expect(page.locator('.b-name')).toHaveValue('B1')
     await expect(page.locator('.b-remember')).toBeChecked()
     await expect(page.locator('.b-untracked')).toHaveValue('')
+
+    // try again with updated values
+    await page.fill('#name', 'E')
+    await page.uncheck('#remember')
+
+    await page.getByRole('link', { name: 'Navigate away' }).click()
+    await shouldBeDumpPage(page, 'get')
+    await page.goBack()
+
+    await expect(page).toHaveURL('remember/multiple-components')
+    await expect(page.locator('#name')).toHaveValue('E')
+    await expect(page.locator('#remember')).not.toBeChecked()
   })
 
   test('restores remembered data when pressing the back button from another website', async ({ page }) => {


sheriffderek:
I believe this isn't React-specific. [packages/core/src/history.ts](https://github.com/inertiajs/inertia/pull/2401/commits/42ff3abd346cda408e5b13aced2cf4d278f7aa1d#diff-57f4958525f8191350650773493192bf1bb0a3933bc41d9134e4e6279394b178) and would solve for Vue as well

sheriffderek:
https://github.com/user-attachments/assets/8f286d89-30af-45df-ad89-bb88142e2daa

I patched with this -- and it solves for Vue as well.

```js
const form = useForm('uniqueKeyPageNameEtc', {
    full_address: props.initialListingData?.full_address ?? '',
    // in our case - initially using temp session data
});
```

pascalbaljet:
Oh, that's a nice side effect because I mainly focused on React. The updated test in d8adf8a only failed in React, so I'll check tomorrow if I can add a failing test for Vue as well (before the patch).

pascalbaljet:
For reference: the `initialState` variable was introduced in PR #2023

pascalbaljet:
I've added additional tests that would fail on Svelte and Vue before the fix. 

Also tried to refactor `initialState` away, but that breaks `deferred-props.spec.ts` (*we are not caching deferred props after reload*) and `remember.spec.ts` (*restores remembered data when pressing the back button from another website*).

PR #2405, opened 2025-06-25
Add event callbacks to `ClientSideVisitOptions`

### LINKED ISSUE #2325: There is no way to react to finished client visits (`replace=true`)
### Versions:

- `@inertiajs/core` version: 2.0.4
- `@inertiajs/vue2` version: #.#.#
- `@inertiajs/vue3` version: 2.0.4
- `@inertiajs/react` version: #.#.#
- `@inertiajs/svelte` version: #.#.#

### Describe the problem:

There is currently no way to know, when a client visit is done (with `replace: true`)
It uses promises internally but those are not exposed. Neither an event is fired.
- `popstate` only fires on navigation (not when `replaceState` or `pushState` is used).
- `insertia:finish` only fires when doing server visits
- `inertia:navigate` only fires when `replace: false`

### Steps to reproduce:

Hard to come up with a reproduction that shows that nothing happens :D

```js
document.addEventListener('inertia:finish', console.log)
document.addEventListener('inertia:navigate', console.log)
window.addEventListener('popstate', console.log)

router.replace({
  preserveScroll: true,
  preserveState: true,
  props: {
    foo: 5
  }
})
```

### Possible solutions

Imo, the best solution by far would be to just return the promises by returning it here:

https://github.com/inertiajs/inertia/blob/4efe6826ddcac89f090091f85077505d0de25424/packages/core/src/router.ts#L290-L307

and here:

https://github.com/inertiajs/inertia/blob/4efe6826ddcac89f090091f85077505d0de25424/packages/core/src/router.ts#L282-L288

Another solution could be adding yet another event. However, that is a bit weird to use when you only want to react to one specific router visit. Hence I think that exposing the promises is the superior solution

pascalbaljet:
This PR introduces the ability to pass an `onError`, `onFinish`, and `onSuccess` callback when making client-side visits:

```js
router.replace({
  props: (props) => ({ ...props, foo: 'bar' }),
  onError: (errors) => {
    // ...
  },
  onFinish: () => {
    // ...
  },
  onSuccess: () => {
    // ...
  }
});
```

Fixes #2325.

### DIFF
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -297,20 +297,33 @@ export class Router {
 
     const props = typeof params.props === 'function' ? params.props(current.props) : (params.props ?? current.props)
 
-    const { onSuccess, ...pageParams } = params
+    const { onError, onFinish, onSuccess, ...pageParams } = params
 
-    currentPage.set(
-      {
-        ...current,
-        ...pageParams,
-        props,
-      },
-      {
-        replace,
-        preserveScroll: params.preserveScroll,
-        preserveState: params.preserveState,
-      },
-    ).then(() => onSuccess?.())
+    currentPage
+      .set(
+        {
+          ...current,
+          ...pageParams,
+          props,
+        },
+        {
+          replace,
+          preserveScroll: params.preserveScroll,
+          preserveState: params.preserveState,
+        },
+      )
+      .then(() => {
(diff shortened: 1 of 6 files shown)


pascalbaljet:
I've also added support for `onError` and `onFinish`, and an `errorBag` parameter.

Issue #2411, opened 2025-06-27
useForm - reset the form in onSucces shows unexpected behavior


thoresuenert:
### Version:

- `@inertiajs/react` version: 2.0.11

### Describe the problem:

I have a create form in a sheet which stays mounted after submit, i manage open/close state manually for different purposes.
After Submit in `onSuccess` i want to reset the form to the initial provided defaults (clear the form), but the default behavior is an overwrite of the defaults to the submitted data of the form.
This behavior makes totally sense because most of the time an edit form stays on the page and a create form redirect to the new data.

To opt out i provide a `form.reset()` in the `onSuccess` callback to reset the form, but the hook runs to late.

### Steps to reproduce:
```js
let form = useForm({ default: 'value'});

form.post('/some-url', {
 onSuccess: (page) => form.reset();
})
```

PR #2412, opened 2025-06-27
Allow to reset the form after successful submit and sync the defaults afterwards

thoresuenert:
Provide a Fix to #2411.

After digging into issues and code i found the following [comment from Jess Archer](https://github.com/inertiajs/inertia/issues/553#issuecomment-1700217765) which describe the steps to reproduce, but for the vue form.
Which leads me to the form helper of vue and i found different behavior vue vs react.

The critical `setDefaults(data)` in `onSuccess` was introduce by https://github.com/inertiajs/inertia/pull/2236 to fix an issue with reactivity of the isDirty prop.

### Solution:
To bring vue and react in sync an solve this bug, we have to move the `setDefaults` after running the `onSuccess` hook, which bring the reset data and the defaults in sync.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -129,7 +129,6 @@ export default function useForm<TForm extends FormDataType>(
             setHasErrors(false)
             setWasSuccessful(true)
             setRecentlySuccessful(true)
-            setDefaults(cloneDeep(data))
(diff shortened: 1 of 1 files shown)


pascalbaljet:
I'm closing this one in favor of #2539. It's more or less the same fix, but the implementation is a little different, and there are some additional fixes and tests.

Issue #2285, opened 2025-03-27
router.reload with data containing arrays gets accumulated instead of replaced


HassanDomeDenea:
### Version:

- `@inertiajs/vue3` version: #.#.#

### Describe the problem:

When using router.reload, with preserveUrl is false, and providing data that contains array. it works for first time when the current url query does not containing the array key, but next times it keep adding values to array instead of replacing it.

### Steps to reproduce:

```JS
router.reload({
    data: {
      sorts: [{field: 'id', direction: 'asc',},{field: 'created_at', direction: 'asc',}]
    },
  })
```

At first execute of the above method, when current url query is empty, the result will be:
```
http://localhost:3333/index?sorts[][field]=id&sorts[][direction]=asc
```

Which is fine, but when it get executed again, whether with same data or different one, the result will be (if same data):
```
http://localhost:3333/index?sorts[][field]=id&sorts[][direction]=asc&sorts[][field]=id&sorts[][direction]=asc
```

### Expected results:
when an array is provided in the data object of router.reload, it should replace the existing value, not accumulating to it.

### Current workaround:

using router.visit instead, like : 

```JS
router.visit(window.location.pathname,{
      preserveUrl: false,
      preserveState: true,
      method: 'get',
      replace: true,
      data: {
        sorts: [{field: 'id', direction: 'asc',},{field: 'created_at', direction: 'asc',}]
      },
```

HassanDomeDenea:
#1709 
Same as this form v1

aeruggiero:
Hello, I'm facing the same issue now.
In order to have a reusable component, router.reload() is something I need, and now I need to use router.get()
```
router.get(
    route(route().current() as string, {
      ...route().params,
      _query: { ...props.filters, rows: pagination.value.pageSize, page: pagination.value.pageIndex + 1 },
    }),
    {},
    {
      preserveState: true,
      preserveScroll: true,
      replace: true,
      only: ['players'],
      onFinish: () => (loading.value = false),
    },
  );
```

Hopefully this will be fixed soon!

mfeitoza:
Same issue for Svelte.

pascalbaljet:
I've opened PR #2417 to fix this.

pascalbaljet:
I'll close this issue as the PR is merged.

PR #2416, opened 2025-06-28
ensure router.reload data replaces query params

HamidRezaLS:
PR fixes a bug where `router.reload({ data: { ... } })` would incorrectly merge new data with existing query string parameters instead of replacing them.

**The Problem:**
The old behavior was unpredictable and led to incorrect state. During local testing, I observed the following scenario:
1.  A page loads with a prop, for example `items: [1, 2]`.
2.  A reload is triggered with `data: { items: [3, 4] }`. The props correctly update to `['3', '4']`.
3.  A second reload is then triggered with `data: { items: [99] }`.
4.  **The Bug:** The props incorrectly become `['99', '4']`, because the `items[1]=4` parameter from the previous URL was not cleared during the new request.

This change modifies how `GET` requests handle the `data` option. The provided `data` object is now treated as the single source of truth for the new query string, completely overwriting any old parameters.

For developers using Inertia, this makes the `router.reload` function significantly more predictable and easier to reason about, which helps prevent unexpected behavior in applications.

I've also confirmed that this does not change or break the documented behavior of partial reloads. This fix only affects how the request URL is constructed in `url.ts`. The client-side prop merging that handles `only` and `except` is a separate process and remains untouched.

Fixes: #2285


pascalbaljet:
Thanks for this! Your change also removes other keys from the query string, so I've opened PR #2417 to handle it more gracefully.

PR #2439, opened 2025-07-15
Prevent scrolling to top before page has rendered

### LINKED ISSUE #2125: Scroll resets before next page renders, causing flash
### Version:

- `@inertiajs/react` version: 2.0.0-beta-3 and 1.0.0

### Describe the problem:

When navigating from one page to another, just before the next page appears, the initial page scroll resets to the top of the page. This causes a janky flash just before the next page renders.

This is already described in issue #1802 but that issue is closed as fixed, and I am unsure whether my comment there has been noticed, so I am opening a new issue.

### Steps to reproduce:

This is described in #1802 :

> Create a page that causes the browser viewport to scroll
Include a link to another page
Click that link while scrolled down, and observe a flash of scroll-to-top before the transition

pascalbaljet:
Fixes #2125. It's hard to reproduce because it's only noticeable on very heavy pages, and even then it doesn’t always consistently fail.

Still draft because it misses tests, and I'm not sure if the new behavior is desired.

### DIFF
--- a/packages/react/src/App.ts
+++ b/packages/react/src/App.ts
@@ -5,6 +5,7 @@ import PageContext from './PageContext'
 
 let currentIsInitialPage = true
 let routerIsInitialized = false
+let swapPromise = null
 let swapComponent: PageHandler = async () => {
   // Dummy function so we can init the router outside of the useEffect hook. This is
   // needed so `router.reload()` works right away (on mount) in any of the user's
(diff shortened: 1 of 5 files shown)


wescopeland:
Hi @pascalbaljet 👋 

Thanks for your work on the issue (and your work on Inertia in general).

We're affected heavily by this, and it seems there's a question of whether this PR's changes are desirable. Short of forking/patching Inertia, is it possible for developers to implement some flavor of this fix in their own apps? Or is it a core problem that requires changes in Inertia itself?

pascalbaljet:
> We're affected heavily by this, and it seems there's a question of whether this PR's changes are desirable. 

Yes, I definitely want to merge it, but I haven't found a reliable way to reproduce/test it yet with Playwright.

> Short of forking/patching Inertia, is it possible for developers to implement some flavor of this fix in their own apps? Or is it a core problem that requires changes in Inertia itself?

I'm afraid this should really be fixed in Inertia itself. I'll try to work on this next week!

WarcraftLogs:
Any update here? I have sites that are heavily affected by this.

wescopeland:
Still indeed affected by this. Would love to have a fix available in some form, even if it's a configurable option.

pascalbaljet:
Closed in favor of #2721.

PR #2443, opened 2025-07-16
Support for passing custom component to `as` prop of `Link` component.

### LINKED ISSUE #1746: Link component as doesn't render other components

pascalbaljet:
Fixes #1746. Supports custom components instead of just HTML tags:

```jsx
<Link as={CustomButton} method="post" href="/user" data={{ test: 'data' }}>
    Custom Component with Data
</Link>
```

Replaces PR #2264 and #1940 (will add you as a co-author). I've opened a new PR as this builds upon PR #2441 and I don't think Inertia should alter DOM elements. It now always passes the `href` attribute to custom components and it's up to the user to handle this (or not).


nxmndr:
Hello. How are we supposed to use this in Vue+Vuetify 3 (built with webpack) ?

Because this doesn't work and I would much rather not have to rewrite all my import sequences :
``` vue
<template>
  <Link :as="VBtn" :href="..." method="post" block color="secondary">Disconnect</Link>`
</template>
<script>
  import { VBtn } from 'vuetify/components'
  export default {
    components: { ApplicationLogo, VBtn },
    ...
  }
</script>
```

The following works but I still have to rewrite all my buttons, inertia tabs, etc. and it changes the hierarchy, possibly forcing CSS changes as well. And Vuetify upgrades are already quite a full plate.
``` vue
<Link as="v-btn" :href="route('logout')" method="post">
  <v-btn block color="secondary">
    Disconnect
  </v-btn>
</Link>
```

Can we please have it at least as a native plugin ?

nxmndr:
@pascalbaljet Sorry to bother you ; is there a way to get the historical syntax ? That would be wonderful.
``` vue
<Link as="v-btn" ...>Disconnect</Link>
```

PR #2445, opened 2025-07-17
Make errorBag parameter optional

joelstein:
Before today, I could write the following code:

```ts
router.replace({
  url: `/some-url`,
  preserveScroll: true,
  preserveState: true,
})
```

However, with the latest release, now I receive a Typescript error:

> Argument of type '{ url: string; preserveScroll: true; preserveState: true; }' is not assignable to parameter of type 'ClientSideVisitOptions'.    │
│   Property 'errorBag' is missing in type '{ url: string; preserveScroll: true; preserveState: true; }' but required in type 'ClientSideVisitOptions'.

This pull request makes the `errorBag` property optional.


pascalbaljet:
Thanks!

PR #2461, opened 2025-07-27
allow passing partial errors object to `setError()`

sudo-barun:
The documentation at https://inertiajs.com/forms#form-helper says following about `setError()` for React:
```js
const { setError } = useForm({ ... })

// Set a single error...
setError('field', 'Your error message.');

// Set multiple errors at once...
setError({
  foo: 'Your error message for the foo field.',
  bar: 'Some other error for the bar field.'
});
```
The documentation does not say that errors object passed to `setError()` should contain error for all form fields. And, the functionality of `setError()` is as per the documentation. But, because of the type, errors object should contain error for all form fields.

For the following code:
```ts
import { useForm } from "@inertiajs/react";

const form = useForm({
  name: '',
  address: '',
});
form.setError({
  name: '',
});
```
the type error is:
```
Argument of type '{ name: string; }' is not assignable to parameter of type 'Record<"name" | "address", string>'.
  Property 'address' is missing in type '{ name: string; }' but required in type 'Record<"name" | "address", string>'.ts(2345)
```

This change allows partial errors object.

Similar change is applied to the package for Vue.

Regarding package for Svelte, the type is now more specific similar to React and Vue.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -38,7 +38,7 @@ export interface InertiaFormProps<TForm extends FormDataType> {
   clearErrors: (...fields: FormDataKeys<TForm>[]) => void
   resetAndClearErrors: (...fields: FormDataKeys<TForm>[]) => void
   setError(field: FormDataKeys<TForm>, value: string): void
-  setError(errors: Record<FormDataKeys<TForm>, string>): void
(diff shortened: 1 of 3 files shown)


pascalbaljet:
Thanks! I've merged it into the `typescript-improvements` branch where I'm gathering more TypeScript PRs, which will then be merged into `master`.

Issue #2466, opened 2025-07-29
Queue hangs when response callbacks has an error, all subsequent inertia requests fail to process


pintend:
### Inertia adapter(s) affected

- [x] React
- [x] Vue 3
- [x] Svelte
- [x] Not Applicable

### JS package version

2.0.17

### Backend stack (optional)

Laravel 12.x
PHP 8.x (not applicable)

### Describe the problem

If you have a form which does some processing after it's submitted but before the page is redirected and in that processing there is an uncaught error (issue is with inertia router not only the form)

For example:
```js
form.submit('some-route', {
    onSuccess: () => {
        throw new Error('simulated error');
    }
})
```

The page will still render the new component as expected. The problem is on the new page all Inertia requests get stuck in the processing stage. So now we cant render new components and the application gets stuck until a full page refresh

```js
form.submit('some-other-route', {
    onSuccess: () => {
        // This never runs
    },
    onBefore: () => {}, // runs
    onStart: () => {}, // runs
    onFinish: () => {}, // runs
    onError: () => {}, // doesnt run, and we dont expect it to run
})
``` 

Now when this form gets submitted, the request gets successfully sent to the server, inertia response received, but when inertia tries to process the response it returns the error from the previous request. this response get added to the queue for processing but never gets processed (every new Inertia request will just get added to the queue and the queue is stuck in a endless loop of returning the previous failed request)

https://github.com/inertiajs/inertia/blob/6241efe010337aa15f19c4ee7051a1f08c741809/packages/core/src/response.ts#L32 

If we take a look inside the queue
https://github.com/inertiajs/inertia/blob/6241efe010337aa15f19c4ee7051a1f08c741809/packages/core/src/queue.ts#L11
we see we only `processNext` when `processingPromise` is null but when an item on the queue fails `processingPromise` holds the error and will then always return the error 

(this was discovered in an old project which was updated to v2, the error which was happening didnt affect anything so it was never discovered until v2 caused inertia to stop routing)

### Steps to reproduce

```js
form.submit('some-route', {
    onSuccess: () => {
        throw new Error('simulated error');
    }
})
```

new inertia requests dont process

PR #2467, opened 2025-07-29
Fix core: Queue processing when an item fails

pintend:
Changed the queue helper to always set `processingPromise` to null when the promise is resolved or rejected, previously only resolved items allowed the queue to continue

Resolves https://github.com/inertiajs/inertia/issues/2466

Is there a specific reason the queue was using `.then()`?


pascalbaljet:
Thanks!

PR #2470, opened 2025-07-30
improve typescript configuration

sudo-barun:
It is better to include Typescript compiler options like `emitDeclarationOnly` directly in `tsconfig.json` file.

With this change, when `npx tsc` is run inside `packages/core/`, `.js` files are not created alongside `.ts` files.


jrmajor (approved):

pascalbaljet:
Thanks!

PR #2471, opened 2025-07-30
improve type in `useForm.ts` files

sudo-barun:
This PR improves type in `useForm.ts` files.

If you enable `strict: true` in React and Vue packages' `tsconfig.json`, you can see that all type error in those files are gone. In the case of Svelte, the type problem still exist and all are related to `setStore()`.


pascalbaljet:
Thanks! I'll get back to this when #2468 and #2469 are merged.

PR #2479, opened 2025-08-02
Only reset useForm processing and progress with onFinish

### LINKED ISSUE #2463: useForm `processing` and `progress` should only be reset in `onFinish`

- [x] React
- [x] Vue 3
- [x] Svelte
- [ ] Not Applicable

### JS package version

2.0.x

### Backend stack (optional)

_No response_

### Describe the problem

Currently the form helper's `onSuccess`, `onError`, `onCancel`, and `onFinish` callbacks _all_ set `processing = false` and `progress = null`. I'm sure this was done for a reason, but I'm suggesting it _only_ be called in `onFinish`.

Aside from the pedantic and DRY reasons to do this, this also helps those of us using the feature of `onSuccess` and `onError` where you can [return a promise](https://inertiajs.com/manual-visits#event-callbacks), and `onFinish` won't be called until it resolves. 

I'm using that feature in some forms that do work in a background job, which I poll the status of inside the promise, and it'd be nice if the form's `processing` was true during that entire time and not just the initial submission.

### Steps to reproduce

n/a

mattwigham:
Closes #2463

Previously the form helper's onSuccess, onError, onCancel, and onFinish callbacks all set `processing = false` and `progress = null`. This PR makes it so that now only happens in `onFinish`.

Aside from the pedantic and DRY reasons to do this, this also helps those of us using the feature of onSuccess and onError where you can [return a promise](https://inertiajs.com/manual-visits#event-callbacks), and onFinish won't be called until it resolves.

I'm using that feature in some forms that do work in a background job, which I poll the status of inside the promise, and this change allows the form's processing to remain true during that entire time and not just the initial submission.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -124,8 +124,6 @@ export default function useForm<TForm extends FormDataType>(
         },
         onSuccess: (page) => {
           if (isMounted.current) {
-            setProcessing(false)
-            setProgress(null)
             setErrors({})
             setHasErrors(false)
             setWasSuccessful(true)
@@ -144,8 +142,6 @@ export default function useForm<TForm extends FormDataType>(
         },
         onError: (errors) => {
           if (isMounted.current) {
-            setProcessing(false)
-            setProgress(null)
             setErrors(errors)
             setHasErrors(true)
           }
@@ -155,11 +151,6 @@ export default function useForm<TForm extends FormDataType>(
           }
         },
         onCancel: () => {
-          if (isMounted.current) {
-            setProcessing(false)
-            setProgress(null)
-          }
-
           if (options.onCancel) {
             return options.onCancel()
           }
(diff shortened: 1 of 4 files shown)


pascalbaljet:
This is technically a breaking change. Code that checks `form.processing` inside `onSuccess`/`onError` would now see `true` instead of `false`. For example, if you're showing a spinner based on processing within those callbacks, it would still be visible until `onFinish` fires.

That said, I think this change makes sense since it aligns with the documented promise behavior. I'll merge this in v3.

pascalbaljet:
@mattwigham Can you tick the 'Allow edits from maintainers' checkbox so I can add more tests? Thanks!

mattwigham:
@pascalbaljet I just realized this is from our gh org, so I don't have that option to allow edits. Would you like me to resubmit the PR as a user or do you want to recreate these changes yourself since they're minimal? Sorry for the hassle!

pascalbaljet:
> @pascalbaljet I just realized this is from our gh org, so I don't have that option to allow edits. Would you like me to resubmit the PR as a user or do you want to recreate these changes yourself since they're minimal? Sorry for the hassle!

No problem, I'll open a new PR and add you as co-author!

PR #2485, opened 2025-08-07
[3.x] Pass props as second argument to resolve() in createInertiaApp()

HichemTab-tech:
Re‑attempt of #2330

_closes #2400_


pascalbaljet:
Thinking out loud, but wouldn't it be nice if we'd pass the whole page object as a second argument, so that you also have access to the URL, version, etc.? Or doesn't that make sense?

HichemTab-tech:
> Thinking out loud, but wouldn't it be nice if we'd pass the whole page object as a second argument, so that you also have access to the URL, version, etc.? Or doesn't that make sense?

Actually it does make sense, I mean we would have more context using page instead of props.
I'll make the changes, Thanks

HichemTab-tech:
@pascalbaljet it's done thanks for the suggestion.

btw I noticed an error in the **Coding Standards** workflow, I'll open a separated issue for that.

pascalbaljet:
I think we need to push this PR to the next major release of Inertia. The problem is that it changes the signature of some public methods like `router.resolveComponent()`, and that would be a breaking change. I tried making the `page` argument optional throughout the PR, but that leads to other problems where users need to check if the argument is passed, and there might even be some edge cases where this would break users’ apps or TS checks. To be clear, I'm still in favor of this PR so I'm not closing it yet, but I also don't see a way to change this into a non-breaking change.

Issue #968, opened 2021-11-11
Dynamic properties in form


bajki:
### Versions:

- `@inertiajs/inertia` version: 0.10.1
- `@inertiajs/inertia-vue` version: 0.7.2

### Describe the problem:

Is there any way to add property to form if it's not defined  on start?
Like this it won't post custom field.
Thanks

### Steps to reproduce:
```
data() {
        return {
            sending: false,
            form: this.$inertia.form({
                name: "",
      })
};
},
methods: {
        submit() {
            this.sending = true;

            this.form.post(this.route("test"), {
                preserveScroll: true,
                onSuccess: () => (this.sending = false),
                onError: () => (this.sending = false)
            });
        }
}
mounted(){
  this.form.custom = "example".
}
```

ajnsn:
Hey @bajki 
the Form Helper only includes the attributes you set when initaliazing it. What you could do is to build up the form in the mounted hook, e.g: 

```js
export default {
    data() {
        return {
            form: {},
        };
    },
    mounted() {
        this.form = this.$inertia.form({
            name: '',
            custom: 'example',
        });
    },
};
```

bajki:
So there is no way to reinitaliazing after you set it.
 
The idea is that i don't know if  i will use custom field or don't beacuse i dynamically import components.

This way works, it will send custom field but i created all with helper.
```
<template>
  <input v-model="form.custom" />
</template>
<script>
export default {
  data() {
    return {
      form: {},
    };
  },

  methods: {
    submit() {
      this.$inertia.post(this.route("test"), this.form);
    },
  },
};
</script>

```

ajnsn:
You can either build the whole form object in your submit()-method or use the transform-callback, Example [here](https://github.com/inertiajs/inertia/discussions/382#discussioncomment-338461).

claudiodekker:
Thanks for answering this one @ajnsn 👍

juanparati:
There are many ways of add dynamic properties with the "form" helper.

### - Creating a dynamic property (First alternative solution)
You should create a root property which value is an empty object.

Example:
```
this.form = useForm({root: {}});
```

Then you need to provide the "root" as reference:

```
<input type="text" v-model="form.root.mydynamicprop" />
```

The disadvantage is that you will need to use the main/root property as reference. If you are using Laravel and you want to keep the request validation without unfold the main/root node you can transform the request before to send.


### - Populate the properties from a Laravel model (Second alternative solution)

[...]

PR #2495, opened 2025-08-11
Add Support for Dynamic Root-level Form Properties in Vue and Svelte

flexponsive:
Adds support for dynamic root-level form properties in Vue and Svelte, matching existing React behavior. Properties can be added after initialization, which helps avoid pre-defining every possible field on create pages and supports temporary values like “type your username to confirm.”  

**Changes**  
- Update `form.data()` in Vue and Svelte to track all form properties, including those added at runtime, not only those defined at initialization.  
- Update example pages to demonstrate dynamic properties.  
- Add Playwright tests for dynamic root-level and nested properties in all adapters.  

**Motivation**  
Fixes [#968](https://github.com/inertiajs/inertia/issues/968) and follows [discussion #859](https://github.com/inertiajs/inertia/discussions/859). Ensures consistent behavior across adapters.

### DIFF
--- a/packages/react/test-app/Pages/FormHelper/Data.jsx
+++ b/packages/react/test-app/Pages/FormHelper/Data.jsx
@@ -1,10 +1,12 @@
 import { useForm, usePage } from '@inertiajs/react'
+import { useEffect, useState } from 'react'
 
 export default (props) => {
   const form = useForm({
     name: 'foo',
     handle: 'example',
     remember: false,
+    custom: {},
   })
 
   const page = usePage()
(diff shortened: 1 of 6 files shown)


flexponsive:
@pascalbaljet Thanks for all the recent work on improving the forms, especially the TypeScript support — the new type checks for form fields are a big robustness win.

That said, in JavaScript apps the long-standing issue remains: form fields added at runtime are still silently dropped in `data()` for Vue and Svelte. This is orthogonal to the TypeScript improvements and could be resolved with a very small code change, as suggested [here](https://github.com/inertiajs/inertia/pull/2495/files#diff-6a34bba1f476223d7b1c47c1b1d060324601d770f5d6c9b4a723707bd91dc728).

Would you be open to a PR along these lines? I can take care of the test app updates across all adapters and ensure no backward-compatibility issues.

pascalbaljet:
Thanks for the PR, but I’m closing this after consideration. The form helper is designed around defining fields upfront, which provides type safety. Allowing dynamic properties works against this and could mask typos. `form.name = 'foo'` would silently create a new property rather than being ignored.

PR #2499, opened 2025-08-13
Reset form fields by name in Form components

skryukov:
This is an addition to upcoming `Form` component that makes `reset` function aligned with one from `useForm` by allowing passing input names to reset a subset of inputs:

`useForm` example:
```js
const { reset } = useForm({ ... })

// Reset the form...
reset()

// Reset specific fields...
reset('field', 'anotherfield')
```

`Form` component example:

```jsx
export default function FormReset() {
  const formRef = useRef(null)
  
  // Reset the form...
  formRef.current?.reset()
  
  // Reset specific fields...
  formRef.current?.reset('field', 'anotherfield')

  return <Form ref={formRef} ...>...</Form>
}
```


pascalbaljet:
I did a refactor so it takes the defaults internally from the `<Form>` component and not from the form elements themselves so that we can change the defaults (like with `useForm()`). It now also dispatches just one event on the form element instead of an event for all reset input elements. This matches the behavior on classic HTML forms. I also wrote an extensive test suite to cover all form element types. There's something wrong with the Svelte implementation that I'm going to check now!

skryukov:
> it takes the defaults internally from the `<Form>` component and not from the form elements themselves

Shouldn't we refactor React and Vue to use `fields = [...defaults.keys()]` as well, since `formElement.reset()` uses form elements defaults?

pascalbaljet:
> Shouldn't we refactor React and Vue to use `fields = [...defaults.keys()]` as well, since `formElement.reset()` uses form elements defaults?

Yes, I'll move that to `resetFormFields()` in a follow-up PR where I'll add support for the `defaults()` method.

pascalbaljet:
> Yes, I'll move that to `resetFormFields()` in a follow-up PR where I'll add support for the `defaults()` method.

For reference: merged in PR #2507

PR #2503, opened 2025-08-13
Add `onSubmitComplete` prop to `Form` component

joetannenbaum:
This PR adds an `afterSubmit` prop to the `Form` component, allowing you to `reset`, `resetAndClearErrors`, and/or `clearErrors` after a form submission.

```tsx
<Form method="post" afterSubmit={(props) => props.reset('name')}>
    <input type="text" name="name" id="name" placeholder="Name" defaultValue="John Doe" />
    <input type="email" name="email" id="email" placeholder="Email" defaultValue="john@doe.biz" />
</Form>
```

This is fired right after the `onFinish` event in the request lifecycle.

I also updated the `Form` component `reset` function to receive field names in the event that the user wants to reset individual fields instead of the whole form.

Questions:

- Do we like `afterSubmit`?
- Is the timing right in the request lifecycle (after `onFinish`)?
- I whittled down the props being sent to the callback to just the above functions as the other props didn't seem relevant to this action, anything else I may have missed?
- I decided against the `beforeSubmit` prop as `onBefore` does essentially the same thing

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -434,26 +434,6 @@ export type FormComponentProps = Partial<
   disableWhileProcessing?: boolean
 }
 
-export type FormComponentonSubmitCompleteArguments = {
-  clearErrors: (...fields: string[]) => void
-  resetAndClearErrors: (...fields: string[]) => void
-  reset: (...fields: string[]) => void
-}
-
-export type FormComponentChildProps = {
-  errors: Record<string, string>
-  hasErrors: boolean
-  processing: boolean
-  progress: {
-    percentage: number
-  } | null
-  wasSuccessful: boolean
-  recentlySuccessful: boolean
-  setError: (fieldOrFields: string | Record<string, string>, maybeValue?: string) => void
-  isDirty: boolean
-  submit: () => void
-} & FormComponentonSubmitCompleteArguments
-
 export type FormComponentMethods = {
   clearErrors: (...fields: string[]) => void
   resetAndClearErrors: (...fields: string[]) => void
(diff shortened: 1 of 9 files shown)


joetannenbaum:
Changed the prop name from `afterSubmit` -> `onSubmitComplete` to be more consistent with other event props. Open to other names if you have a better idea.

joetannenbaum:
Docs PR: https://github.com/inertiajs/inertiajs.com/pull/447

pascalbaljet:
> Changed the prop name from afterSubmit -> onSubmitComplete to be more consistent with other event props. Open to other names if you have a better idea.

I like the `on` prefix as well. An alternative might be `onSubmitEnd`, but only because it's shorter. I don't think it's necessarily better.

> Is the timing right in the request lifecycle (after onFinish)?

I think it's a good thing that it comes after all existing Inertia events so that it's clear that this is specifically for the Form component.

> I whittled down the props being sent to the callback to just the above functions as the other props didn't seem relevant to this action, anything else I may have missed?

We're missing `defaults()`, but that's missing in the Form component in general. I will add that in a follow-up PR.

> I decided against the beforeSubmit prop as onBefore does essentially the same thing

Agreed.

skryukov:
Sorry for hunting you in PRs @pascalbaljet 😂

Here's an example action I have in the Rails starter:

```js
const submit: FormEventHandler = (e) => {
    e.preventDefault()

    patch(settingsEmailPath(), {
      preserveScroll: true,
      onSuccess: () => reset("password_challenge"),
      onError: (errors) => {
        if (errors.password_challenge) {
          reset("password_challenge")
          currentPasswordInput.current?.focus()
        }
      },
    })
  }
```

Right now I can't port `onError` logic to `Form` component without `formRef`, do you think we need `onSubmitComplete`'s evil tween—`onSubmitError`?

joetannenbaum:
@skryukov we actually just greatly simplified this (literally just now in v2.1.2!): https://github.com/inertiajs/inertia/pull/2514

PR #2504, opened 2025-08-13
Add `Form` component `disableWhileProcessing` prop

joetannenbaum:
This PR adds a `disableWhileProcessing` prop to the `Form` component. When this prop is used, the `inert` attribute is automatically added to the `form` HTML tag while it is processing.

```tsx
const App = () => (
  <Form disableWhileProcessing>
      {/* ... */}
  </Form>
)
```

Open to other names for this prop, it's a bit verbose but I'm going for clarity.


joetannenbaum:
Docs PR: https://github.com/inertiajs/inertiajs.com/pull/446

pascalbaljet:
I thought about `disableOnSubmit`, but I like `disableWhileProcessing` more because we also have `processing` in `useForm()`.

PR #2506, opened 2025-08-14
Introduce temporary props

skryukov:
Inertia temporary props are server-provided properties that should be available only for the next render and then be discarded. Inertia deliberately does not cache them in history. This gives you predictable one-time UI feedback and prevents stale data (like status messages ) from resurfacing when users navigate back/forward or when a cached page is reused.

Usage example (requires implementation in Laravel adapter):

```php
    /**
     * Show the login page.
     */
    public function create(Request $request): Response
    {
        return Inertia::render('auth/login', [
            'canResetPassword' => Route::has('password.request'),
            'status' => Inertia::temp($request->session()->get('status')),
            // should we support Inertia::defer(...).temp, Inertia::optional(...).temp?
        ]);
    }
```

I believe this is a simple yet useful addition to the Inertia API, but you might want to keep the public API short, so before implementing the Laravel adapter PR, I'd love to hear your thoughts.

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -66,6 +66,9 @@ class History {
 
   protected getPageData(page: Page): Promise<Page | ArrayBuffer> {
     return new Promise((resolve) => {
+      if (page.temporaryProps) {
+        page.temporaryProps.forEach((key) => { delete page.props[key] })
+      }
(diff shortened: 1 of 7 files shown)


skryukov:
Closing in favor of upcoming `Inertia::flash()` prop

PR #2520, opened 2025-08-17
[v2.x] fix: type error by changing page props type to `any`

peaklabs-dev:
## Changes

- fix: type errors. Example error: `Type unknown is not assignable to type boolean.` I get this error (and more for each line) when trying to use `$page.props.username`, because props are by default typed to `unknown`.

### Example

Example 1:

```ts
    type FormValues = {
        username: string;
        notifications_enabled: boolean;
    };

    const defaultValues: FormValues = {
        username: $page.props.username,
        notifications_enabled: $page.props.notifications_enabled,
    };
```

Example 2:

- Showing a flash message in svelte (I am aware I can get the flash message via `$props()` in svelte and type it there or leave it, which works because `$props()` has type of `any`). The following should be possible, but it is not currently, as I get type errors.

```svelte
    {#if $page.props.flash.message}
      <div class="alert">{$page.props.flash.message}</div>
    {/if}
```


pascalbaljet:
I don't like to change this across all adapters, as in React/Vue you can do `usePage<PageProps>()`. Also, `unknown` in this case is preferred over `any`. Can't we make the change in `packages/svelte/src/page.ts`?

```ts
type SveltePage = Omit<Page, 'props'> & {
  props: Page['props'] & {
    [key: string]: any
  }
}

const { set, subscribe } = writable<SveltePage>()
```

peaklabs-dev:
@pascalbaljet Thanks for the review. I have changed it, let me know if anything else is needed.

pascalbaljet:
Thanks! I've added a test so we know it keeps working in the future.

PR #2532, opened 2025-08-19
[Vue] Fixing action attribute on Form Component when using Wayfinder

nicolagianelli:
This PR addresses an issue where the Form component in Vue did not correctly handle action attribute when it was provided as an object by Wayfinder. Previously, the component was missing a check to determine if the action was an object, which prevented it from correctly using the url property.

This change aligns the Vue Form component's behavior with its React counterpart, ensuring that action provided as object (e.g., from Wayfinder) are correctly processed as url for the attribute.


pascalbaljet:
Thanks!

PR #2535, opened 2025-08-20
Replace html-escape with built-in function on Svelte package

kresnasatya:
The references come from: https://github.com/sveltejs/svelte/blob/main/packages/svelte/src/escaping.js (Svelte 5) and https://github.com/sveltejs/svelte/blob/svelte-4/packages/svelte/src/shared/utils/escape.js (Svelte 4). By using the built-in function, we can reduce third-party dep on Svelte package.


kresnasatya:
Hmm, I'm confused why I fail in coding standards. I have run `pnpm run format`.

kresnasatya:
> I’m on board with this change 👍 The Svelte escape is a better fit for our attribute use case, since it escapes exactly what we need, avoids over-escaping, and is also seems to be bit more efficient.
> 
> @kresnasatya please make to pass the second `true` argument so it escapes correctly for an HTML attribute.

Alright. I have added it. Please review @pedroborges . Thank you.

pedroborges:
@pascalbaljet I fixed the merge conflict but now these ES2020 checks for React and Vue are failing, I'm not sure what those are about. Any ideas?

pascalbaljet:
> @pascalbaljet I fixed the merge conflict but now these ES2020 checks for React and Vue are failing, I'm not sure what those are about. Any ideas?

Ah yes, npmjs is giving 403/409/422 errors all day in CI, nothing to worry about it seems :)

pascalbaljet:
Can't we use `escape` from `lodash-es`? We're already using that in the `<Head>` components. That would avoid having to maintain this code ourselves.

pedroborges:
@kresnasatya please give lodash-es escape a try and run the tests. From a little research I did it seems very similar to the escape utility from Svelte.

kresnasatya:
I see the escape function from lodash-es pass the svelte test. How do you think @pedroborges? If it's okay then, I just need one more commit to remove the `escape.ts` from the Svelte package.

pedroborges (changes requested):
Nice! Only thing left is removing `escape.ts`.

kresnasatya:
Alright, all is good. Feel free to check again @pedroborges

pascalbaljet:
Thanks!

PR #2538, opened 2025-08-22
Fix keyboard activation when using `prefetch: 'click'`

### LINKED ISSUE #2274: Using link prefetching with 'click' option breaks accessibility
### Version:

- `@inertiajs/svelte` version: 2.0.3

Any link that uses `use:inertia={{ prefetch: 'click' }}` stops responding to enter key while focused. 
Links with regular use:inertia work just fine, navigating on enter key press.

pedroborges:
Keyboard activation (Enter/Space) isn't working for links using `prefetch: 'click'` because the click event handler prevented default behavior but didn't trigger the `mousedown`/`mouseup` events used for prefetching and navigation. Fixes #2274.

Without proper keyboard support, links become unusable for users with motor disabilities, screen reader users, and anyone who navigates with a keyboard.

I fixed this issue by adding:
- `keydown`/`keyup` event handlers to mirror the existing `mousedown`/`mouseup` behavior
- `shouldActivateOnKey()` to ensure Space only activates buttons (not links, per accessibility standards)
- tests to verify keyboard activation works correctly


pascalbaljet:
Thanks!

PR #2545, opened 2025-08-23
[v2.x] feat: allow adding type to `router.restore`

peaklabs-dev:
## Changes

- feat: allow adding type to state restore

### Example

I want to use the state to restore form values which need to be typed for my forms to work properly (I use TanStack form).

```ts
type FormState = {
  username: string
  notifications_enabled: boolean
}

const FORM_STATE_KEY = 'TEST'
const restoredState = router.restore<FormState>(FORM_STATE_KEY)
```

### DIFF
--- a/packages/react/test-app/Pages/Remember/Router.tsx
+++ b/packages/react/test-app/Pages/Remember/Router.tsx
@@ -1,0 +1,43 @@
+import { router } from '@inertiajs/react'
+import { useState } from 'react'
+
+export default () => {
+  const [foo, setFoo] = useState('-')
+  const [bar, setBar] = useState(0)
+
+  function remember() {
+    router.remember('foo')
+    router.remember(42, 'bar')
+  }
+
+  function restore() {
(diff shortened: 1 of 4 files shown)


pascalbaljet:
Where does `'TEST'` come from in this context? I would expect it to behave like this (open for discussion, though!):

```ts
type RememberedState = {
  name: string
  isAdmin: boolean
}

const key = 'isAdmin'

router.restore('id') // no generic, should always work
router.restore<RememberedState>('name') // should be string | undefined
router.restore<RememberedState>('email') // ts-expect-error, email doesn't exist on RememberedState
router.restore<RememberedState>(key) // should be boolean | undefined
```

peaklabs-dev:
@pascalbaljet Oh, the `FORM_STATE_KEY` is just the key for where the data is stored, nothing more. I remember the form data via a listener like this:
```ts
        listeners: {
            onChange: ({ formApi }) => {
                router.remember(formApi.state.values, FORM_STATE_KEY);
            },
        },
```

PR #2550, opened 2025-08-28
Prevent errors caused by null href value

fritz-c:
The adapters all use the same logic for retrieving the method for a Link component:
```js
method = typeof href === 'object' ? href.method : /* ... */
```
This logic, however, throws a `Cannot read properties of null (reading 'method')` error at runtime when the value of `href` is `null`, because `typeof null === 'object'` is also true.

Supposing the dev was paying attention to the dev console or using TypeScript, they have a chance to catch this: href is only allowed to be a string or an object, and `null` fails prop validation. I use TypeScript in my Vue project, but the use of `<component :is="as || Link"` got rid of type safety, and the case of `href` coming up as `null` was an unknown edge case to begin with, so I only noticed it after seeing production error logs with the error. So I believe that it's in the best interest of Inertia usability for it to have a bit of a safety net here.

On a side note about project maintainability, the project was mostly easy to set up, and I followed `CONTRIBUTING.md`, but still encountered some bumps in the road to contributing:

* buidl for types
* playwright install
* pr template 
* no unit tests

### DIFF
--- a/packages/react/src/Link.ts
+++ b/packages/react/src/Link.ts
@@ -60,7 +60,7 @@ const Link = forwardRef<unknown, InertiaLinkProps>(
     const hoverTimeout = useRef<number>(null)
 
     const _method = useMemo(() => {
-      return typeof href === 'object' ? href.method : (method.toLowerCase() as Method)
+      return href && typeof href === 'object' ? href.method : (method.toLowerCase() as Method)
     }, [href, method])
 
     const _as = useMemo(() => {
--- a/packages/svelte/src/link.ts
+++ b/packages/svelte/src/link.ts
@@ -137,7 +137,10 @@ function link(
 
     cacheTags = cacheTagValues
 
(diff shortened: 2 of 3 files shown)


pascalbaljet:
Thanks, I've updated it a bit to use `isUrlMethodPair` throughout the library. I'll improve the generation of TS files in another PR. We definitely prefer E2E tests over unit tests, so that's why you won't find a lot of them :)

fritz-c:
Nice! Looking at your changes, I can't find where the `isUrlMethodPair` function is exported from. Yet the CI tests here still pass. Did I overlook it getting added somewhere? 😅

pascalbaljet:
> I can't find where the `isUrlMethodPair` function is exported from.

It comes from the `core` package:

https://github.com/inertiajs/inertia/blob/62681e62dae9bdec8ffc31e3f51ad0df403d1b30/packages/core/src/url.ts#L90-L92

fritz-c:
Oh, it existed already! Nice.

PR #2554, opened 2025-09-01
Remove data from the dependency array of setDefaults

jasonlbeggs:
In Laravel Cloud's edit dialogs, we have effects that watch for changes to the model we're passing in, then set the form data to that model's properties and sets the defaults.

Previously, we had to leave `setDefaults` out of the effect's dependency array and ignore the ESLint error because `setDefaults`'s memoization was dependent on the `data` property, so it would cause an infinite loop:

```
useEffect(() => {
    setData(model);
    setDefaults();
}, [model, setData]); // eslint-disable-line react-hooks/exhaustive-deps
```

The correct way would be to include the `setDefaults` function as a dependency:

```
useEffect(() => {
    setData(model);
    setDefaults();
}, [model, setData, setDefaults]);
```

This PR uses Kent C. Dodd's "latest ref" pattern to remove `data` from the dependency array. [This is an older article](https://www.epicreact.dev/the-latest-ref-pattern-in-react), but it still explains the idea.

### DIFF
--- a/packages/react/src/useForm.ts
+++ b/packages/react/src/useForm.ts
@@ -222,12 +222,18 @@ export default function useForm<TForm extends FormDataType<TForm>>(
 
   const [dataAsDefaults, setDataAsDefaults] = useState(false)
 
+  const dataRef = useRef(data)
+
+  useEffect(() => {
+    dataRef.current = data
+  })
+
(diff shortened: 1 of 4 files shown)


pascalbaljet:
Thanks!

PR #2594, opened 2025-09-28
Additional InfiniteScroll component tests

pascalbaljet:
Some additional tests to ensure that the component's state is reset when you navigate to the same page (optionally with additional query params).


pascalbaljet:
Closed in favor of #2595.

Issue #2498, opened 2025-08-12
Hydration mismatch occurs when rememberedState is present in history during reload


sudo-barun:
### Inertia adapter(s) affected

- [x] React
- [x] Vue 3
- [x] Svelte
- [ ] Not Applicable

### JS package version

2.0.17

### Describe the problem

When rememberKey is used in form and SSR is enabled, hydration mismatch occurs after making some changes in the form and reloading the page.

After reloading the page, `window.history.state.page.rememberedState` is supposed to get reset to empty object before DOM node and vNode are compared in hydration. But for some reason, the reset happens later. If we check the value of `window.history.state.page.rememberedState` in console after hydration fails, it is an empty object.

This problem only occurs in version Inertia v2 and not in v1 as mentioned in the "Steps to reproduce" section.

### Steps to reproduce

1. Clone this repo: https://github.com/sudo-barun/pingcrm

1. Go to branch [`hydration-mismatch-when-rememberedState-is-present-in-history`](https://github.com/sudo-barun/pingcrm/tree/hydration-mismatch-when-rememberedState-is-present-in-history)

1. Setup the app as instructed in `README.md`. Make sure SSR is working.

1. Open the app in browser and submit login form

1. Go to user-edit page (like `/users/3/edit`)

1. Remove the text inside `First name` form field

1. Reload the webpage and check the console. You should see following message:
   ```
   Hydration completed but contains mismatches.
   ```

To check previous working version, go back few commits in the same git branch before the commit where npm package `@inertiajs/vue3` and composer package `inertiajs/inertia-laravel` are updated to version 2.

pascalbaljet:
This has been fixed in #2591.

Issue #2402, opened 2025-06-25
IOS Chrome does not remember browse history


RyanDaDeng:
### Version:

- `@inertiajs/vue3` version: #.#.#

### Describe the problem:

The problem is simple, just open any app and the navigate deep then just click IOS Chrome back history, it will just go back wherever you started.

### Steps to reproduce:

1. Open IOS Chrome 
2. Navigate a few pages using <Link>, visit()
3. Use browser back button

pascalbaljet:
Possibly related to #1999 and #2401.

RyanDaDeng:
> Possibly related to [#1999](https://github.com/inertiajs/inertia/pull/1999) and [#2401](https://github.com/inertiajs/inertia/pull/2401).

I installed the new version, but the problem still remains in IOS Chrome - it just can't navigate back property using browser Back button - it only happens on IOS Chrome browser not others.

RyanDaDeng:
The problem still remains in IOS Chrome - whenever i click the broswer back button, the problem appears, it just back to a page that i already visited a few pages before.

wescopeland:
This still exists in the latest Inertia build and is, unfortunately, a major pain point for mobile users.

The only workaround I know to solve the issue is to track URL changes globally, and if the user is on mobile, force a hard page refresh. This defeats the purpose of having a SPA.

pascalbaljet:
I've fixed a race condition in the history manipulation logic that could cause issues with browser history state management. It's available starting in v2.2.7. Could you please update the Inertia library and confirm if the issue still exists or not?

pascalbaljet:
I could reproduce it and it's fixed in v2.2.8.

Issue #2361, opened 2025-05-31
Inertia's History class is replacing required existing history state after a page reload (F5)


stringfold:
### Versions:

- `@inertiajs/vue3` version: 2.0.11

(I believe the issue is in the code common to all client types)

### Describe the problem:

I am in the middle of porting a sizeable webapp from Adonisjs v4/Vuejs v2 to Adonisjs v6/Vuejs v3.

A lot has changed on both the server and client side, including the introduction of Inertia as the glue in the framework. But the most pertinent point to this issue is I am still using the original **vue-router** router and not Inertia's own router. 

My app client's routing code also uses the window.history object to restore state after a full page refresh, which worked fine before I migrated my code to the new framework version. But now that Inertia is in the mix, the saved window.history state is being overwritten _after_ the very end of the F5 page refresh. The result is that the refresh works once, but if you hit F5 again, the route state is already gone and the webapp functionality breaks.

I eventually tracked the culprit down to Inertia's history management code in **packages/core/src/history.ts** (doReplaceState:175) where you replace the existing history state with component and scroll state data. The fact that you delay the state replacement until the next event loop tick (replaceState:156) made it all the more difficult to figure out what was going on, since from the app's point of view, the page refresh completes fine with the routing state intact.

I understand that you need to use the window.history object to implement the ability to save component state between refreshes in your homegrown router, and I fully understand that I might be an edge case where I am continuing to use **vue-router** for legacy reasons, but given you delay overwriting the app-saved history state until after all processing within the app is complete, it makes it very difficult to workaround with 100% certainty. 

At the moment, I have to save my history state in the onMounted() hook of my Main component and then replace your replacement within a setTimeout(100) call. Not very satisfactory.

So, what I am asking is can you fix in the Inertia code so it does not wipe out the existing history when adding in the state you require? 

Perhaps you could check for a specific property in the existing history state, like "params" and make sure that's carried over if present? That seems like a very safe fix to me, since it doesn't change the timing of the call. Maybe you can have a global setting that allows developers to specify which history.state properties to carry forward?

While my case might be an edge case, I can certainly see that other developers who are using window.history for other reasons could be tripped up by this issue. It's not easy to debug changes to the window.history object find out where they're coming from, especially when they're made outside the scope of the app's execution.

### Steps to reproduce:

Put the following code in the main page component setup on the Vue 3 client:

    onMounted(() => {
      // Any arbitrary app state
[...]

Issue #2300, opened 2025-04-08
Client visits incorrect history state


sean-dickinson:
### Version:

- `@inertiajs/vue3` version: 2.0.3

### Describe the problem:
- Inertia added the ability to make client-side visits in v2 which is a great addition. However I think the reason for wanting to be able to do these visits would mostly be around using query parameters to do some filtering on the frontend and this currently does not work as expected due to the way state is handled. 

### Steps to reproduce:

- Create a component that relies on query parameters from the `usePage` composable.
  ```ts
  const page = usePage();
  const isFiltered = computed(() =>  page.url.includes('filter=true'));
  const baseUrl = computed(() => page.url.split('?')[0]);
  const updateFilter = () => {
    if(isFiltered.value){
      router.push({
       url: baseUrl.value
      });
    } else {
     router.push({
       url: `${baseUrl.value}?filter=true`
     });
    }
  }
```

- call your function that makes a client side visit to update the url
- Watch as everything works as expected
- Press the browser back button
- Watch as the url becomes out of sync with the state

I think there's a larger opportunity here for figuring out a better way to handle client-side visits in a way that allows handling of query parameters more naturally. My exact same component here works perfectly and stays in sync with the history state if I use `router.get()`.

I feel like the scenario here is a common one where you want to support query parameters in the url and still get to decide how they are acted on (frontend or backend) and how they affect the state of the page. While inertia handles the history state via javascript for client side visits with no url changes, there's no way to bookmark a page or send the url to someone else and get that same state.

I can make my page support the query parameters, but I can't get the history state to support them without fighting the built in history state that inertia is trying to provide.

**Comment by artygrand** [comment]
Maybe it unrelated, idk, but...
```
router.push({
     props: {any}
});
```
without `url` provided, will replace state, instead of really adding one, so I can't go back
It works like `router.replace()`

pascalbaljet:
I've fixed a race condition in the history manipulation logic that could cause issues with browser history state management. It's available starting in v2.2.7. Could you please update the Inertia library and confirm if the issue still exists or not?

PR #2624, opened 2025-10-07
Fix race condition in `history.ts`

### LINKED ISSUE #2406: Inconsistent site / back navigation when using useForm
### Versions:

- `@inertiajs/vue3` version: ^2.0.x

### Describe the problem:

Since version 2.x, there's been inconsistent behavior when doing back navigation through a site's history.

There are currently two scenarios for how site navigation works, which are different based on whether you're navigating to a page that makes use of the useForm helper with or without a remember key (as described in https://inertiajs.com/forms).

### Steps to reproduce:

Scenario 1 (using a remember key):

1. Start on a page that doesn't make use of a useForm element (for example users.index)
2. Navigate to a user edit page (via e.g. a table row click, so we land on users.1.edit page, that has a useForm usage like useForm('EditUserForm', { id: props.user.id, name: props.user.name }))
3. Navigate back (via browser back button) and you will land on the users.index page again as expected
4. Navigate to another user edit page (e.g. user.2.edit)
5. When navigating back, you will now land on the user.1.edit page, instead of the expected users.index page
6. If you navigate back once further, it will completely skip the users.index page, and go back to a previous page

Scenario 2 (not using a remember key)
1. Once again, start on the users.index page
2. Navigate to users.1.edit (where we make use of useForm without a rememberKey, so useForm({id: props.user.id, name, props.user.name})
3. Navigate back, you will land on users.index as expected
4. Navigate to users.2.edit
5. Navigate back, you will once again land on users.index as expected
6. Continue the above pattern infinitely, works as expected

In version 1.x of the InertiaJS library, site navigation always worked the same way as described in scenario 2, which I would say is the expected scenario. If nothing else, an end user would expect the site navigation to behave consistently, regardless if we use the form helper with or without rememberKeys.

Up to version 2.0.13, there was also an issue with the forms misremembering data, but that was fixed in the 2.0.14 version by #2401.

pascalbaljet:
This PR fixes a race condition in `history.ts` where multiple calls to methods like `pushState` and `replaceState` would corrupt the browser history.

Fixes #2406 (and possibly #2122, #2140, #2300, #2361, #2402, and #2498).

### DIFF
--- a/packages/react/test-app/Pages/FormHelper/RememberEdit.tsx
+++ b/packages/react/test-app/Pages/FormHelper/RememberEdit.tsx
@@ -6,21 +6,15 @@ interface User {
   email: string
 }
 
-export default ({ user, withRememberKey = false }: { user: User; withRememberKey?: boolean }) => {
-  const form = withRememberKey
-    ? useForm('EditUserForm', {
-        name: user.name,
-        email: user.email,
-      })
-    : useForm({
-        name: user.name,
-        email: user.email,
-      })
+export default ({ user }: { user: User }) => {
+  const form = useForm('EditUserForm', {
+    name: user.name,
+    email: user.email,
+  })
 
   return (
     <div>
       <h1>Edit User {user.id}</h1>
-      <p>With remember key: {String(withRememberKey)}</p>
       <form>
         <div>
           <label>Name:</label>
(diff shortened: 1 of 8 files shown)


pascalbaljet:
For reference: I'm not sure if we need that `isChromeIOS` check anymore. Maybe it was a workaround at the time for the same problem this PR fixes. See also PR #1992.

PR #2549, opened 2025-08-27
TypeScript cleanup + Strict mode

pascalbaljet:
What started as a round of TS cleanup in preparation for another PR, resulted in enabling the `strict` flag across the adapter. This will lead to higher quality exported types, and bugs might see the light of day earlier.

Fixes #1793.

PR #2636, opened 2025-10-16
Support for type-hinting shared Page Props

pascalbaljet:
This PR allows you to type-hint shared page props in a global type definition file, for example, `resources/js/types/globals.d.ts`.

```ts
  export interface InertiaConfig {
    sharedPageProps: {
      auth: { user: { id: number; name: string } | null }
      flash: { success?: string; error?: string }
    }
  }
}
```

Now the shared page props are automatically type-hinted throughout the app when you use `usePage()`.

```js
const page = usePage()
const successMessage = page.props.flash.success 
```

In the example above, `successMessage` will be typed as `string | undefined`. In Svelte and Vue, this also works in the template.

```vue
<template>
  <p>{{ $page.props.flash.error }}</p>

  <!-- @vue-expect-error - 'message' does not exist on flash -->
  {{ $page.props.flash.message }}
</template>
```

You can still pass an interface to `usePage()` for page-specific props, which will automatically be merged with the declared `sharedPageProps`.

```ts
type PageProps = {
  posts: { id: number; title: string }[]
}

const page = usePage<PageProps>()
const successMessage = page.props.flash.success
const postTitles = page.props.posts.map((post) => post.title)
```

In this case, `postTitles` will be typed as `string[]`.

For reference, previously, you could already partially do this by extending the `PageProps` interface.

```ts
type SharedPageProps = {
  flash: { error?: string; success?: string }
}

  interface PageProps extends PageProps, SharedPageProps {}
}
```

The problem with this approach is that the shared page props are overwritten when you pass a specific interface to `usePage()`.

```ts
type PageProps = {
  posts: { id: number; title: string }[]
}

const page = usePage<PageProps>()
const successMessage = page.props.flash.success // <== No longer typed! (shared props are lost)
const postTitles = page.props.posts.map((post) => post.title)
```

This was originally planned in PR #2549, but I decided to split it up into multiple PRs for clarity.


pascalbaljet:
I also brings `usePage()` to the Svelte adapter!

```svelte
<script lang="ts">
  import { usePage } from '@inertiajs/svelte'

  type PageProps = {
    posts: { id: number; title: string }[]
  }

  const page = usePage<PageProps>()

  $: error = $page.props.flash.error
  $: postTitles = $page.props.posts.map((post) => post.title)
</script>
```

PR #2638, opened 2025-10-16
Remove wrong `shouldIntercept()` call in `keydown` event handler in `<Link>` component

pascalbaljet:
The `keydown` event handler shouldn't call `shouldIntercept()`, because that's meant for mouse events.

This was originally planned in PR #2549, but I decided to split it up into multiple PRs for clarity.


pascalbaljet:
For reference, this is tested in `prefetch.spec.ts`:

* _can prefetch using link props_
* _can prefetch using link props with keyboard events_

PR #2650, opened 2025-10-21
refactor(core): replace `cloneDeep` with `structuredClone`

peaklabs-dev:
- Lodash is a large dependency that is no longer actively maintained see https://github.com/es-tooling/module-replacements/blob/main/docs/modules/README.md for this and more.
- This PR replaces `cloneDeep` with the native `structuredClone` function.
- A few Lodash functions remain `get`, `has`, `isEqual`, `set`, `escape`, `debounce` which could either be inlined or replaced with a smaller package. What do you think is the best thing to do with these?


pascalbaljet:
Just for reference: Lodash was replaced in #2210 and that was reverted in #2527.

peaklabs-dev:
Ah, never mind then. I also just noticed that `structuredClone` cannot clone functions, so this probably won’t work across the board. I’ll close it for now.

For v3, it would make sense to remove:
- `Lodash`
- `qs` https://github.com/es-tooling/module-replacements/blob/main/docs/modules/qs.md -> can probably be done now with a few lines.
- `axios`  https://github.com/es-tooling/module-replacements/blob/main/docs/modules/axios.md -> this is probably not possible.

pascalbaljet:
> For v3, it would make sense to remove:

Yep, got it on the list! Thanks!

PR #2655, opened 2025-10-23
Configure global defaults and update during runtime

### LINKED ISSUE #1669: How to increase recentlySuccessful time?
There is form prop `recentlySuccessful` which is set to true for 2 seconds. Is there any way to set this prop to different time?

pascalbaljet:
This PR brings support for settings configuration defaults, and the ability to update them during runtime. Currently, it supports three configuration settings, and this might be expanded in the future.

You can pass a default config object to the `createInertiaApp()` method. You don't have to pass all keys, just the ones you want to tweak.

```ts
createInertiaApp({
  // resolve, setup, etc.
  defaults: {
    form: {
      recentlySuccessfulDuration: 5_000,
    },
    prefetch: {
      cacheFor: 60_000,
    },
    visitOptions: (href: string, options: VisitOptions) => {
      return { headers: { ...options.headers, 'X-Foo': 'Bar' } }
    },
  },
})
```

The most interesting one here is the `visitOptions` callback, which acts more or less as an interceptor for Inertia requests, giving you the ability to change the options before the request is sent.

Each adapter now also exports a `config` instance to update the values during runtime. It supports the dot-notation.

```ts
import { config } from '@inertiajs/vue3'

config.set('form.recentlySuccessfulDuration', 1000)
```

You may also update multiple values at once by passing an object.

```ts
config.set({
  'form.recentlySuccessfulDuration': 1000,
  'prefetch.cacheFor': '5m',
})
```

Fixes #1669.

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -9,7 +9,6 @@ export { default as Form } from './Form'
 export { default as Head } from './Head'
 export { default as InfiniteScroll } from './InfiniteScroll'
 export { InertiaLinkProps, default as Link } from './Link'
-export * from './types'
 export {
   InertiaFormProps,
   SetDataAction,


danmatthews:
@pascalbaljet is there a way to make `visitOptions` here apply to `router.visit()` calls as well?

Issue #2108, opened 2024-11-26
[2.x] Prefetch doesn't load deferred props


javieralarcn10:
### Version:

- `@inertiajs/react` version: 2.0.0-beta.2

### Describe the problem:

When I prefetch a <Link prefetch/> it only loads props that do not have the Inertia::defer method. 

### Steps to reproduce:

Controller: 
`return inertia('Dashboard/Products/Inventory', [
            'shop' => $shop,
            'inventory' => Inertia::defer(fn() => $shop->inventory),
]);`
        
 Frontened:
 `<Link prefetch href={route("inventory")} className="flex gap-2 items-center">
   <span className="font-medium text-white">Inventory</span>
   </Link>`

mreduar:
That is not the objective of defer? That they only load when you enter the page?

RobertBoes:
I agree, it shouldn't prefetch deferred props, as that may take a long time while in the meantime the user could navigate, which would then cause the page to load very slowly

mathewparet:
It should load deferred props too, but as a second request (like how prefetch works normally).

For example, I assume that if a user is hovering over Paginator links chances are they want to click the next page or the page number over which they are hovering.

This works fine except if the page has deferred components, it looks like prefetch does prefetch the initial data, but doesn't seem to go back for the deferred data. The problem with this is now that if a user clicks on a page then nothing shows at all.

Oh! BTW, this issue exists in Vue also.

EmilMoe:
I agree with @mathewparet. At least there should be some kind of option to prefetch deferred 

```
prefetch="hover" :prefetch-deferred="['prop']"
```

Or something more suitable.

darkamenosa:
One more note on this, cache-for also should work similar for non deferred props. So when the user click back, they see rendered data.

pascalbaljet:
I'm closing this issue as deferred props are designed to be loaded after the subsequent page is loaded, so strictly speaking this is not a bug. I'm also not sure about the `prefetch-deferred` Link prop idea with the second request. Prefetch responses are usually quite fast, and you don't want network requests stacking up for all deferred prop groups for all prefetched pages.

PR #2656, opened 2025-10-23
Cancel in-flight deferred prop requests on navigation

alexspeller:
When users rapidly navigate between pages with deferred props, multiple requests fire in quick succession. Previously, deferred prop requests from previous navigations were not cancelled, causing them to complete out-of-order and display stale data.

This change cancels all in-flight async (deferred) requests when starting a new main visit, while preserving concurrent loading of deferred props within the same page load.

Changes:
- Fix RequestStream.cancelInFlight() to cancel ALL requests (not just first)
- Add cancellation logic in Router.visit() after onBefore check
- Add detection to distinguish deferred requests from main navigation
- Add tests suite for cancellation behavior

The cancellation happens after onBefore checks to avoid cancelling deferred props when navigation is prevented by user confirmation dialogs.

### DIFF
--- a/packages/core/src/eventHandler.ts
+++ b/packages/core/src/eventHandler.ts
@@ -92,8 +92,8 @@ class EventHandler {
           return
         }
 
-        // Cancel ongoing requests
-        router.cancelAll()
+        // Cancel ongoing requests except prefetch requests
+        router.cancelAll({ prefetch: false })
 
         currentPage.setQuietly(data, { preserveState: false }).then(() => {
           Scroll.restore(history.getScrollRegions())
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -176,9 +176,9 @@ export class Router {
     this.syncRequestStream.cancelInFlight()
   }
 
-  public cancelAll(): void {
+  public cancelAll({ prefetch = true } = {}): void {
     // Called on popstate navigation
-    this.asyncRequestStream.cancelInFlight()
+    this.asyncRequestStream.cancelInFlight({ prefetch })
     this.syncRequestStream.cancelInFlight()
   }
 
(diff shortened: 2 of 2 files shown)


pascalbaljet:
Thanks for this! I'm going to investigate it, though I must admit I'm a little scared about the 1000-line test file 😅

I haven't checked in detail yet, but is this related to #2108 as well?

alexspeller:
Yeah routing stuff always ends up with loads of edge cases (I did a bunch of work on the Ember.js router back in the day) so lots of tests ftw - a lot of them ended up being about the interaction with prefetch and deferred - there's just so many possible combinations of when things should vs shouldn't be cancelled. Hopefully the test cases are easy enough to follow even though there's loads of them!

Arguably the amount of test cases could be reduced, although each of them documents the desired behaviour given a complex scenario, so whilst it's a lot to read, I think it probably makes sense to see if at least on the face of it the title of each test case fits with the inertia routing vision - I think so but it's a lot easier to see that written out in a test case than infer it from the relatively minimal code changes here.

They might also expose some edge-case inconsistencies that are unrelated - for example [what happens if you pass both only and except to `visit`?](https://github.com/alexspeller/inertia/blob/388dff6f6937da933b8442fdc5f04890a727812d/tests/deferred-props-cancellation.spec.ts#L680-L722) - is that even a supported use case? Maybe it's not worth testing that as part of this PR, but I wanted to make sure that at least something sane would happen when doing so.

I don't think #2108 is related to this - unless this change fixes it by total coincidence. I would expect that it doesn't, as that issue is saying there should be _more_ deferrred requests in some circumstances (when preloading) and theoretically this PR should only result in _less_ (completed) deferred requests in the case that they are stale.

pascalbaljet:
Thanks again @alexspeller! I've refactored it and simplified the tests.

PR #2421, opened 2025-07-03
Add View Transition API Support to Inertia.js

Saad5400:
# Add View Transition API Support to Inertia.js

Disclaimer: I was unable to comprehensively test this feature. Some help in testing would be appreciated!

## Description

This pull request adds native support for the [View Transition API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API) to Inertia.js, enabling smooth animated transitions between page navigations while maintaining the framework's server-driven approach.

## Problem Statement

Currently, Inertia.js provides instant page transitions but lacks support for smooth animations between pages. Developers who want animated transitions must implement custom solutions that are complex, error-prone, and don't integrate well with Inertia's visit lifecycle.

## Solution

This implementation adds comprehensive View Transition API support through:

### Core Changes

1. **New Type Definitions** (`packages/core/src/types.ts`)
   - `ViewTransitionOptions` interface for configuration
   - Extended `Visit` type to include `viewTransition` property

2. **ViewTransitionManager** (`packages/core/src/viewTransition.ts`)
   - Browser feature detection with graceful fallbacks
   - Lifecycle management for view transitions
   - Error handling and callback execution

3. **Router Integration** (`packages/core/src/router.ts`)
   - `setDefaultViewTransition()` method for global configuration
   - Enhanced all visit methods (`get`, `post`, `put`, `patch`, `delete`, `visit`) to support view transition options
   - Proper merging of default and per-visit configurations

4. **Response Integration** (`packages/core/src/response.ts`)
   - Wrapped page setting logic with view transition API
   - Maintains full backward compatibility

5. **Event System** (`packages/core/src/events.ts`)
   - New events: `inertia:view-transition-start` and `inertia:view-transition-end`
   - Consistent with existing Inertia event patterns

### 1. **Enhanced User Experience**
- Smooth, native browser animations between pages
- Professional-looking transitions with minimal effort
- Improved perceived performance

### 2. **Progressive Enhancement**
- Works in modern browsers with View Transition support (Chrome 111+, Edge 111+, Safari 18+)
- Graceful fallback in older browsers (no transitions, but no breakage)
- Zero impact on existing applications

### 3. **Simple API**
```typescript
// Global configuration
router.setDefaultViewTransition({ enabled: true })

// Per-visit configuration
router.visit('/users', { 
  viewTransition: { enabled: true } 
})

// Framework integration
<Link href="/profile" viewTransition={{ enabled: true }}>
  Profile
</Link>
```

[...]

PR #2658, opened 2025-10-24
Support for View Transitions

pascalbaljet:
This PR adds support for [View Transitions](https://developer.chrome.com/docs/web-platform/view-transitions), building upon the great work done in #2421. It has some advantages over the original PR.

- Introduces no new interfaces/types
- Uses the standard [`ViewTransition` interface](https://developer.mozilla.org/en-US/docs/Web/API/ViewTransition)
- Works with Client-Side Visits
- Comes with tests

View transitions can be enabled by using the `viewTransition` option when making a visit.

```js
router.visit('/another-page', { viewTransition: true })
```

You may also pass a callback, which will receive the `ViewTransition` instance.

```js
router.visit('/another-page', {
  viewTransition: (transition) => {
    transition.ready.then(() => doSomething())
    transition.updateCallbackDone.then(() => doSomething())
    transition.finished.then(() => doSomething())
  },
})
```

It also works with the `Link` component.

```vue
<template>
  <Link href="/another-page" view-transition> Navigate </Link>

  <Link href="/another-page" :view-transition="(transition) => transition.finished.then(...)">
    Navigate
  </Link>
</template>
```

Combined with the new configuration feature from PR #2655, you may enable this globally.

```js
createInertiaApp({
  // resolve, setup, etc.
  defaults: {
    visitOptions: (href: string, options: VisitOptions) => {
      return { viewTransition: true }
    },
  },
})
```

### DIFF
--- a/packages/core/src/prefetched.ts
+++ b/packages/core/src/prefetched.ts
@@ -252,6 +252,7 @@ class PrefetchedRequests {
         'onCancelToken',
         'onPrefetching',
         'async',
+        'viewTransition',
       ],
     )
   }
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -235,7 +235,7 @@ export class Router {
 
   public prefetch(
     href: string | URL | UrlMethodPair,
-    options: Omit<VisitOptions, 'viewTransition'> = {},
+    options: VisitOptions = {},
     prefetchOptions: Partial<PrefetchOptions> = {},
   ) {
     const method: Method = options.method ?? (isUrlMethodPair(href) ? href.method : 'get')
@@ -249,6 +249,7 @@ export class Router {
       async: true,
       showProgress: false,
       prefetch: true,
+      viewTransition: false,
     })
 
     const visitUrl = visit.url.origin + visit.url.pathname + visit.url.search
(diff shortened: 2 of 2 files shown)


dnlmr:
Incredibly awesome! I have just seen your tweet and added it to my app in less than 5 minutes. Pure magic! 🎉

For anyone wondering... if you want to activate it globally you can bypass view transitions for infinite scroll (tables!) like this:
```js
createInertiaApp({
  // resolve, setup, etc.
  defaults: {
    visitOptions: (href, options) => {
      // Bypass view transitions for infinite scroll requests
      const isInfiniteScroll = Boolean(
        options.headers?.['X-Inertia-Infinite-Scroll-Merge-Intent'] ??
          options.headers?.['x-inertia-infinite-scroll-merge-intent']
      );
      
      return {
        viewTransition: !isInfiniteScroll,
      };
    },
  },
})
```

PS: @pascalbaljet you have been on 🔥 for quite some time now!!! 
Infinite scroll was another incredibly addition to Inertia! Love it! 🥰

pascalbaljet:
> Incredibly awesome! I have just seen your tweet and added it to my app in less than 5 minutes. Pure magic! 🎉
> 
> For anyone wondering... if you want to activate it globally you can bypass view transitions for infinite scroll (tables!) like this:

Very nice! Maybe you can also check for `options.preserveState` instead of the header.

dnlmr:
You are absolutely right! 😎

Using `viewTransition: !options.preserveState` makes **MUCH** more sense!

PR #2660, opened 2025-10-26
feat: add useParams hook

SantosVilanculos:
This PR adds **`useParams`** hook that returns an object of key/**value pairs** of the dynamic params from the current URL.

## My motivation

**I** started from opening a [discussion](https://github.com/inertiajs/inertia/discussions/2564) on implementing an **`useRoute`** hook. But after **careful** exploration and consideration, **I** realized that:

  - Route name is a Laravel and some **framework-specific** thing, with the vast Inertia core and community drivers (php, django,...) and now with [Wayfinder](https://github.com/laravel/wayfinder) (on the Laravel side at least), **I** decided not to.
  - Query string **handling** is done mostly with the **`window.location.search`** api or a library like [nuqs](https://nuqs.dev/).
  - And, I think there is no need to wrap **other** window apis **shown** in the [discussion](https://github.com/inertiajs/inertia/discussions/2564) in a hook, unless you are extending them (not in that case).

## Why this feature **should** be **implemented**

A dedicated **`useParams`** hook would significantly improve the developer experience (DX) by bringing idiomatic frontend patterns to Inertia applications. While route parameters are technically available as component props via the server, forcing developers to access them through props breaks the standard mental model established by virtually all popular JavaScript routing libraries (like [React Router](https://reactrouter.com/api/hooks/useParams), [Vue Router](https://router.vuejs.org/guide/advanced/composition-api.html), [Tanstack Route](https://tanstack.com/router/v1/docs/framework/react/api/router/useParamsHook), [Next.js](https://nextjs.org/docs/app/api-reference/functions/use-params), [Nuxt.js](https://nuxt.com/docs/4.x/api/composables/use-route), ...).

Since every supported backend driver already has a simple, explicit way to pass these route parameters as component props, the data is always consistently available, making it trivial for a native client-side hook to expose it.

A **`useParams`** hook would allow developers to access this consistently available route data from any nested component or utility hook without the manual toil of **prop-drilling** or creating boilerplate context providers. This standardization would make Inertia apps feel more like traditional SPAs, lowering the learning curve for developers transitioning from client-side routers.

## Backend implementation

- **`inertia-laravel`**: https://github.com/inertiajs/inertia-laravel/pull/785
- **`inertia-rails`**: ?
- **`inertia-django`**: ?
- **`inertia-phoenix`**: ?
 
Anyone interested is free to contribute to the existing drivers or open new Pull Requests with improvements

## Why **`Record<string, string>`** and not **`Record<string, string | number>`** or **`Record<string, any>`** for a parameter type?

URL segments are inherently strings. Even if a route parameter in the URL represents a number (like `/users/123`), the underlying value retrieved by a direct URL inspection is the string **`"123"`**.


pascalbaljet:
I'm closing this PR as I think this is not a responsibility of the Inertia library/protocol. If you want to share route parameters, you can always send it as shared data, for example, in the Laravel middleware, and then create your own hook on the frontend.

```php
class HandleInertiaRequests extends Middleware
{
    public function share(Request $request): array
    {
        return array_merge(parent::share($request), [
            'routeParameters' => $request->route()?->originalParameters(),
        ]);
    }
}
```

PR #2665, opened 2025-10-28
fix: preserve query parameters in test server responses

alexspeller:
Found this whilst looking at the other recent deferred props issue, this one is nice and small though!

The test server was using req.path instead of req.originalUrl, which stripped query parameters from the URL

Just a test server issue in this case, there was already a TODO in the code, but I think the extra tests are worthwhile to test that the correct query params are preserved for deferred requests?

### DIFF
--- a/tests/app/server.js
+++ b/tests/app/server.js
@@ -1562,10 +1562,7 @@ app.post('/form-component/events/errors', async (req, res) =>
 )
 
 app.get('/form-component/headers', (req, res) => inertia.render(req, res, { component: 'FormComponent/Headers' }))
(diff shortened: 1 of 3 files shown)


alexspeller:
Ah I thought this would be a small one, but it breaks a lot of existing tests that are relying on the incorrect behaviour of the test server - I'm happy to fix but it would be a massive diff, let me know if you're interested in proceeding

alexspeller:
Well if you want it I think this is all the changes that are needed

pascalbaljet:
Thanks @alexspeller, this was on my list for a long time!

PR #2671, opened 2025-10-29
[2.x] Add Form Context Support

laserhybiz:
This PR adds the ability to inject form state and methods into descendant components using Vue's provide/inject API.

This is useful when using wrapper components which encapsulates form functionalities.

Before:
Form state is only accessible via slot props or template refs and has to be passed down to children
```vue
<!-- Parent.vue -->
<Form
  action="/users"
  method="post"
  #default="{
    errors,
  }"
>
  <Input name="name" :error="errors.name" />
</Form>

<!-- Input.vue -->
<script setup>
defineProps({
  name: {
    type: String,
    required: true,
  },
  error: {
    type: String,
    default: null,
  },
})
</script>

<template>
  <input type="text" :name="name" />

  <div v-if="error">
    {{ error }}
  </div>
</template>
```

After:
All form state is accessible in any descendant component
```vue
<!-- Parent.vue -->
<Form
  action="/users"
  method="post"
>
  <Input name="name" />
</Form>

<!-- Input.vue -->
<script setup>
import { useFormContext } from '@inertiajs/vue3'

defineProps({
  name: {
    type: String,
    required: true,
  },
})

const form = useFormContext()
</script>

<template>
  <input type="text" :name="name" />

  <div v-if="form.errors[name]">
    {{ form.errors[name] }}
  </div>
</template>
```

 The `useFormContext` hook returns the same interface as the Form's slot props and template ref, including all state (errors, isDirty, processing, etc.) and methods (submit, reset, clearErrors, etc.).

 I'll add the tests if this is approved.

### DIFF
--- a/packages/react/test-app/Pages/FormComponent/Context/MethodsTestComponent.tsx
+++ b/packages/react/test-app/Pages/FormComponent/Context/MethodsTestComponent.tsx
@@ -75,9 +75,6 @@ export default () => {
       <button type="button" onClick={() => form.resetAndClearErrors('name')}>
         resetAndClearErrors('name')
       </button>
-      <button type="button" onClick={() => form.defaults()}>
-        defaults()
-      </button>
 
       <button type="button" onClick={testGetData}>
         getData()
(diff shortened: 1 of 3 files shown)


laserhybiz:
I guess I will also have to add this to the other adapters if it's approved.

laserhybiz:
@pascalbaljet is this something you’d consider merging?

pascalbaljet:
Yes, I like it! Could you add tests and implement it for the other adapters as well? Thanks!

laserhybiz:
@pascalbaljet I’ve implemented this for the React adapter and added tests for both Vue and React. the tests are currently failing because the svelte implementation is still missing.

I don’t have any experience with svelte, could you help with that part? otherwise, I’ll try to make some time to dig into it myself.

pascalbaljet:
Thanks again @laserhybiz!

PR #2672, opened 2025-10-29
Pass `loading` state as property to `Deferred` children

scajal:
This PR updates the `Deferred` component so it will pass the current loading state to the children, when children is a function. It also makes the `fallback` property optional on the component.

This allows the developer to, for example, show a skeleton while the deferred props are being loaded.

Here's a react example:
```jsx
<Deferred data="users">
{({ loading }) => (
    <form>
        <select disabled={loading}>
            {users?.map(user => <option key={user.id} value={user.id}>{user.name}</option>)}
        </select>
    </form>
)}
</Deferred>
```


pascalbaljet:
I'm fine with passing the `loading` state to the children/slot, but I think making the `fallback` optional and removing the error is a breaking change. We could make that change in a next major release, though. Could you update the PR to just expose the `loading` state?

scajal:
@pascalbaljet With a required fallback this wouldn't make any sense. The `fallback` would always be rendered if present and the data is still not loaded, so the `children` would only be rendered after loading the data and the `loading` property would always be true.

scajal:
Implemented it with a custom wrapper component.

PR #2680, opened 2025-11-03
Switch `useContext` to `use` in `usePage()` hook

HichemTab-tech:
### Description
Changed context consumption in the `usePage()` hook from using React’s `useContext` to the new `use` API (added in React 19) as per the documentation here: [https://react.dev/reference/react/use#reading-context-with-use](https://react.dev/reference/react/use#reading-context-with-use) :
> `use` is preferred over `useContext` because it is more flexible.

This lets us call `usePage().props` from anywhere, even conditionally.

### DIFF
--- a/packages/react/src/usePage.ts
+++ b/packages/react/src/usePage.ts
@@ -3,7 +3,7 @@ import React from 'react'
 import PageContext from './PageContext'
 
 export default function usePage<TPageProps extends PageProps = PageProps>(): Page<TPageProps & SharedPageProps> {
-  // backwards compatibility for React <19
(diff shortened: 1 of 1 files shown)


pascalbaljet:
We're still supporting React 16.9+ in Inertia 2, so we should use it conditionally when it's available. Can you refactor it to only use `use` when it's available? Thanks!

https://github.com/inertiajs/inertia/blob/4e30447d10f60013de0006782b86a588ede72fd6/packages/react/src/Form.ts#L27-L30

HichemTab-tech:
Thanks for teh feedback!, I made the changes to support React <19.

pascalbaljet:
Thanks!

PR #2682, opened 2025-11-05
fix(core): update history state after popstate to keep page data in sync

### LINKED ISSUE #2651: [2.x] Browser "Back" & "Forward" button does not return to navigate to the correct page
### @inertiajs/react Version

2.2.9

### Backend stack (optional)

Laravel 12.x
PHP 8.4

### Describe the problem

When navigating using the browser navigation buttons the page is not being rendered correctly.

https://github.com/user-attachments/assets/9c11a4fd-dbfc-43fd-b298-86773f0daf14

### Steps to reproduce

I spun up a brand new instance of the Laravel inertia react starter kit, then the video provided shows the steps I took

HichemTab-tech:
Fixes #2651 .

This fixes the issue where, after using the browser’s Back or Forward buttons, the page data inside the **[history class](https://github.com/inertiajs/inertia/blob/4e30447d10f60013de0006782b86a588ede72fd6/packages/core/src/history.ts)** was not being updated correctly (or as i think so).

I added a small change to update the history state using `replaceState` after a `popstate` event, it updates the page data but leaves the URL untouched (since the browser already updates that part in this popstate case).

```javascript
const { url, ...pageWithoutUrl } = currentPage.get()
history.replaceState({ ...pageWithoutUrl })
```

This fixed the problem in my testing.

### Note

I’m not sure **at all** if this change is completely safe or if it's even correct since I don’t have much experience with Inertia’s navigation internals, so I’d really appreciate feebdack on whether this change makes sense.

### DIFF
--- a/packages/core/src/eventHandler.ts
+++ b/packages/core/src/eventHandler.ts
@@ -100,8 +100,8 @@ class EventHandler {
           fireNavigateEvent(currentPage.get())
         })
 
-        const {url, ...pageWithoutUrl} = currentPage.get()
-        history.replaceState({...pageWithoutUrl} as Page)
+        const { url, ...pageWithoutUrl } = currentPage.get()
+        history.replaceState({ ...pageWithoutUrl } as Page)
       })
       .catch(() => {
         this.onMissingHistoryItem()


pascalbaljet:
I'm converting it to draft as quite some tests are failing. Can you also add a test for this? (that would fail without this change)

pascalbaljet:
I'm closing this pull request due to lack of activity. If you're still working on this or have updates to share, feel free to open a new PR. We'd be happy to take another look. Thanks!

PR #2687, opened 2025-11-07
Optimize page data size and parsing (37% size reduction!)

bram-pkg:
This PR adds support for the page data to be loaded from an `app_page` `<script type="application/json">` element, this saves the SSR daemon quite a lot of type on big payloads because it not longer has to escape the HTML using `&quot;` in the JSON.

It also saves a lot of page size on big payloads.

One of our pages with currently **343kb** in props, went down to **219kb** using this patch. That is a **37% reduction!**

I have tested the changes using the SSR setup in the playgrounds. The changes are also backwards compatible, because the code gracefully falls back to using the `data-page` attribute.

In our profiler dump for the Node SSR process, escapeHtml takes up a non-insignificant of the profile with big payloads:

image

Inspired by my colleague @thomas-brx: https://github.com/inertiajs/inertia/discussions/1960#

Screenshots from the playgrounds working, note the `app_page` element:

<details><summary>Vue</summary>
<p>

image

</p>
</details> 

<details><summary>Svelte</summary>
<p>

image

</p>
</details> 

<details><summary>React</summary>
<p>

image

</p>
</details>

### DIFF
--- a/tests/ssr.spec.ts
+++ b/tests/ssr.spec.ts
@@ -32,23 +32,21 @@ test.describe('SSR', () => {
     })
   })
 
-  test.describe('initial page load with script element', () => {
-    test('embeds page data in a script element instead of data-page attribute', async ({ page }) => {
-      const response = await page.request.get('/ssr/page-with-script-element')
-      const html = await response.text()
+  test('embeds page data in a script element instead of data-page attribute', async ({ page }) => {
+    const response = await page.request.get('/ssr/page-with-script-element')
+    const html = await response.text()
 
-      expect(html).toContain('data-page="app"')
-      expect(html).toContain('<script data-page="app" type="application/json">')
-      expect(html).toContain('Hello from script element!')
+    expect(html).toContain('data-page="app"')
+    expect(html).toContain('<script data-page="app" type="application/json">')
+    expect(html).toContain('Hello from script element!')
 
-      await page.goto('/ssr/page-with-script-element')
-      const scriptContent = await page.locator('script[data-page="app"]').textContent()
-      expect(JSON.parse(scriptContent || '')).toMatchObject({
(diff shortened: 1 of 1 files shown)


bram-pkg:
@pascalbaljet it would be greatly appreciated if you could take a look at this and review, it would lead to a decent performance improvement for SSR and also a payload size reduction.

pascalbaljet:
Awesome! I'll look into this soon.

bram-pkg:
Seems unsuitable, they themselves say that non uniform and deeply nested JSON is more efficient if that is required, and since Inertia props tend to fall into those criteria, let's just stick with a native browser API. No need to overcomplicate things.

pascalbaljet:
I'll refactor it a bit to put it behind a configuration flag. While it's backwards compatible, there's a possibility that people are relying on the `data-page` attribute in E2E and snapshot testing.

bram-pkg:
That makes sense. Thanks! Looks good already.

bram-pkg:
Do you need anything from me still, @pascalbaljet?

pascalbaljet:
> Do you need anything from me still, @pascalbaljet?

No, thanks! I'll work some more on this by adding tests, and then I'll merge it.

pascalbaljet:
I've added tests (after merging #2730) and refactored the `id` attribute away as I didn't really like the hardcoded `_page` suffix. So now the `script` element is rendered as:

```html
<script data-page="app" type="application/json">...</script>
```

bram-pkg:
That does look a lot nicer and cleaner. Great idea! Thanks for helping with the tests @pascalbaljet

bram-pkg:
Amazing! Thank you Pascal 💖

thomas-brx:
> Amazing! Thank you Pascal 💖

And thank you Bram for turning this idea into a PR!

PR #2689, opened 2025-11-10
Fix `InfiniteScroll` scroll preservation

skryukov:
This PR fixes `InfiniteScroll` scroll position preservation for cases when the first child element is invisible.

Before these changes, `captureScrollPosition` always called `getElementsInViewportFromCollection` with the first child element. When this element was invisible, `getElementsInViewportFromCollection` always returned an empty collection.


pascalbaljet:
Thanks!

PR #2690, opened 2025-11-10
Fix array keys misalignment in form data and query

skryukov:
This PR is another attempt to fix the issue described in https://github.com/inertiajs/inertia/pull/1901.

This PR uses a different approach, considering the current behavior as a bug. The introduction of default settings makes it super easy to switch back to `indices` if required.

### Rationale

Currently, multipart form data arrays are always serialized using indices:

```
------WebKitFormBoundarynVnO6UtO8ts1Qa5q
Content-Disposition: form-data; name="model[files][0]"; filename="aliens.png"
Content-Type: image/png


------WebKitFormBoundarynVnO6UtO8ts1Qa5q
Content-Disposition: form-data; name="model[files][1]"; filename="ufo.jpg"
Content-Type: image/jpeg
```

Rails (or more precisely [Rack](https://github.com/rack/rack/blob/main/lib/rack/query_parser.rb#L116-L122)) parses this as hashes with keys `'0'` and `'1'`. To support scenarios like a file input with the `multiple` attribute, we need to allow tweaking this behavior.

This PR makes Inertia.js to apply format from `queryStringArrayFormat` option to FormData:

```js
  form.post(`/models`, {
    queryStringArrayFormat: 'brackets',
  })
```

```
------WebKitFormBoundaryxrZa60F5B7gxV5aG
Content-Disposition: form-data; name="model[files][]"; filename="aliens.png"
Content-Type: image/png


------WebKitFormBoundaryxrZa60F5B7gxV5aG
Content-Disposition: form-data; name="model[files][]"; filename="ufo.jpg"
Content-Type: image/jpeg
```

### DIFF
--- a/tests/form-component.spec.ts
+++ b/tests/form-component.spec.ts
@@ -4,7 +4,6 @@ import { consoleMessages, pageLoads, requests, scrollElementTo, shouldBeDumpPage
 test.describe('Form Component', () => {
   test.describe('Elements', () => {
     test.beforeEach(async ({ page }) => {
-      pageLoads.watch(page)
       await page.goto('/form-component/elements')
     })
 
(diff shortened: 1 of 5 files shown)


tests/form-component.spec.ts
```diff
@@ -94,7 +94,9 @@ test.describe('Form Component', () => {
         token: 'abc123',
         age: '30',
         user: { address: { street: '123 Main St' } },
-        items: [{ name: 'Item 1' }, { name: 'Item 2' }],
+        // Express doesn't support brackets format by default
```
pascalbaljet: Why did this work before?

pascalbaljet:
Thanks! I really like it and refactored it a little bit for readability. The only thing I'm nervous about is the change in `form-component.spec.ts`.

tests/form-component.spec.ts
pascalbaljet: Awesome, thanks!

pascalbaljet:
For reference, when dumping the incoming request data in Laravel, this ends up being the same.

```js
const submitUsingBrackets = () => {
    const formData = new FormData();

    formData.append('items[][name]', 'Item 1');
    formData.append('items[][name]', 'Item 2');
    formData.append('documents[]', new File([], 'doc1.pdf'));
    formData.append('documents[]', new File([], 'doc2.pdf'));

    router.post('/dump', formData);
};

const submitUsingIndices = () => {
    const formData = new FormData();

    formData.append('items[0][name]', 'Item 1');
    formData.append('items[1][name]', 'Item 2');
    formData.append('documents[0]', new File([], 'doc1.pdf'));
    formData.append('documents[1]', new File([], 'doc2.pdf'));

    router.post('/dump', formData);
};
```

pascalbaljet:
Thanks again, @skryukov!

gerardnll:
This is breaking my app when posting a form with files so the data is send as `multipart/form-data`.

Laravel is not applying validation of arrays because how data is sent:

Content-Disposition: form-data; name="ticket_types[][name]" <-- it's missing the index.

The previous release used to send data like this:

Content-Disposition: form-data; name="ticket_types[0][name]" <-- it has the index

pascalbaljet:
> This is breaking my app when posting a form with files so the data is send as `multipart/form-data`.
> 
> Laravel is not applying validation of arrays because how data is sent

Could you share your validation logic/rules?

gerardnll:
Sure but It was working until now and they seem fine, rolling back to .15 solved it. I have simplified them:

```
return [
    'name' => ['required', 'string', 'max:80'],
    'image' => ['nullable', 'image', 'mimes:jpg,png,bmp', 'max:10000', Rule::dimensions()->minWidth(1350)->minHeight(500)->maxWidth(3000)->maxHeight(3000)],
    'ticket_types' => ['required', 'array', 'min:1'],
    'ticket_types.*.name' => ['required', 'string', 'max:60'],
    'ticket_types.*.prices' => ['required', 'array', 'min:1'],
    'ticket_types.*.prices.*.base_amount' => ['required', 'numeric', 'min:0', 'max:1000000'], // cents
];
```

I'm transforming data before sending:

```
transform((data) => ({
    ...data,
    ticket_types: cloneDeep(data.ticket_types).map((ticketType) => {
        ticketType.prices = ticketType.prices.map((price) => ({
            ...price,
            base_amount: Number(price.base_amount) * 100, // Convert to cents
        }));
        return ticketType;
    }),
}));

post()...
```

pascalbaljet:
@skryukov @gerardnll I've put the changes behind a config flag (see #2701), so updating to v2.2.17 should restore the original functionality. I'll investigate the validation and add docs later on! Thanks for sharing!

gerardnll:
Thanks to you guys for the hard work! ;)

skryukov:
Global config is such a great addition to Inertia, thanks @pascalbaljet!

PR #2691, opened 2025-11-10
Export page component type for React adapter

skryukov:
This PR exports the page component type for use with `import.meta.glob` and default layouts.

Note: I've used `ResolvedComponent` to align with the Svelte adapter.


pascalbaljet:
Thanks!

PR #2707, opened 2025-11-19
fix(core): Fix issue where anchor scroll randomly doesn't work when React page loads

mark-youssef-bitwarden:
## Problem

When navigating to a (React) page with an anchor ID, sometimes the anchor scroll doesn't work and the scroll position stays at the top of the page.

It seems like the scroll util isn't catching the anchor element in those cases because the React page hasn't fully rendered yet. I noticed there a check for the anchor element in a non-blocking thread using setTimeout, but it's still not quite consistent.

## Solution

My attempt instead wraps the anchor check in a double requestAnimationFrame to wait for next render cycle before checking for the anchor element.

### DIFF
--- a/packages/core/src/scroll.ts
+++ b/packages/core/src/scroll.ts
@@ -35,11 +35,12 @@ export class Scroll {
     this.save()
 
     if (anchorHash) {
-      // We're using a setTimeout() here as a workaround for a bug in the React adapter where the
-      // rendering isn't completing fast enough, causing the anchor link to not be scrolled to.
-      setTimeout(() => {
-        const anchorElement = document.getElementById(anchorHash.slice(1))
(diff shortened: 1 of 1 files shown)


pascalbaljet:
Is this on the initial page load, or on subsequent requests? There's a fix in v2.2.19 for scrolling to an element on the initial page load.

mark-youssef-bitwarden:
The issue I was seeing was mainly regarding subsequent page load, but I discovered a large amount of our scroll issues are due to other factors (other libraries calling `scrollTo` and layout shift from loading elements). I resolved most of our issues, but I also found that changing scroll behavior to `"instant"` produced more consistent results across browsers. Having the option to set the scroll behavior in inertia would be a nice to have.

Anyways I'll close this PR since it's not necessary.

PR #2709, opened 2025-11-19
Use FormValue in Form component

skryukov:
This PR replaces hardcoded error type with one from global config.


pascalbaljet:
Thanks!

PR #2712, opened 2025-11-21
[3.x] Drop support for React < 19

pascalbaljet:
Drops support for React 16 and 17.

Maybe we even should drop support for React 18. The latest *real* release was in April 2022, and they usually drop support when a new major version has been released: https://github.com/reactjs/react.dev/issues/1745#issuecomment-466767389

Also, we now have two workarounds to support React < 19. It would be nice to remove those.


pascalbaljet:
I've updated the PR to also drop support for React 18, as React 19 has been out for a while now.

PR #2722, opened 2025-11-27
Fix brackets notation qs parsing

skryukov:
The PR replaces `parseArrays: false` with `arrayLimit: 0` to fix an issue where array values using bracket notation (e.g., `frameworks[]=react&frameworks[]=vue`) were being incorrectly parsed. This change ensures that bracket notation arrays are properly handled while disabling indexed array parsing.


pascalbaljet:
Thanks! I've added some tests and lowered it to `-1`, which solved a backward compatibility issue.

PR #2723, opened 2025-11-27
React: fix layout shortcut

skryukov:
This PR adds tests for layout shortcuts in React adapter and also fixes the `Component.layout = MyLayout` shortcut (don't like the `try`/`catch` as a fix, but I don't think there's a better solution for functional components).

Note: both svelte and vue adapters already support and test the shortcuts.

### DIFF
--- a/packages/react/src/App.ts
+++ b/packages/react/src/App.ts
@@ -10,7 +10,7 @@ import {
 import { createElement, FunctionComponent, ReactNode, useEffect, useMemo, useState } from 'react'
 import HeadContext from './HeadContext'
 import PageContext from './PageContext'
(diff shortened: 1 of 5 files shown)


pascalbaljet:
Thanks for this! I'm converting it to draft because it doesn't work for layouts without hooks, like the one from the docs.

```tsx
import { Link } from '@inertiajs/react'

export default function Layout({ children }) {
    return (
        <main>
            <header>
                <Link href="/">Home</Link>
                <Link href="/about">About</Link>
                <Link href="/contact">Contact</Link>
            </header>
            <article>{children}</article>
        </main>
    )
}
```

Also, I'm not too fond of the try/catch solution either.

pascalbaljet:
I'm closing this PR in favor of #2859.

PR #2731, opened 2025-12-02
Expose internal axios instance to consumers

### LINKED ISSUE #2715: Inertia's axios instance not accessible

- [x] React
- [x] Vue 3
- [x] Svelte
- [ ] Not Applicable

### JS package version

2.2.18

### Backend stack (optional)

_No response_

### Describe the problem

Reporting this as a bug, since Pascal may be under the impression that [inertia is using the global axios instance](https://github.com/inertiajs/inertia/pull/2104#issuecomment-2969560453), which is not the case.

Inertia is [creating it's own instance](https://github.com/inertiajs/inertia/blob/11930ca1006fc329b8b70b9c589dcb7944878928/packages/core/src/request.ts#L43), but not exporting that instance for consumers to use.

Even attempting to resolve axios packages to the exact same version will not guarantee that a _global_ axios instance will be shared.

Thus far, I've used some brittle workarounds that involve my bundler pulling the axios dependency from the internal inertia node_modules folder, but it breaks from package manager to package manager, depending on duplicate dependency methodologies.

### Steps to reproduce

- Ensure there's only one version of axios installed in your project (via lockfile)
- Import axios locally in your project
- Attach an [interceptor](https://axios-http.com/docs/interceptors) and modify the request somehow (maybe throw a debugger in there)
- Submit a request and attempt to hit the debugger or observe modified request

chasegiunta:
This pull request adds support for importing the axios instance from the React, Svelte, and Vue 3 Inertia adapters.

With this change, consumers of each adapter can access the exact same axios instance used internally.

This ensures advanced use cases—such as adding interceptors or custom configuration—maintain consistent behavior and prevents issues that can occur due to axios version mismatches or package manager deduplication strategies.

Fixes #2715


pascalbaljet:
I'm closing this one as I don't want to export Axios from Inertia. That's not its responsibility and not something we want to maintain. I'm open to investigating further (see #2715), but I need clear steps (or demo repo) to reproduce the actual issue first.

PR #2748, opened 2025-12-11
bugfix(whenVisible-vue): Fix loaded state when data already exists

ClaraLeigh:
When the data already exists, there is nothing that either triggers the re-loading or marks the data as loaded already.

This bugfix resolves this issue by setting the loaded value to true when the data already exists.

I am currently experiencing this issue when the data already exists in history and the back button is used for navigation.


pascalbaljet:
Thanks! I'll add tests and check if this is also a problem with the other adapters.

PR #2751, opened 2025-12-11
Escape forward slashes when using useScriptElementForInitialPage

kirk-loretz-fsn:
Followup of #2687.

Prevents the HTML parser from prematurely closing the script element when `</script>` is in a prop value.


pascalbaljet:
Thanks!

PR #2756, opened 2025-12-12
Expose InertiaPrecognitiveForm type

lcdss:
This only exposes the new InertiaPrecognitiveForm types so we can easily use it when passing the form instance as a param to other functions:

```ts
export const UserForm = ({ form, onSubmit }: { form: InertiaPrecognitiveFormProps<UserFormData>, onSubmit: () => void }) => {
  //
}
```


pascalbaljet:
Thanks!

PR #2769, opened 2025-12-16
Add support for protocol-relative urls in url.ts

### LINKED ISSUE #2768: Subdomain gets lost by mergeDataIntoQueryString() when using protocol-relative URLs

- [x] React
- [x] Vue 3
- [x] Svelte
- [ ] Not Applicable

### JS package version

2.1.4

### Backend stack (optional)

_No response_

### Describe the problem

I'm developing a multi sub-domains application with Inertia, and experienced an issue with React, but I think it affects all packages.

I have the following setup for my login/route:

```bash
php artisan route:list|grep login
  GET|HEAD        dashboard.mydomain.lcl/login ......... login › Laravel\Fortify › AuthenticatedSessionController@create
```

Wayfinder correctly generates the route helper:

```ts
export const login = (options?: RouteQueryOptions): RouteDefinition<'get'> => ({
    url: login.url(options),
    method: 'get',
})

login.definition = {
    methods: ["get","head"],
    url: '//dashboard.mydomain.lcl/login',
} satisfies RouteDefinition<["get","head"]>

```

Yet, when I'm `mydomain.lcl` and use `<Link href={login()}>Log in</Link>` or  `<Link href={login().url}>Log in</Link>` I get the following link generated, without the subdomain:

```html
<a href="/login">Log in</a>
```

Using a regular `<a />` instead of `<Link />` yields the expected output.

Using the debugger, I was able to pinpoint the exact problem to this test in `mergeDataIntoQueryString`:

https://github.com/inertiajs/inertia/blob/7089a8e81992617c23efc88e27a9f3637f8d79d5/packages/core/src/url.ts#L54C19-L54C33

```ts
 const hasHost = urlHasProtocol(href.toString())
```

this gets set to `false` for a protocol-relative URL. The fact that has*Host* is checking for *Protocol* is in fact the problem, since it should really be checking for the host instead.

### Steps to reproduce

Already explained above

machour:
Fixed #2768 by updating the way we check if a given `href` has a defined host in it.

This allow using protocol-relative urls (`//domain.com`) without Inertia losing the host.

Looking at the codebase, I figured we really wanted `urlHasHost()` instead of `urlHasProtocol()` so I went ahead and swapped the functions everywhere. And added a few tests for this new function.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -13,7 +13,14 @@ export { shouldIntercept, shouldNavigate } from './navigationEvents'
 export { hide as hideProgress, progress, reveal as revealProgress, default as setupProgress } from './progress'
 export { resetFormFields } from './resetFormFields'
 export * from './types'
-export { hrefToUrl, isUrlMethodPair, mergeDataIntoQueryString, urlHasProtocol, urlToString, urlWithoutHash } from './url'
+export {
+  hrefToUrl,
+  isUrlMethodPair,
+  mergeDataIntoQueryString,
+  urlHasProtocol,
+  urlToString,
+  urlWithoutHash,
+} from './url'
 export { type Config, type Router }
 
 export const router = new Router()


machour:
NB: I'm actually seeing a CORS issue when trying to click the link in my app, but I guess that's a different issue, and this PR still fixes the url generation part.

Would appreciate any pointer for the CORS issue, meanwhile, I'll resort to a good ol `<a />`

pascalbaljet:
Thanks! I've reverted the removal of `urlHasProtocol` (because that would be a breaking change) and added the logic to it.

machour:
hadn't thought of the BC breakage, thanks!

PR #2784, opened 2025-12-30
Refresh the Infinite Scroll triggers after partial reload

### LINKED ISSUE #2781: Using Infinite scroll with forms, breaks resetting
### @inertiajs/vue3 Version

2.3.4

### Backend stack (optional)

Laravel 12.x
Laravel Scout (Typesense)
FrankenPHP

### Describe the problem

When you first open the page with a Infinite scroll, it works fine. I get all the times on first page enter (e.g. 12). When I change a form value like sort, the number of items is incorrectly returned (e.g. 4).

It seems to be server related, as it indeed only passes 4, instead of the expected 12 items.

### Steps to reproduce

1. First open
2. Filter/sort/search/reset 
3. Limit subset returned

pascalbaljet:
Fixes #2781

### DIFF
--- a/packages/core/src/infiniteScroll.ts
+++ b/packages/core/src/infiniteScroll.ts
@@ -1,4 +1,5 @@
 import { requestAnimationFrame } from './domUtils'
+import { router } from './index'
 import { useInfiniteScrollData } from './infiniteScroll/data'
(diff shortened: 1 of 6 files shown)


francoism90:
@pascalbaljet Thanks for the fix!

PR #2787, opened 2025-12-30
fix: update has more state when resetting before updating page

AydinHassan:
Hey,

I noticed an issue with infinite scroll in manual mode and filtering. When I filtered using `only` and `reset` - the `hasMore` value is not correctly updated. Eg if I filter and there are no more pages, when I remove the filter (where there should be multiple pages), `hasMore` would still return `false`.

I _think_ that `resetState` was only being called after the prop was updated and the page rerendered, and therefore the scroll prop state was being used for the old page. I updated so that `scrollProp.reset` is checked on `beforeUpdate` and use it to reset the state before we rerender.

I added some tests to hopefully show the issue, maybe you would like to fix it a different way. Regardless, the fix seems to work in my project!

### DIFF
--- a/packages/core/src/infiniteScroll.ts
+++ b/packages/core/src/infiniteScroll.ts
@@ -45,6 +45,7 @@ export default function useInfiniteScroll(options: UseInfiniteScrollOptions): Us
       options.onCompleteNextRequest()
       requestAnimationFrame(() => elementManager.processServerLoadedElements(loadedPage), 2)
     },
+    onReset: options.onDataReset,
   })
 
   const addScrollPreservationCallbacks = (reloadOptions: ReloadOptions): ReloadOptions => {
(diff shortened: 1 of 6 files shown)


pascalbaljet:
I've converted this to draft due to the failing tests.

AydinHassan:
hey @pascalbaljet most tests are green now apart from the Svelte ones, I don't have any experience with Svelte - do you have any suggestions? Maybe my overall solution is also not how you'd approach it? 

I can get it to work by adding the page prop as a dependency like:

```
hasPrevious: ($page, infiniteScrollInstance?.dataManager.hasPrevious()) || false,
```

But it seems a bit like a hack to me, but again, I know nothing about Svelte. 

edit: I added the example fix as the last commit

pascalbaljet:
I've refactored it a bit so we keep using the `success` event, as I'm afraid of side effects by changing that. Now it signals from the Data Manager to update the component state when the data is reset.

AydinHassan:
@pascalbaljet nice, that makes sense, thanks

pascalbaljet:
Thanks for the PR @AydinHassan!

PR #2791, opened 2025-12-31
[3.x] Remove qs dependency

pascalbaljet:
This PR replaces the `qs` package with an internal `queryString.ts` implementation. By removing the dependency, we achieve a smaller bundle, have one less package to maintain/audit, and gain full control over the parsing/stringifying behavior.

Many tests were already added in #2417, but a few new tests for special character encoding have been included.


bretto36 (approved):

pascalbaljet:
The removal shaves around 93KB from the build size, around 18KB gzipped.

osbre:
Might also consider letting axios go in favour of [ky](https://github.com/sindresorhus/ky). It has the same features ie real progress tracking, but is built on top of `fetch` to archive minimal bundle size. It'd be another 14kb reduction.

pascalbaljet:
> Might also consider letting axios go in favour of [ky](https://github.com/sindresorhus/ky). It has the same features ie real progress tracking, but is built on top of `fetch` to archive minimal bundle size. It'd be another 14kb reduction.

See #2833

PR #2804, opened 2026-01-09
Fix `<WhenVisible>` re-registering observer when params change

pascalbaljet:
This PR fixes the `<WhenVisible>` component from re-registering its `IntersectionObserver` when the `params` prop changes. When `params` changed, the `useEffect` dependency array would trigger a re-run, causing the observer to re-register. Since the element is already visible, this would immediately fire the observer callback and make an unnecessary request.

### DIFF
--- a/packages/vue3/src/whenVisible.ts
+++ b/packages/vue3/src/whenVisible.ts
@@ -56,7 +56,9 @@ export default defineComponent({
           return
         }
 
-        this.$nextTick(this.registerObserver)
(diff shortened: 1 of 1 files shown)


pascalbaljet:
Draft because the Vue test suite fails in WebKit.

pascalbaljet:
Alright, also a small fix for Vue where it only re-registers the observer when it doesn't exist yet or when data needs to be fetched.

PR #2808, opened 2026-01-11
Fix React Precognition Error Sync (issue #2806)

### LINKED ISSUE #2806: Bug: React useForm `onError` doesn't sync errors to precognition validator
### @inertiajs/react Version

2.3.8

### Backend stack (optional)

_No response_

## Description

When using the `Form` component (or `useForm` with precognition) in React, validation errors from a failed form submission are not synchronized to the precognition validator's internal state. This causes precognition success responses (204) to fail to clear the corresponding field errors.

## Root Cause

In `packages/react/src/useForm.ts`, the `onError` handler in the `submit` function directly sets React state without syncing to the precognition validator:

```typescript
// React adapter - BROKEN
onError: (errors) => {
  if (isMounted.current) {
    setProcessing(false);
    setProgress(null);
    setErrors(errors as FormDataErrors<TForm>); // ← Direct state setter
    setHasErrors(true);
  }
  // ...
};
```

**Line reference**: `packages/react/src/useForm.ts:236-241` 

Compare with the Vue adapter which correctly uses `setError()`:

```typescript
// Vue adapter - CORRECT
onError: (errors) => {
  this.processing = false;
  this.progress = null;
  this.clearErrors().setError(errors as FormDataErrors<TForm>); // ← Uses setError() which syncs
  // ...
};
```

**Line reference**: `packages/vue3/src/useForm.ts:292-295` (master branch)

The `setError` function in React's useForm **does** sync to the validator:

```typescript
const setError = useCallback(
  (fieldOrFields, maybeValue) => {
    setErrors((errors) => {
      const newErrors = { ... }
      setHasErrors(Object.keys(newErrors).length > 0)
      validatorRef.current?.setErrors(newErrors)  // ← Syncs to validator!
      return newErrors
    })
  },
  [setErrors, setHasErrors],
)
```

**Line reference**: `packages/react/src/useForm.ts:167-178` 

## Why This Matters

When the precognition validator receives a success response, it clears errors for the validated fields:

```typescript
// In laravel-precognition validator.ts
onPrecognitionSuccess: (response) => {
  [
    ...setValidated([...validated, ...only]),
    ...setErrors(omit({ ...errors }, only)), // ← Clears 'only' fields from validator's errors
  ].forEach((listener) => listener());
};
```

But since the validator's internal `errors` was never populated (form submission errors went directly to React state), `omit({}, ["name"])` returns `{}` - no change occurs, and the `errorsChanged` event doesn't fire.

When there's a 422 error, it **does** work because:

```typescript
onValidationError: (response) => {
  setErrors(merge(omit({ ...errors }, only), response.data.errors)); // ← Actually sets errors
};
```

This triggers `errorsChanged`, which overwrites the React form errors with the validator's errors.

## Impact

This creates a confusing user experience:

-  Typing an **invalid** value clears other errors (because 422 response syncs state) (Correct)
-  Typing a **valid** value doesn't clear the error (because 204 response can't clear unsynced state) (wrong)

## Comparison with Vue Adapter

| Aspect                              | Vue (Works)                        | React (Broken)   |
| ----------------------------------- | ------------------------------------- | ------------------- |
| `onError` implementation            | `this.clearErrors().setError(errors)` | `setErrors(errors)` |
| Syncs to validator?                 | Yes                                   | No                  |
| Precognition success clears errors? | Yes                                   | No                  |

## Suggested Fix

**Option 1**: Match Vue adapter's implementation

In `packages/react/src/useForm.ts`, change the `onError` handler to use `setError`:

```typescript
onError: (errors) => {
  if (isMounted.current) {
    setProcessing(false);
    setProgress(null);
    clearErrors();
    // Use setError to sync with precognition validator
    setError(errors as FormDataErrors<TForm>);
  }

  if (options.onError) {
    return options.onError(errors);
  }
};
```

**Option 2**: Make the internal `setErrors` also sync to validator

Modify the internal state setter to always sync when validator is present:

```typescript
const [errors, _setErrors] = useState({} as FormDataErrors<TForm>);

const setErrors = useCallback((errors: FormDataErrors<TForm>) => {
  _setErrors(errors);
  setHasErrors(Object.keys(errors).length > 0);
  validatorRef.current?.setErrors(errors);
}, []);
```

Then use `setErrors` consistently throughout.

## Workaround

Until fixed, manually sync errors in your component:

```tsx
const formRef = useRef(null);

<Form
  ref={formRef}
  action="/users"
  method="post"
  onError={(errors) => {
    // Manually sync to precognition validator
    Object.entries(errors).forEach(([key, value]) => {
      formRef.current?.setError(key, value);
    });
  }}
>
  {/* form fields */}
</Form>;
```

## Test Case

- **Success**: `204 No Content` with headers `Precognition: true`, `Precognition-Success: true`
- **Validation Error**: `422 Unprocessable Entity` with headers `Precognition: true` and JSON body `{"errors": {...}}`

## Related Files

- Issue location: `packages/react/src/useForm.ts` (line 236-241)
- Working reference: `packages/vue3/src/useForm.ts` (line 292-295)

### Steps to reproduce

1. Create a form with precognition validation:

```tsx
<Form action="/users" method="post">
  {({ errors, invalid, validate }) => (
    <>
      <input name="name" onChange={() => validate("name")} />
      {invalid("name") && <p>{errors.name}</p>}

      <input name="email" onChange={() => validate("email")} />
      {invalid("email") && <p>{errors.email}</p>}

    </>
  )}
</Form>
```

2. Submit the form with all fields empty → Server returns validation errors for all fields
3. Type a valid value in the "name" field → Triggers `validate("name")`
4. Server returns **204 with Precognition-Success: true** (name is now valid)
5. **Expected**: The "name" error is cleared
6. **Actual**: The "name" error persists

joetifa2003:
## Description

Fixes #2806

This PR fixes a bug in the React adapter where form submission errors (`onError`) were not correctly synchronized with the precognition validator. This caused a state mismatch where the validator remained unaware of defects, preventing subsequent successful precognition requests (204) from correctly clearing the field errors in the UI.

This is working OK for vue and svelte, you can see issue for more details.

## Changes

- Updated `packages/react/src/useForm.ts`: Modified the `onError` handler to use the `setError` helper method instead of the raw `setErrors` state setter. This ensures that errors are propagated to the internal precognition validator.

## Verification

- Added a new reproduction test case `clears submission errors on subsequent precognition success` to `tests/precognition.spec.ts`.
- Verified that the test failed before the fix and passes after the fix.
- Confirmed that Vue and Svelte adapters already exhibited the correct behavior and pass the new test.

### DIFF
--- a/packages/svelte/test-app/Pages/FormHelper/Precognition/ErrorSync.svelte
+++ b/packages/svelte/test-app/Pages/FormHelper/Precognition/ErrorSync.svelte
@@ -1,12 +1,6 @@
 <script lang="ts">
   import { useForm } from '@inertiajs/svelte'
 
-  /**
-   * Test page for GitHub issue #2806:
-   * React useForm `onError` doesn't sync errors to precognition validator
-   *
-   * Uses useForm hook (Form Helper) instead of Form component
-   */
   const form = useForm({
     name: '',
     email: '',
--- a/tests/app/server.js
+++ b/tests/app/server.js
@@ -540,8 +540,6 @@ app.post('/precognition/dynamic-array-inputs', upload.any(), (req, res) => {
   }, 250)
 })
 
-// Route to test that form submission errors are properly synced to precognition validator
-// This reproduces GitHub issue #2806
 app.post('/precognition/error-sync', upload.any(), (req, res) => {
   const isPrecognition = req.headers['precognition'] === 'true'
 
--- a/tests/precognition.spec.ts
+++ b/tests/precognition.spec.ts
@@ -558,31 +558,22 @@ integrations.forEach((integration) => {
     })
 
     test(prefix + 'clears submission errors on subsequent precognition success', async ({ page }) => {
-      /**
-       * Test for GitHub issue #2806:
-       */
-
       await page.goto('/' + integration + '/precognition/error-sync')
 
-      // Step 1: Submit the form with empty fields to trigger onError
+      // Submit with empty fields to trigger validation errors
       await page.click('#submit-btn')
-
-      // Wait for the error response and errors to appear
       await expect(page.locator('#name-error')).toBeVisible()
       await expect(page.locator('#email-error')).toBeVisible()
       await expect(page.locator('#name-error')).toHaveText('The name field is required.')
       await expect(page.locator('#email-error')).toHaveText('The email field is required.')
 
-      // Step 2: Type a valid name and blur to trigger precognition validation
+      // Fill valid name and trigger precognition validation
       await page.fill('input[name="name"]', 'John Doe')
       await page.locator('input[name="name"]').blur()
-
-      // Wait for precognition validation to complete
       await expect(page.locator('#validating')).toBeVisible()
       await expect(page.locator('#validating')).not.toBeVisible()
 
-      // Step 3: Assert that name error is cleared but email error remains
-      // This is where React fails - the name error persists
+      // Name error should be cleared, email error should remain
       await expect(page.locator('#name-error')).not.toBeVisible()
       await expect(page.locator('#email-error')).toBeVisible()
     })


pascalbaljet:
Thanks!

PR #2865, opened 2026-02-04
Add global configuration support for withAllErrors in form components

skryukov:
This PR introduces `form.withAllErrors` in the global config to enable showing all validation errors without manual `.withAllErrors()` calls.

### Context
Introduction of the `withAllErrors: false` logic broke Inertia Rails applications, since we never had middleware cherry-picking the first error from the array. I'd love us to default `withAllErrors` to `true`, but even if we leave the current breaking change in place, we really want this global setting to exist.


pascalbaljet:
Thanks!

PR #2869, opened 2026-02-06
[3.x] Rename `invalid` and `exception` events and add visit callbacks

### LINKED ISSUE #2631: [Question] Why don’t Inertia’s visit APIs (e.g. router.reload) handle Axios/network errors via onError callback?

- [ ] React
- [ ] Vue 3
- [ ] Svelte
- [x] Not Applicable

### JS package version

2.2.6

### Backend stack (optional)

_No response_

### Describe the problem

I had a question when i experienced a tiny problem. When using Inertia’s `router` methods (e.g. `router.reload`), I noticed that **network/axios errors** (e.g. the server is offline, DNS failure, timeout) never trigger the `onError` callback supplied in the options. Instead, the promise simply rejects (or throws) and bypasses the `onError` path entirely.

Because the Inertia internals use `.then()` and not `.catch()` for dispatching callbacks, I can’t intercept those lower-level request errors via `onError`. Also, wrapping calls in `try / catch` is ineffective (since the error occurs asynchronously in the promise chain).

In [here](https://github.com/inertiajs/inertia/blob/2d27a56c80731d13053313583ab838fca4f45388/packages/core/src/request.ts#L79) I see something like:

```js
return axios({
      //....
    })
      .then((response) => {
        this.response = Response.create(this.requestParams, response, this.page)

        return this.response.handle()
      })
      .catch((error) => {
        //....
      })
      .catch((error) => {
          //...
          return Promise.reject(error) // This is the uncaught error that we can't intercept
        }
      })
      .finally(() => {
        //...
      })
```

That means that if the server is down, or there is no response, I can’t gracefully handle or show a custom error for that specific request via the `onError` callback. The only fallback is a global uncaught error which then can't be very straitforward to identify which request failed, it's not really good for ux.

### Questions

1. **Is there a specific reason** Inertia’s internal request handling deliberately bypasses `.catch()` and doesn’t invoke `onError` for network/axios-level failures?

   like we could add something like :
```
this.requestParams.all()['onError'](error);
```
in teh last `catch` in promise chain, instead of just rejecting the promise

2. **Would a callback like `onAxiosError` (or `onNetworkError`) make sense?**

   * That way, only network-level or unexpected failures would hit that callback, while `onError` remains reserved for validation or server-returned errors.
   * If `onAxiosError` is not provided, fall back to the existing `Promise.reject(error)` behavior.

### Why

In my specific scenario, I wanted to simulate a server being offline and present a user-friendly message (e.g. “Server unreachable, try again later”) and add *that particular request* to a retry queue,. But because Inertia swallows the error in the promise chain (or does not route it to `onError`), I have no hook to do that.

Maybe I’m missing some context or design reasoning behind this, but I just wanted to understand if this behavior is intentional or if a change like that could make sense.

### Steps to reproduce

load the page, turn off the server and call : 
```
router.reload({
  onError: () => {
      console.log("error here !");
  },
});
```

pascalbaljet:
This PR renames two events and promotes them to full visit callbacks:

- `invalid` -> `httpException`
- `exception` -> `networkError`

The old names were confusing. The new names clearly describe what they represent: `httpException` fires for non-Inertia responses (e.g. a 500 error), and `networkError` fires when the request fails entirely (e.g. network disconnect).

**Visit callbacks**

Both events are now available as visit callbacks (`onHttpException`, `onNetworkError`), just like the other events. Previously, these were only available as global events.

```js
router.post('/users', data, {
  onHttpException: (response) => {
    // Handle non-Inertia response
  },
  onNetworkError: (error) => {
    // Handle network failure
  },
})
```

Returning `false` from either callback prevents the global event and its default behavior:

```js
router.post('/users', data, {
  onHttpException: (response) => {
    // Custom handling, prevent the error modal from showing
    return false
  },
})
```

Fixes #2631.


pascalbaljet:
Related: https://github.com/inertiajs/inertia-laravel/pull/823

PR #2874, opened 2026-02-10
[3.x] Add a singleton for the usePage function

DavidCCoimbra:
## Summary                                                                                                                                                          

`usePage()` currently creates a new `reactive()` object with 13 `computed()` properties on every call (Vue), and a new `derived()` store on every call (Svelte).

Since all instances read from the same underlying page ref, these extra allocations are redundant.

### The problem

Every call to `usePage()` allocates:
  - **Vue:** 1 `reactive()` + 13 `computed()`
  - **Svelte:** 1 `derived()` store subscription

In a component that calls `usePage()`, rendered in a loop of 30 items:
  - **Before:** 30 reactive proxies + 390 computed, all watching the same ref
  - **After:** 1 reactive + 13 computed, shared across all calls

This adds up in scenarios like virtual scrollers or large loops, where components mount and unmount frequently, each mount creates a new reactive wrapper that gets thrown away on unmount, only to be recreated on the next mount.

This also gives developers more flexibility to call usePage() freely without worrying about performance, whether it's in a loop, a deeply nested child, or called multiple times in the same component. 

From experience, it's common for developers to create wrapper composables around usePage(), and when a page uses 3-4 of those composables, each one calling usePage() internally, the allocations multiply quickly, especially inside loops.

Although it's a small optimization, we never know what codebases are using this :)

### The fix

A singleton solves this. The reactive wrapper is created once on the first call and returned on every subsequent call. This is safe because:
  - All computed properties / derived stores reference the same shared page state
  - The returned object is reactive, it reflects navigation changes automatically
  - No behavior changes for consumers
                                                                                                               
The pageAccessor is just computed getters over that same page.value, so it reads whatever the current request set. No different from before.

You can see the test pages by navigating to /use-page/page1. 

Screenshot 2026-02-10 at 14 00 31

### What's not affected
React's `usePage()` already reads from a Context provider with no allocations, so no change is needed there.
  
  
@pascalbaljet  If this pattern was intentional, happy to hear the reasoning, otherwise this should be a small but useful optimization. Wish you all the best, Inertia is an amazing product keep the good work!

Happy to port this to the 3.x branch as well if this gets merged.

### DIFF
--- a/packages/svelte/test-app/Pages/UsePage/Child.svelte
+++ b/packages/svelte/test-app/Pages/UsePage/Child.svelte
@@ -1,0 +1,17 @@
+<script lang="ts">
+  import { usePage } from '@inertiajs/svelte'
+  import type { Readable } from 'svelte/store'
+
+  export let parentPage: Readable<Record<string, unknown>>
+
+  const page = usePage()
+  // eslint-disable-next-line svelte/require-store-reactive-access
+  const sameAsParent = page === parentPage
+</script>
+
+<div>
+  <p><em>From child component:</em></p>
+  <p data-testid="child-url">{$page.url}</p>
+  <p data-testid="child-component">{$page.component}</p>
+  <p data-testid="child-same-ref">Same instance as parent: <strong>{sameAsParent ? 'yes' : 'no'}</strong></p>
+</div>
--- a/packages/svelte/src/page.ts
+++ b/packages/svelte/src/page.ts
@@ -7,6 +7,7 @@ type SveltePage<TPageProps extends PageProps = PageProps> = Omit<Page<TPageProps
   }
 }
 
+let pageAccessor: Readable<SveltePage> | null = null
 const { set, subscribe } = writable<SveltePage>()
 
 export const setPage = set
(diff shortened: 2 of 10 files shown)


packages/svelte/test-app/Pages/UsePage/Child.svelte
```diff
@@ -0,0 +1,17 @@
+<script lang="ts">
+  import { usePage } from '@inertiajs/svelte'
+  import type { Readable } from 'svelte/store'
+
+  export let parentPage: Readable<Record<string, unknown>>
+
+  const page = usePage()
+  // eslint-disable-next-line svelte/require-store-reactive-access
```
DavidCCoimbra: I disabled this, cause in this case we need to know if the reference is the same.

pascalbaljet:
Thanks for this! I merged it into 3.x because I'm not totally sure of the side effects of this change in existing apps.

DavidCCoimbra:
Awesome, happy to help @pascalbaljet! Let's hope it works smoothly, feel free to ping me if anything comes up. Thanks for the quick review!

Issue #2900, opened 2026-02-19
onFlash does not fire as expected during partial reload when flash exists in response


CTOJoe:
### Inertia adapter(s) affected

- [x] React
- [x] Vue 3
- [x] Svelte
- [ ] Not Applicable

### JS package version

2.3.15

### Describe the problem

The `onFlash` callback does not fire during partial reloads when `flash` exists in the response but the `flash` payload is equal to the previous `flash` payload.

This behavior appears to be caused by this condition in `packages/core/src/response.ts`:
https://github.com/inertiajs/inertia/blob/b2cacf7b09f2909f442a4774192edda298c9afe5/packages/core/src/response.ts#L90-L95

This suppresses `onFlash` during partial reloads when the `flash` payload matches the previous `flash` payload.

However, `flash` exists outside `page.props` and is not included in the `X-Inertia-Partial-Data` protocol. Partial reload filtering applies only to props, not `flash`. Flash is a separate top level field in the Inertia response.

As a result, the server can explicitly return `flash` data in the response, the `flash` data is applied to the page, but `onFlash` does not fire because the payload matches the previous `flash` value and the visit was a partial reload.

### Expected behavior
`onFlash` should fire **whenever** `flash` exists in the response, regardless of partial reload status or whether the `flash` payload matches the previous value.

Flash is response scoped protocol data and is not governed by `X-Inertia-Partial-Data` filtering. Since `flash` exists outside props and cannot be controlled via only or except, its lifecycle should be tied to the response rather than prop equality suppression.

### Proposed fix
Remove the partial reload condition and fire `onFlash` whenever `flash` exists in the response:
```ts
if (Object.keys(flash).length > 0) {
  fireFlashEvent(flash)
  this.requestParams.all().onFlash(flash)
}
```

This aligns `flash` lifecycle with response semantics and ensures consistent behavior regardless of partial reload status.

### Related discussion
This behavior has also been discussed here: https://github.com/inertiajs/inertia/discussions/2783

In that discussion, another developer noted that the `flash` event only fires once unless the `flash` changes and had to add a timestamp to force the event to fire.

### Steps to reproduce

Server:
```php
use Inertia\Inertia;

Route::post('/example', function () {
    Inertia::flash('records_affected', 1);

    return Inertia::render('Example');
});
```

Client:
```ts
router.reload({
  only: ['users'],
  onFlash: (flash) => {
    console.log('flash fired', flash)
  }
})
```

[...]

PR #2902, opened 2026-02-19
[2.x] Normalize flash event behavior on partial requests

CTOJoe:
In current versions of Inertia v2, the `onFlash` callback is only triggered when the `flash` prop differs from the previous response.

This occurs because, during partial requests only, Inertia compares the resolved `flash` prop using an equality check. If the `flash` payload contains identical values between partial requests, Inertia treats it as unchanged and does not trigger `onFlash`.

So:
```php
Inertia.flash({ records_affected: 1 });
```

followed by:
```php
Inertia.flash({ records_affected: 1 });
```

will only trigger `onFlash` once, even though both responses explicitly include flash data.

This behavior is inconsistent with the [documented purpose of flash data](https://inertiajs.com/docs/v2/data-props/flash-data), which is intended to **represent one-time, per-response state**. Flash data should be treated as an event-like payload rather than persistent state, and its presence in the response should be sufficient to trigger `onFlash`, regardless of value equality.

This makes it difficult to reliably use `onFlash `for user feedback such as success messages or action confirmations when the underlying value does not change.

**Please note:**
I was unable to run the test suite locally. However, I updated the relevant tests to reflect what I believe to be the expected behavior introduced by this PR.

Fixes Issue: https://github.com/inertiajs/inertia/issues/2900

### DIFF
--- a/packages/core/src/response.ts
+++ b/packages/core/src/response.ts
@@ -89,7 +89,7 @@ export class Response {
 
     const { flash } = currentPage.get()
 
-    if (Object.keys(flash).length > 0 && (!this.requestParams.isPartial() || !isEqual(flash, previousFlash))) {
+    if (Object.keys(flash).length > 0) {
       fireFlashEvent(flash)
       this.requestParams.all().onFlash(flash)
     }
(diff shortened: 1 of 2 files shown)


pascalbaljet:
I'm closing this one in favor of #2941 as that one handles deferred props. I'll backport it to 2.x.

PR #2926, opened 2026-03-05
[3.x] Fix axios adapter being required

skryukov:
This PR fixes a build error if axios isn't installed. It uses dynamic import since exporting adapter broke `instanceof` usage for adapters.

Related PR: https://github.com/laravel/precognition/pull/142


pascalbaljet:
Thanks again!

PR #2929, opened 2026-03-05
[3.x] Support more createInertiaApp() patterns in SSR transform

skryukov:
The Vite plugin SSR transform now correctly handles more `createInertiaApp()` patterns that were previously ignored:

```js
void createInertiaApp({...})
createInertiaApp({...}).catch(console.error)
void createInertiaApp({...}).catch(console.error)
```

We still strip down everything.


pascalbaljet:
Thanks!

PR #2936, opened 2026-03-08
[3.x] Fix duplicate base path in Vite SSR CSS link URLs

### LINKED ISSUE #2932: [3.x] Vite SSR Dev Server adds duplicate base path
### Inertia version

3.x (beta)

- [ ] React
- [x] Vue 3
- [ ] Svelte
- [ ] Not Applicable

### JS package version

@inertiajs/vite: 3.0.0-beta.1

### Backend stack (optional)

I'm using Symfony with `pentatrion/vite-bundle`. But it shouldn't matter here.

### Describe the problem

Hey, while testing the new Dev Server SSR support in the Inertia 3 Beta, I noticed that CSS URLs are incorrectly constructed when a custom base path is defined in the Vite configuration.

I'm not really experienced with the inner workings of Vite, but i suspect that problem is that the `origin` already contains the base path here:
https://github.com/inertiajs/inertia/blob/76906f33ecf8b9d44b8e1ed4668c6a64c9b815a0/packages/vite/src/css.ts#L39

___

Current link tag output:
```html
<link rel="stylesheet" href="https://localhost:5172/build/build//@fs//Users/<...>/assets/styles/app.css">
```

Expected output:

```html
<link rel="stylesheet" href="https://localhost:5172/build/@fs//Users/<...>/assets/styles/app.css">
```

### Steps to reproduce

```typescript
// vite.config.ts

export default defineConfig({
    plugins: [
        inertia({ ssr: 'assets/ssr.ts' })
    ],
    base: '/build/', // This causes the issue
});
```

isaackaara:
Fixes #2932

`resolveDevServerOrigin()` was returning the full resolved URL from Vite, which already includes the configured base path (e.g. `http://localhost:5172/build`). Then `collectCSSFromModuleGraph()` added the base path again via `basePrefix`, producing doubled paths like `/build/build/` in the generated CSS link tags.

Fixed by using URL parsing to extract just the origin (protocol + host + port) from the resolved URL, so the base path is only added once.

### DIFF
--- a/packages/vite/src/css.ts
+++ b/packages/vite/src/css.ts
@@ -33,7 +33,7 @@ export function collectCSSFromModuleGraph(server: ViteDevServer, entry: string):
 
   const origin = resolveDevServerOrigin(server)
   const base = server.config.base || '/'
-  const basePrefix = base === '/' ? '' : base
+  const basePrefix = base === '/' ? '' : base.replace(/\/$/, '')
 
   return cssModules.map(({ url, id }) => {
     const href = `${origin}${basePrefix}${url}`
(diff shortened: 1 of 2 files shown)


pascalbaljet:
Thanks!

PR #2937, opened 2026-03-08
[3.x] Stop collecting CSS sub-dependencies during SSR

### LINKED ISSUE #2935: SSR CSS collection injects duplicate sub-dependency stylesheets, overriding Tailwind @theme customizations
### Inertia version

3.x (beta)

- [x] React
- [ ] Vue 3
- [ ] Svelte
- [ ] Not Applicable

### JS package version

3.0.0-beta.2

### Backend stack (optional)

Laravel 12
PHP 8.4
inertiajs/inertia-laravel v2
Tailwind CSS 4.2.1
@tailwindcss/vite 4.2.1
Vite 7.3.1

### Describe the problem

_Hey all! I've created a branch for my app to try out Inertia.js v3, so far it's been great! However I noticed a weird CSS issue. I threw this to Claude and it said this was an Inertia bug, so thought I'd throw its output here! Hope it helps_

---

During SSR in dev mode, `collectCSSFromModuleGraph` in `@inertiajs/vite` walks the SSR module graph and injects `<link>` tags for every CSS module it finds — including transitive sub-dependencies of CSS files (i.e. CSS files imported via `@import` inside other CSS files).

This is problematic because CSS-processing Vite plugins like `@tailwindcss/vite` already compile `@import` sub-dependencies into the parent stylesheet. Injecting them again as separate `<link>` tags causes duplicate, conflicting CSS.

For example, when `app.css` contains `@import 'tailwindcss'` and a @theme block with custom overrides, `@tailwindcss/vite` compiles both together — the output of `app.css` contains the correct custom values. But the SSR CSS collector also finds `tailwindcss/index.css` in the module graph and injects it as a separate `<link>` tag. This separate stylesheet contains the default Tailwind theme values and loads after `app.css,` overriding the custom @theme values (e.g. custom `--font-sans` is lost).

**Expected:** Only root CSS files (directly imported by JS modules) should be injected as <link> tags. CSS @import sub-dependencies are already compiled into their parent by their respective Vite plugins.

**Actual:** All transitive CSS dependencies are injected, causing duplicate stylesheets with conflicting values.

### Steps to reproduce

1. Create a standard Inertia v3 React app with SSR enabled via the `@inertiajs/vite` plugin
2. In `resources/css/app.css`:
```css
@import 'tailwindcss';


@theme {
  --font-sans: 'MyCustomFont', ui-sans-serif, system-ui, sans-serif;
}
```
3. In `resources/js/app.tsx`:
```tsx
import '../css/app.css';
import { createInertiaApp } from '@inertiajs/react';
import { hydrateRoot } from 'react-dom/client';

createInertiaApp({
  resolve: name => {
    const pages = import.meta.glob('./pages/**/*.tsx');
    return pages[`./pages/${name}.tsx`]();
  },
  setup({ el, App, props }) {
    hydrateRoot(el, <App {...props} />);
  },
});
```
4. Run `bun run dev` (Vite dev server)
5. Load any page and inspect the HTML <head> — you'll see both:
```html
<link rel="stylesheet" href="...app.css" data-vite-dev-id="...">
<link rel="stylesheet" href="...tailwindcss/index.css">
```
6. Check the computed value:
```js
// Returns the Tailwind DEFAULT, not 'MyCustomFont'
getComputedStyle(document.documentElement).getPropertyValue('--font-sans')
```

**Workaround:** Split `@import 'tailwindcss'` into individual imports and pre-import `tailwindcss/theme.css` from JavaScript before app.css, so it's collected first in the depth-first module graph traversal and app.css loads after it (winning the cascade).

pascalbaljet:
During SSR in dev mode, the CSS collector walks the module graph and injects `<link>` tags for every CSS module it encounters, including CSS files imported via `@import` inside other CSS files. Vite plugins like `@tailwindcss/vite` already compile these sub-dependencies into the parent stylesheet.

The fix stops traversing into a CSS module's own imports, since only root CSS files directly imported by JS modules need their own `<link>` tags.

Fixes #2935.

### DIFF
--- a/packages/react/test-app/Pages/SSR/Page1.tsx
+++ b/packages/react/test-app/Pages/SSR/Page1.tsx
@@ -1,26 +1,32 @@
-import { Link } from '@inertiajs/react'
+import { Link, usePage } from '@inertiajs/react'
 
-export default ({ user, items, count }: { user: { name: string; email: string }; items: string[]; count: number }) => (
-  <div>
-    <h1 data-testid="ssr-title">SSR Page 1</h1>
+export default ({ user, items, count }: { user: { name: string; email: string }; items: string[]; count: number }) => {
+  const page = usePage()
 
-    <div data-testid="user-info">
-      <p data-testid="user-name">Name: {user.name}</p>
-      <p data-testid="user-email">Email: {user.email}</p>
+  return (
+    <div>
+      <h1 data-testid="ssr-title">SSR Page 1</h1>
+
+      <p data-testid="page-url">URL: {page.url}</p>
+
+      <div data-testid="user-info">
+        <p data-testid="user-name">Name: {user.name}</p>
+        <p data-testid="user-email">Email: {user.email}</p>
+      </div>
+
+      <ul data-testid="items-list">
+        {items.map((item) => (
+          <li key={item} data-testid="item">
+            {item}
+          </li>
+        ))}
+      </ul>
+
+      <p data-testid="count">Count: {count}</p>
+
+      <Link href="/ssr/page2" data-testid="navigate-link">
+        Navigate to another page
+      </Link>
     </div>
-
-    <ul data-testid="items-list">
-      {items.map((item) => (
-        <li key={item} data-testid="item">
-          {item}
-        </li>
-      ))}
-    </ul>
-
-    <p data-testid="count">Count: {count}</p>
-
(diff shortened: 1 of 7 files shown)


entity:
Great job! This should link to my initial issue on this, https://github.com/inertiajs/inertia/issues/2935

PR #2940, opened 2026-03-09
[3.x] fix: remove request from stream on network failure

Anthony14FR:
When `request.send()` rejects (network failure), the request was never removed from the `requests` array, causing a memory leak, `hasPendingOptimistic()` returning stale results, and `interruptInFlight()` targeting ghost requests.

**Fix:** pass the cleanup callback as both fulfillment and rejection handler.

```ts
// before
request.send().then(() => {
  this.requests = this.requests.filter((r) => r !== request)
})

// after
const remove = () => { this.requests = this.requests.filter((r) => r !== request) }
request.send().then(remove, remove)
```

Includes a Playwright test that aborts Inertia XHR requests via `page.route()` and verifies subsequent requests still complete normally across all three adapters.

### DIFF
--- a/packages/react/test-app/Pages/NetworkError.tsx
+++ b/packages/react/test-app/Pages/NetworkError.tsx
@@ -1,7 +1,7 @@
 import { router } from '@inertiajs/react'
 import { useState } from 'react'
 
-export default ({ status }: { status?: string }) => {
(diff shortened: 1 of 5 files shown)


pascalbaljet:
Thanks!

Issue #2930, opened 2026-03-05
Shared props not typed properly


peaklabs-dev:
### Inertia version

3.x (beta)

### @inertiajs/svelte Version

2.3.1 and 3.0.0-beta.2

### Describe the problem

When using `page` to access a typed shared prop, it is `any` instead of the type specified.

```svelte
<script lang="ts">
	import { page } from "@inertiajs/svelte";
</script>

<svelte:head>
	<title>Dashboard - {page.props.appName}</title> <!-- type is any instead of string -->
</svelte:head>
```

### Steps to reproduce

Add this `global.d.ts` file:
```ts
declare module "@inertiajs/core" {
	export interface InertiaConfig {
		sharedPageProps: {
			appName: string;
		};
		errorValueType: string[];
	}
}
```

The `tsconfig.json` checks all `.ts` file via `"include": ["resources/js/**/*.ts", "resources/js/**/*.svelte"]`, explicitly including `.d.ts` files did not fix the issue.

I am using bun as the package manager and installing `@inertiajs/core` did also not fix the issue.

PR #2945, opened 2026-03-10
[v3.x] fix(types): module augmentation example

peaklabs-dev:
## Changes

- See more https://github.com/inertiajs/docs/pull/45

## Issues

- fixes: https://github.com/inertiajs/inertia/issues/2930


pascalbaljet:
Thanks!

PR #2946, opened 2026-03-10
Use SharedPageProps in GlobalEventsMap event types

hamedelasma:
The `beforeUpdate`, `navigate`, and `success` events in `GlobalEventsMap` currently use bare `Page` instead of `Page<SharedPageProps>`. This means that when users augment `SharedPageProps` via declaration merging, their custom types are not reflected in global event callbacks.

**Before:**
```ts
router.on('success', (event) => {
  event.detail.page.props.mySharedProp // ❌ Type error - not recognized
})
```

**After:**
```ts
router.on('success', (event) => {
  event.detail.page.props.mySharedProp // ✅ Correctly typed
})
```

This also applies to `beforeUpdate` and `navigate` events.


pascalbaljet:
Thanks!

PR #2952, opened 2026-03-12
[3.x] fix(type): add Record<string, string> to UrlMethodPair type

stafast:
Working on a project with `laravel/wayfinder-next`. Tried out v3-beta and got TypeScript errors for `:action="store()"` and f`orm.submit(store())`. Only `submit(store().method, store().url)`  was okay.

Reason is that `Record<string, string>` was added to `RouteDefinition` in wayfinder and miss now in [3.x].
[laravel/wayfinder#180](https://github.com/laravel/wayfinder/pull/180)


pascalbaljet:
Thanks!

Issue #2966, opened 2026-03-16
[3.x] useHttp hook attempting to parse JSON from HTTP 204 response


jameshine:
### Inertia version

3.x (beta)

### Inertia adapter(s) affected

- [ ] React
- [ ] Vue 3
- [ ] Svelte
- [ ] Not Applicable

### JS package version

3.0.0-beta.3

### Backend stack (optional)

Laravel 12.x
PHP 8.4

### Describe the problem

Using the new `useHttp` hook to post to an endpoint with a 204 no content response. The response contains no body but the hook is trying to parse the json response.

### Steps to reproduce

1. Create a route that returns no content
`Route::post('/foo', fn () => response()->noContent());`

2. Create request
```
import { useHttp } from '@inertiajs/react';

export default function Welcome() {
    const { post } = useHttp()

    const doSomething = () => {
        post('/foo')
    }

    return <div>
        <button onClick={() => doSomething()}>Click</button>
    </div>
}
```

3. See console - Uncaught (in promise) SyntaxError: Unexpected end of JSON input

<img width="485" height="214" alt="Image" src="https://github.com/user-attachments/assets/8de09725-15f5-4208-bde7-eab600a56e94" />

entity:
Should be fixed in https://github.com/inertiajs/inertia/pull/2967 :tada:

pascalbaljet:
Thanks for reporting!

PR #2967, opened 2026-03-16
[3.x] Prevent useHttp from parsing empty 204 responses

entity:
Added some extra handling to the `response.data` call to only parse when the data exists, prevents a syntax error from 204 no content responses.

Fixes https://github.com/inertiajs/inertia/issues/2966


pascalbaljet:
Thanks!

Issue #2963, opened 2026-03-14
[3.x] setLayoutProps() causes hydration mismatches with SSR - store leaks between requests and is never reset


entity:
### Inertia version

3.x (beta)

### Inertia adapter(s) affected

- [x] React
- [ ] Vue 3
- [ ] Svelte
- [ ] Not Applicable

### JS package version

3.0.0-beta.3

### Backend stack (optional)

Laravel 12
PHP 8.4
inertiajs/inertia-laravel 3.0.0-beta
Vite 8.0.0
React 19

### Describe the problem

_Claude Code helped me track this down and write this up._

`setLayoutProps()` causes hydration mismatches on every page load when SSR is enabled. Two separate issues going on here.

**1. The layout props store leaks between SSR requests**

The store is a module-level singleton (`var store = createLayoutPropsStore()`, index.js:23). `resetLayoutProps()` only runs during client-side navigation (in a `useEffect`, index.js:119) — never between SSR renders.

So if I visit `/` (which sets a title via `setLayoutProps`), then open `/about` in a new tab (which doesn't call `setLayoutProps` at all), the `/about` SSR HTML still has `/`'s title baked in because the store was never cleared.

On the client, the store starts fresh. `useLayoutProps()` returns defaults. Server HTML has stale data. React throws.

Console on `/about` — a page that never sets any layout props:

```diff
  <AppLayout>
    <div className="min-h-scre...">
      <nav>
+     <main className="mx-auto max-w-4xl px-4 py-8">
-     <header className="bg-blue-600 py-8">
```

That `<header>` belongs to the home page. It leaked through the store.

**2. React renders the layout before the page**

Even if the store were properly reset, the render order works against you. React goes top-down: the layout calls `useLayoutProps()` (reads empty store, gets defaults) before the page calls `setLayoutProps()` (writes to store, too late). The store notifies listeners via `queueMicrotask()` (core/index.js:4232) so a re-render happens after the initial pass — but during hydration, React already compared the initial pass to the server HTML and found the mismatch.

**Workaround**

Static props via tuples skip the store entirely and work fine:

```tsx
Page.layout = [AppLayout, { title: 'Dashboard' }]
```

For dynamic values, wrap the layout directly and opt out of the default:

```tsx
function Page() {
  return <AppLayout title="Something dynamic">...</AppLayout>
}
Page.layout = null
```

### Steps to reproduce

Reproduction repo: https://github.com/entity/inertia-v3-sandbox

1. Clone and install
[...]

PR #2970, opened 2026-03-17
[3.x] Fix layout props store leaking between SSR requests

pascalbaljet:
The layout props store is a module-level singleton that persists across SSR requests. When page A sets layout props and page B doesn't, the server-rendered HTML for page B still contains page A's values. On the client the store starts fresh, causing hydration mismatches.

This resets the layout props store at the start of each SSR render in all three adapters and synchronizes the framework-specific reactive state so the reset is immediately reflected during rendering.

See #2963.


pascalbaljet:
Closing this one in favor of #2972.

PR #2972, opened 2026-03-19
[3.x] Simplify layout props by passing them directly as component prop

### LINKED ISSUE #2963: [3.x] setLayoutProps() causes hydration mismatches with SSR - store leaks between requests and is never reset

pascalbaljet:
This PR simplifies the layout props architecture by removing the `useLayoutProps()` hook. This is a breaking change from earlier 3.x betas, but acceptable since 3.x is still in beta.

Instead of layouts consuming props through a hook with a defaults object, layout props from the page's layout tuple and from `setLayoutProps()` / `setLayoutPropsFor()` are now merged and passed directly as component props to the layout. Layouts just declare their props using standard framework conventions (`withDefaults(defineProps())` in Vue, function parameters with defaults in React, `$props()` in Svelte).

The layout definition on page components now supports passing static props via tuples:

```js
// Single layout with props
layout: [AppLayout, { title: 'My Page' }]

// Nested layouts with props
layout: [
  [AppLayout, { title: 'My Page' }],
  [ContentLayout, { padding: 'lg' }],
]

// Named layouts with props
layout: {
  app: [AppLayout, { showSidebar: true }],
  content: [ContentLayout, { padding: 'lg' }],
}
```

For cases where layout props need to be derived from page props, `layout` also accepts a callback:

```js
layout: (props) => [AppLayout, { title: 'Profile: ' + props.user.name }]
```

This callback is evaluated before rendering, so it works during SSR (unlike `setLayoutProps()`, which only runs on the client).

These static and callback props are merged with any dynamic props from `setLayoutProps()` / `setLayoutPropsFor()` and passed directly to the layout component. This removes a layer of indirection and the `mergeLayoutProps()` utility from core. The merging now happens in each adapter's `App` component at render time, which also fixes SSR issues where layout prop state could leak between requests.

New SSR tests verify that layout props don't leak across requests, that hydration works correctly with both tuple props and the callback form, and that callback-derived values appear in the server-rendered HTML.

Fixes #2963


entity:
This is great! Cannot wait to try this :)

SantosVilanculos:
I think it would be great to have something like this:

```jsx
import { createPage } from '@inertia/react';
import Layout from './Layout';

export const Page = createPage({
    layout: [Layout, { title: 'Title' }],
    component: () => <></>
});
```
or:

```jsx
import { createPage, setLayoutProps } from '@inertia/react';
import Layout from './Layout';

function Component() {
    setLayoutProps({
        title: 'Dashboard',
        showSidebar: false
    });

    return <>{/* ... */}</>;
}

export const Page = createPage({
    layout: [Layout, { title: 'Title' }],
    component: Component
    // contributors can add more useful things hire
});
```

with autocomplete and the ability to add other parameters to the 'createPage' method.

TimKunze96:
~~Let's say I have some compute heavy properties that are derived from page props and are needed in the page component and the layout. Without the useLayoutProps hook I am forced to do this computation twice, since I can't directly pass the result from inside the page component. Am I missing something here?~~

I did indeed miss the new setLayoutProps hook

TimKunze96:
~~Neither I am able to react to user i/o anymore~~

PR #2982, opened 2026-03-22
[3.x] Allow history encryption to be unsupported in non-SSL environments

jlswanson28694:
### Problem
I'm using the `EncryptHistory` middleware on my Laravel backend, but my dev environment isn't using SSL. Unfortunately, Inertia throws an error under these conditions, and I am unable to use my app.

### Proposed Solution
Check if encryption is supported at the encryption entrypoint `encryptHistory` function, and log the warning once to let users know history encryption is not supported without SSL, but then continue without encrypting instead of throwing an error.

An alternative solution would be adjusting the server-side adapters to allow for a global history encryption kill switch (one that also turns off the Laravel middleware for example), but the approach I took here made the most sense to me.

This PR also updates a few parameter types to increase type safety.

### DIFF
--- a/packages/core/src/encryption.ts
+++ b/packages/core/src/encryption.ts
@@ -23,7 +23,7 @@ export const historySessionStorageKeys = {
   iv: 'historyIv',
 }
 
-export const decryptHistory = async (data: any): Promise<any> => {
+export const decryptHistory = async (data: BufferSource): Promise<any> => {
   const iv = getIv()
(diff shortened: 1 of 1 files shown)


pascalbaljet:
Thanks for the PR! I'd rather keep this as an explicit error rather than silently degrade. Local SSL is easy to set up with [Laravel Herd](https://herd.laravel.com).

PR #3004, opened 2026-03-25
[3.x] Resolve pages at @

skryukov:
This PR makes the default resolver to also search `@/pages` and `@/Pages` in addition to `./pages` and `./Pages`.

This makes zero-config page resolution work for setups where pages live under the `@` alias, and not just next to inertia entry point.


SuperDJ:
Why not make it a configurable prefix?

skryukov:
Hmm, if you have to configure something, you can already use `pages: '~/pages'`, so we already have that. Or am I missing something?

pascalbaljet:
Thanks for the PR! I'm closing this one for now as implicitly expanding the default resolver feels risky. Since `pages: '@/pages'` already works as an explicit opt-in, I'd rather keep the default narrow and have users configure it manually.

PR #3007, opened 2026-03-25
[3.x] Warm up SSR module graph on dev server start

pascalbaljet:
When the Vite dev server starts, the first SSR request pays the full module transformation cost. This eagerly loads the SSR entry module as soon as the server starts listening, so the module graph is already warm by the time the first real request comes in.


entity:
I definitely feel this cold start on deploy, this will make a huge difference 🙌

PR #3019, opened 2026-03-30
fix: prevent explicit undefined values from overwriting visit defaults (#3018)

### LINKED ISSUE #3018: `Uncaught TypeError: Cannot read properties of undefined (reading 'length')` when visit options contain explicit `undefined` values

- [x] React
- [x] Vue 3
- [x] Svelte
- [ ] Not Applicable

### JS package version

2.3.18

### Backend stack (optional)

_(Not applicable, the bug is in `@inertiajs/core`)_

### Describe the problem

In `getPendingVisit` https://github.com/inertiajs/inertia/blob/b58c5bf6b4a2aceb0e5567140ad6105b80d0ab7b/packages/core/src/router.ts#L621
default values are correctly defined in the `mergedOptions` object (typed as `Visit`). 

However, `options` parameters is typed as `VisitOptions` (which is `Partial<Visit & VisitCallbacks>`), which means TypeScript allows properties to be explicitly set to `undefined`. E.g. when an `options` object like `{ only: undefined }` is spread into `mergedOptions`, the explicit `undefined` value **overwrites** the default `[]`, producing an object whose runtime value contradicts its declared type (`Visit`), without `exactOptionalPropertyTypes` TypeScript does not flag this.

Later in the `visit` method, `visit.only.length`, `visit.except.length`, and `visit.reset.length` are accessed directly (introduced by PR #2843) without any null/undefined guard (which is fine because the type tell that these properties should be defined). This causes a runtime crash:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
```

It happens in practice when options are built dynamically (e.g. conditionally spreading partial option objects, or when a wrapper function passes through options that may be `undefined`). This bit me in production while migrating from 1.x to 2.x, which is how I ended up here 😅

Expected behavior: `router.visit('/url', { only: undefined })` should behave the same as `router.visit('/url', {})`. The default value (`[]`) should not be overwritten by an explicit `undefined`. Otherwise, types should not allow this.

Possibles solutions:
- Strip `undefined` values from `options` and `configuredOptions` before merging. This is the most pragmatic and simple fix.
- Enable `exactOptionalPropertyTypes` in `tsconfig.json` to make TypeScript reject `{ only: undefined }` at the type level. This is the "correct" TS solution but has a much larger blast radius (tried it, a lot of errors comes up, unrelated to this, although this does indeed involve other potential risks).

### Steps to reproduce

```ts
import { router } from '@inertiajs/react' // or vue, or svelte

router.visit('/some-url', { only: undefined })
// => Uncaught TypeError: Cannot read properties of undefined (reading 'length')
```

The issue is reproducible on the latest `main` branch (v3).

I'm willing to open a PR if solution 1 or 2 is fine.

Vitalcheffe:
## Problem

When visit options contain explicit `undefined` values (e.g. `{ only: undefined }`), the spread operator overwrites the default `[]` with `undefined`, causing a runtime crash:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
```

This happens when options are built dynamically with conditional spreads or wrapper functions that pass through undefined. TypeScript's `Partial<Visit>` allows this without error.

## Reproduction

```ts
router.visit('/some-url', { only: undefined })
// => TypeError: Cannot read properties of undefined (reading 'length')
```

## Fix

Added a `stripUndefined` helper that removes keys with `undefined` values from `options` and `configuredOptions` before they're spread into `mergedOptions`. This ensures defaults (`only: []`, `except: []`, `reset: []`) are preserved.

This is a minimal, surgical fix — 12 lines added, no behavior changes for correctly-formed options.

Fixes #3018

### DIFF
--- a/packages/core/src/objectUtils.ts
+++ b/packages/core/src/objectUtils.ts
@@ -1,3 +1,15 @@
+export const stripTopLevelUndefined = <T extends Record<string, unknown>>(obj: T): T => {
+  const result = {} as T
+
+  for (const key of Object.keys(obj) as (keyof T)[]) {
+    if (obj[key] !== undefined) {
+      result[key] = obj[key]
+    }
+  }
+
+  return result
+}
+
 export const objectsAreEqual = <T extends Record<string, any>>(
   obj1: T,
   obj2: T,
--- a/packages/core/src/router.ts
+++ b/packages/core/src/router.ts
@@ -631,6 +631,19 @@ export class Router {
       ? defaultVisitOptionsCallback(href.toString(), cloneDeep(options)) || {}
       : {}
 
+    // Strip explicit undefined values so they don't overwrite defaults during spread.
+    // TypeScript's Partial<> allows { only: undefined } which would overwrite the
+    // default [] and later crash on .length access (see #3018).
+    const stripUndefined = <T extends Record<string, unknown>>(obj: T): Partial<T> => {
+      const result: Partial<T> = {}
+      for (const key of Object.keys(obj) as (keyof T)[]) {
+        if (obj[key] !== undefined) {
+          result[key] = obj[key]
+        }
+      }
+      return result
+    }
+
     const mergedOptions: Visit = {
       method: 'get',
       data: {},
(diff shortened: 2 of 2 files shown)


pascalbaljet:
Thanks!

PR #3024, opened 2026-03-31
[3.x] Fix flash data re-firing on partial reloads

### LINKED ISSUE #3015: Flash data re-fires on partial reloads (v3)
### Version

- `@inertiajs/vue3`: 3.0.0
- `@inertiajs/core`: 3.0.0
- `inertiajs/inertia-laravel`: 3.0.1

### Describe the problem

Flash data set via `Inertia::flash()` re-fires on subsequent partial reloads, even though the docs state: *"After the flash data is sent to the client, it is cleared and will not appear in subsequent requests."*

The flash data is correctly consumed on the server side (session flash is aged), but the client retains `page.flash` after the initial event fires. On a partial reload, the server response contains no `flash` key, yet the retained flash data appears to re-trigger the `flash` event.

### Steps to reproduce

1. Perform an action that flashes a toast via `Inertia::flash()` and redirects back (e.g., adding a member to a section)
2. The toast shows and self-dismisses — this is correct
3. Trigger a partial reload on the same page: `router.reload({ only: ['sectionMembers'] })`
4. The same toast notification fires again

This happens consistently across multiple partial reload sites in our app. It does **not** happen in production on Inertia v2 (where `router.on('flash')` didn't exist).

### Expected behavior

Flash data should fire once when initially received from the server, then not re-fire on subsequent partial reloads to the same page.

### Workaround

Clearing the consumed flash key on the client after showing the notification:

```js
router.on('flash', (event) => {
    const notification = event?.detail?.flash?.toast;
    if (!notification?.message) return;

    toast[notification.type](notification.message);

    // Clear consumed toast so partial reloads don't re-fire it
    router.flash(({ toast: _, ...rest }) => rest);
});
```

### Notes

This is related to previously reported issues #1184 and #1643, which were closed. The v3 flash system (`Inertia::flash()` / `page.flash`) was intended to address flash persistence in history state, but the partial reload case appears to remain unresolved.

pascalbaljet:
During partial reloads, flash data from the current page was always merged into the response. This meant that if the server didn't send flash on a partial reload (the normal Laravel behavior, since session flash is aged after the first request), the old flash data survived and the flash event fired again.

This changes the merge logic so that existing flash data is only preserved for deferred props requests (background fetches that replay the same flash from the initial load). Regular partial reloads now use whatever the server sent, which correctly clears stale flash when the server omits it.

Fixes #3015.

### DIFF
--- a/packages/core/src/response.ts
+++ b/packages/core/src/response.ts
@@ -426,10 +426,11 @@ export class Response {
       }
     }
 
-    // Preserve flash data and merge with new flash data on non-deferred requests
-    pageResponse.flash = {
-      ...currentPage.get().flash,
-      ...(this.requestParams.isDeferredPropsRequest() ? {} : pageResponse.flash),
+    // Preserve flash data on deferred props requests (background fetches that replay
+    // the same flash from the initial load), but let regular partial reloads use
+    // whatever the server sent (which may be empty, clearing stale flash)
+    if (this.requestParams.isDeferredPropsRequest()) {
+      pageResponse.flash = { ...currentPage.get().flash }
     }
 
     const currentOriginalDeferred = currentPage.get().initialDeferredProps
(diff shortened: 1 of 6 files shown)


francoism90:
Just to confirm: it indeed seem to have fixed the multiple flash issues. :)

pascalbaljet:
> Just to confirm: it indeed seem to have fixed the multiple flash issues. :)

Thanks for reporting back!

PR #3028, opened 2026-04-01
Update axios version

SuperDJ:
Update axios version to match https://github.com/laravel/laravel/compare/12.x...13.x and fix https://safedep.io/axios-npm-supply-chain-compromise/

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,14 +35,14 @@
     "format": "prettier --write ."
   },
   "dependencies": {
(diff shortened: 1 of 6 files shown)


pascalbaljet:
The corrupted package has already been removed from npm. Axios has been removed from the `laravel/laravel` repo, as well as from the Starter Kits.

Also, for developing Inertia we're using `pnpm`, which blocks scripts by default.

PR #3038, opened 2026-04-05
[3.x] Fix flash event not firing when response contains errors

skryukov:
Flash events (`fireFlashEvent`, `onFlash`) were silently dropped when the response also contained validation errors, because `Response.process()` returned early in the error path before reaching the flash event code.

This PR moved flash event firing to happen right after `setPage()`, before the error check, so flash data is delivered regardless of whether errors are present.


pascalbaljet:
Thanks!

PR #3041, opened 2026-04-06
[3.x] Fire flash event after full reload

skryukov:
This PR fixes Inertia not firing flash event after 409 full page reloads.

Draft since built on top of #3038


pascalbaljet:
Thanks!

PR #3057, opened 2026-04-10
[3.0] Fix React `useForm` submitting stale data after synchronous `setData`

skryukov:
In the React adapter, calling `setData()` (or `reset()`) and then immediately submitting the form in the same event handler submits the pre-update data, not the value just written.

```tsx
const { data, setData, post } = useForm({ code: '' })

onChange={(value) => {
  setData('code', value)
  if (value.length === 6) {
    post('/verify-otp') // submits the *previous* value ❌
  }
}}
```

Root cause: `submit()` reads form data from `dataRef.current`, but `dataRef` is only refreshed inside a post-commit `useEffect`. Vue and Svelte are unaffected.

See https://github.com/inertiajs/inertia-rails/issues/300


pascalbaljet:
Thanks!

PR #3069, opened 2026-04-28
fix(core): support nonce for progress styles

### LINKED ISSUE #3060: Progress component requires 'unsafe-inline' content security progress directive

- [x] React
- [x] Vue 3
- [x] Svelte
- [ ] Not Applicable

### JS package version

2.3.13

### Backend stack (optional)

Any server setting style CSP headers **without** "unsafe-inline".

### Describe the problem

When an Inertia app is created with createInertiaApp(), it defaults to including the Progress Component with CSS inserted.

This function ([progresscomponent.ts:286](https://github.com/inertiajs/inertia/blame/fa241cad5a803724e9beacd09a0d8d43978f209c/packages/core/src/progress-component.ts#L286)) then inserts CSS as a style tag at the end of the document head.  This will fail content security policy checking unless 'unsafe-inline' is specified, because no nonce is specified.  

**Temporary workarounds**
- Use unsafe-inline for CSP styles
- Configure the progress bar with includeCSS first

**Suggested solution**
In the `injectCss()` function above, check if the page includes a meta tag with a CSP nonce.  If so, add it as an attribute to the style element.  Something like:

```js
const injectCSS = (color: string): void => {
    const element = document.createElement('style')

    // ...

    const metaNonce = document
        .querySelector<HTMLMetaElement>('meta[name="csp-nonce"]')
        ?.content;

    element.setAttribute('nonce', metaNonce)

    // ....

    document.head.appendChild(element)

}
```

This should offer a solution to people with a strict CSP policy.  I am happy to open a pull request for this, but it would be my first one so thought would check whether this would be welcome first.  

### Steps to reproduce

1. Create a new inertia app with default settings
2. Set a content security policy header without unsafe inline for styles `style-src 'self'`
3. Observe the CSP error introduced

artemxknpv:
Fixes #3060.

Adds a `progress.nonce` option and applies it to the injected progress `<style>` tag when Inertia's built-in progress CSS is enabled.

This lets apps keep a CSP without `style-src 'unsafe-inline'` while still using the default progress styles.

Tested:

- `corepack pnpm --filter @inertiajs/core test`
- `corepack pnpm --filter @inertiajs/core build`
- `corepack pnpm exec prettier --check packages/core/src/progress-component.ts packages/core/src/progress.ts packages/core/src/types.ts packages/core/tests/progress-component.test.ts`
- `git diff --check`


pascalbaljet:
Thanks for this! I've moved the test to our existing Playwright test suite.

PR #3093, opened 2026-05-07
[3.x] Add `mode` option to `router.poll`

pascalbaljet:
This PR adds a `mode` option to `router.poll()` and `usePoll()` that controls how concurrent requests are handled. The default behavior is unchanged, but when responses take longer than the configured interval, requests can stack up on the server.

The new option accepts `overlap` (default, current behavior), `cancel` (abort the in-flight request before starting the next), or `rest` (wait the configured interval after each response, so requests never overlap).

```js
router.poll(2000, {}, { mode: 'cancel' })
router.poll(2000, {}, { mode: 'rest' })
```


neluba:
@pascalbaljet  I think another good option would be to queue in the interval. So that the timer only starts after the last request is done. That way the the duration between an end of a request and the start of a request would always be the same.

It wouldn't really be an interval though. Rather a timeout after the previous request finished.

nicolagianelli:
> @pascalbaljet I think another good option would be to queue in the interval. So that the timer only starts after the last request is done. That way the the duration between an end of a request and the start of a request would always be the same.
> 
> It wouldn't really be an interval though. Rather a timeout after the previous request finished.

This would be useful, maybe with overlap: 'queue' ?

pascalbaljet:
Thanks for the feedback! I dropped the `timeout` option entirely for now. It's a bigger feature than this PR needs to be, and I'd rather ship it on its own.

For polls, I pivoted to a single `mode` option with three values:

- `overlap` (default): current behavior, parallel requests allowed
- `cancel`: abort the in-flight request before starting the next
- `rest`: wait the configured interval after each response completes, so requests never overlap

The `rest` mode makes the interval become the gap between requests rather than the time between starts. Naming borrowed from Laravel's `--rest` queue worker flag.

PR #3105, opened 2026-05-18
[3.x] Add Inertia.once for events that fire once

sebastiandedeyne:
We came across something where we wanted to react to an Inertia event once instead of indefinitely.

This can be done by unsubscribing from the event in the callback.

```ts
const remove = router.on('before', () => {
    remove();
    // Do something…
});
```

Other libraries (Node.js, Vue, the DOM) have first-class support for this behavior. This PR adds a `once` method to the router to fire a callback one time.

```ts
router.once('before', () => {
    // Do something…
});
```

## Alternative API

There are two possible public API's here, and both are used in the wild.

```ts
// Node.js-style (distinct method)
router.once('before', () => { /* … */ });

// DOM-style (options object)
router.on('before', () => { /* … */ }, { once: true });
```

I personally find the Node.js-style more succinct, readable, and easily discoverable so I went with that one. Although if there's interest in this feature and you prefer DOM-style, I can update this PR.


pascalbaljet:
Thanks! I also find the Node-style more readable.

PR #3121, opened 2026-05-21
[3.x] Support network URLs for loading CSS assets in SSR dev server

simonellensohn:
When running Vite inside a container or behind a reverse proxy (e.g. Caddy) with a custom `server.host`, Vite places the dev server URL in `resolvedUrls.network` instead of `resolvedUrls.local`.

Without this fix I was having issues on my current project that any SSR loaded CSS files were served over `https://localhost:5173` instead of `https://dev.example.com:5173`, leading to SSL issues.


pascalbaljet:
Thanks!

PR #3122, opened 2026-05-21
[3.x] Await user-provided Form onSuccess callback before completing submission

mattwigham:
The `<Form>` component wraps the user-provided `onSuccess` callback in a synchronous closure that fires `onSubmitComplete` immediately after the user's callback returns. `onSubmitComplete` flips `processing` back to `false`, which removes `disableWhileProcessing`'s `inert` attribute and lets `processing`-driven UI (spinners, disabled inputs) reset.

If the user's `onSuccess` returns a Promise — for example, polling a background job before reloading — the form's processing state clears before that work finishes, defeating `disableWhileProcessing` and any UI bound to `processing`.

`useForm`'s internal submit already awaits `options.onSuccess` and returns its result, so the Form wrappers are the only place this awaitability is dropped. This PR propagates it.

### DIFF
--- a/packages/react/src/Form.ts
+++ b/packages/react/src/Form.ts
@@ -236,8 +236,8 @@ const Form = forwardRef<FormComponentRef, FormProps>(
         onProgress,
         onFinish,
         onCancel,
-        onSuccess: (...args) => {
-          onSuccess(...args)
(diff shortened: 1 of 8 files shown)


pascalbaljet:
Thanks!

PR #3125, opened 2026-05-24
[3.x] Pass props to withApp callback

CL0Pinette:
When the `withApp` callback was added, I felt like it was a regression to use it instead of setup because we couldn't access the props passed to the `initialPage` (which can be necessary in ssr mode to access user locale for example).

I make this PR to improve user experience when using the `withApp` callback to start inertia application

### DIFF
--- a/packages/react/src/createInertiaApp.ts
+++ b/packages/react/src/createInertiaApp.ts
@@ -30,7 +30,10 @@ type ComponentResolver = (
   page?: Page<SharedPageProps>,
 ) => ReactComponent | Promise<ReactComponent> | { default: ReactComponent }
 
-type ReactWithApp = (app: ReactElement, options: { ssr: boolean, props: InertiaAppProps<SharedProps> }) => ReactElement
+type ReactWithApp<SharedProps extends PageProps> = (
+  app: ReactElement,
+  options: { ssr: boolean; page: Page<SharedProps> },
+) => ReactElement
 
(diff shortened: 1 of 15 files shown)


mrleblanc101:
Perfectly timed.
Would solve #3126

pascalbaljet:
Thanks!

pascalbaljet:
I refactored it to the page object so you can access the component and URL as well.

mrleblanc101:
Thanks for the quick fix 🙌

mrleblanc101:
I realize the docs haven't been updated to reflect the new props.
I'll try to open a PR when I have some time.

pascalbaljet:
> I realize the docs haven't been updated to reflect the new props. I'll try to open a PR when I have some time.

I updated the docs right after the release: https://github.com/inertiajs/docs/pull/72

mrleblanc101:
@pascalbaljet I guess I should've scrolled more.
I searched for `withApp` and clicked [Customizing the App](https://inertiajs.com/docs/v3/installation/client-side-setup#customizing-the-app) and the first `i18n` example was unchanged.

Maybe this specific example should use `app.component()` (which doesn't rely on the app context) instead of `app.use(i18n)` (which most likely will use the app context to set the initial locale) to avoid confusion ?

Capture d’écran, le 2026-05-28 à 11 20 59

PR #3127, opened 2026-05-26
[3.x] Fix Svelte infinite scroll unmount race

hschne:
The Svelte adapter's `<InfiniteScroll>` component defers its observer setup with `setTimeout` in `onMount`, but did not cancel that timer in `onDestroy`.

```js
onMount(() => {
  setTimeout(setupInfiniteScrollInstance);
});

onDestroy(() => infiniteScrollInstance?.flush());
```

If the component is destroyed before the deferred callback runs, the stale callback can run against detached DOM. In practice this can pass `null` to `MutationObserver.observe()` and throw:

```text
TypeError: Failed to execute 'observe' on 'MutationObserver':
parameter 1 is not of type 'Node'.
```

This can happen when `<InfiniteScroll>` is conditionally rendered and quickly unmounted, for example during responsive layout switches, bottom-sheet/container state changes, or rapid `{#if}` toggles. This is Svelte-specific as far as I can tell. The React and Vue adapters are not affected in the same way

The fix deferred setup from running after the component has already been destroyed, while preserving the existing cleanup for initialized instances.

## Tests

No automated regression test is included, because timing issues like that are a PITA without introducing a bunch of additional dependencies (which I was not willing to do).


pascalbaljet:
Thanks for this, I was able to reproduce it! The `setTimeout()` wrapper has been there since the introduction of the component, but all tests pass without it, and removing it also fixes the race condition. So I've updated the PR by removing the `setTimeout()` altogether and adding tests.

hschne:
Thanks a lot! 

I was wondering if that timeout was needed, but didn't dare touch it 😅

PR #3134, opened 2026-06-02
[3.x] Instant visit: only preserve scroll behavior on instant visit

yoeriboven:
Whenever you make an instant visit, it will reset the scroll position if you don't use `preserveScroll`.

I think the `preserveScroll` setting should be used on the instant visit, but never on the server request.

## The problem
Here is an example of a slow route that we access using an instant visit. It works well but when you scroll down the page while the request is still going you will be teleported to the top.

https://github.com/user-attachments/assets/07470877-d09c-4931-b46c-7c3ae1f93cb9

You could change this behavior by setting `preserveScroll` on the visit but that means it won't just be applied to the real visit but also to the instant visit. Definitely not what you want.

https://github.com/user-attachments/assets/fe78af0b-e95c-4a12-907a-3f4014a428c6

## Solution
With this PR the `preserveScroll` behavior will be used on the instant visit and `preserveScroll` will be set to `false` on the server request.

https://github.com/user-attachments/assets/ed5e7d32-d525-4265-a0a5-3d549f0791fb

## Tests
I didn't find any tests for the other three properties that get custom values (`preserveState`, `replace` and `viewTransition`) so skipped it for `preserveScroll` as well.

## Playground
I added some code to the Vue 3 playground so the behavior can be tested. This code can be removed before merge.


pascalbaljet:
Thanks!

PR #3139, opened 2026-06-05
[2.x] Bump axios floor to ^1.16.0 to exclude vulnerable range

Andy9822:
### What & why

On the 2.x line, `axios` is a runtime dependency of `@inertiajs/core` (and a dev/playground dependency elsewhere), currently pinned to `^1.13.5`. That range still allows installs affected by two High-severity advisories:

- [GHSA-777c-7fjr-54vf](https://github.com/advisories/GHSA-777c-7fjr-54vf) - fetch adapter ignores request/response size limits, so it can be pushed into resource exhaustion (axios 1.7.0–1.15.x).
- [GHSA-hfxv-24rg-xrqf](https://github.com/advisories/GHSA-hfxv-24rg-xrqf) - ReDoS via the XSRF cookie name (axios ≤1.15.0).

Both were fixed in axios 1.16.0.

A fresh install today already resolves to a patched 1.16.x, but since `^1.13.5` overlaps the vulnerable range, `pnpm audit` (now running on push/releases via #3132) and downstream audits still flag it. Bumping the floor to `^1.16.0` keeps the affected versions out for good, and it's still the same major so nothing breaks.

### What changed

- `axios` `^1.13.5` → `^1.16.0` everywhere it's declared:
  - `packages/core` (the runtime `dependencies` one consumers actually get)
  - `packages/react`, `packages/svelte`, `packages/vue3` (`devDependencies`)
  - `playgrounds/{react,svelte4,svelte5,vue3}`
- Regenerated `pnpm-lock.yaml` (just the specifiers; the resolved version was already 1.16.1).

### Verification

- `pnpm audit --audit-level high` comes back clean — no axios advisories left. The few remaining entries are pre-existing moderate svelte/vite ones this PR doesn't touch.
- Lockfile diff is only the axios specifier bumps, nothing else drifts.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -54,7 +54,7 @@
   },
   "dependencies": {
     "@types/lodash-es": "^4.17.12",
-    "axios": "^1.13.5",
+    "axios": "^1.16.0",
     "laravel-precognition": "^1.0.2",
     "lodash-es": "^4.18.1",
     "qs": "^6.15.0"
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -55,7 +55,7 @@
   "devDependencies": {
     "@types/react": "^19.2.14",
     "@types/react-dom": "^19.2.3",
-    "axios": "^1.13.5",
+    "axios": "^1.16.0",
     "es-check": "^9.6.1",
     "esbuild": "^0.27.3",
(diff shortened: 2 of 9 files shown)


pascalbaljet:
Thanks!

PR #3142, opened 2026-06-07
[3.x] Do not throw handled useHttp validation errors

### LINKED ISSUE #3140: useHttp throws HttpResponseError on validation errors

- [ ] React
- [x] Vue 3
- [ ] Svelte
- [ ] Not Applicable

### JS package version

3.0.3

### Backend stack (optional)

Laravel 12.X
PHP 8.4

### Describe the problem

Requests made through the `useHttp` hook currently throw a `HttpResponseError` for `422` responses. This causes expected Laravel validation errors to bubble up to frontend error reporting, for example Sentry:

```
HttpResponseError - Request failed with status 422
```

The hook already contains logic to treat `422` responses as validation errors instead of HTTP exceptions:

```
if (error.response.status === 422) {
  const responseData = JSON.parse(error.response.data)
  const validationErrors = responseData.errors || {}
  const processedErrors = (
    withAllErrors.enabled() ? validationErrors : toSimpleValidationErrors(validationErrors)
  ) as FormDataErrors<TForm>

  form.clearErrors().setError(processedErrors)
  options.onError?.(processedErrors as Errors)
} else {
  options.onHttpException?.(error.response)
}
```

However, after this validation handling runs, the original `HttpResponseError` is still thrown.

This is unexpected for Laravel’s default validation response behavior. A `422` should be handled as form validation feedback by default: form errors should be set, `options.onError` should be called, and no `HttpResponseError` should be thrown automatically.

Consumers that want validation errors to throw can still do so explicitly inside their `onError` callback. Removing the automatic `HttpResponseError` throw for handled `422` responses would better match the current hook behavior and expected Laravel validation semantics.

### Steps to reproduce

1. Make a request using the `useHttp` hook.
2. Return a `422 Unprocessable Entity` response with validation errors from the backend.
3. Observe that a `HttpResponseError` is thrown despite the validation errors being handled by the form.

cyphercodes:
Fixes #3140.

## Summary

- Treat `useHttp` 422 responses as handled validation errors after setting form errors and firing `onError`
- Keep non-422 responses flowing through `onHttpException` and the existing `HttpResponseError` throw path
- Add regression coverage across the shared useHttp test apps to assert validation errors do not surface as thrown exceptions

## Benefit

This prevents expected Laravel validation responses from bubbling into frontend error reporting while preserving validation feedback on the form. Consumers can still throw from `onError` if they want validation failures to reject.

## Tests

- `pnpm install --frozen-lockfile`
- `pnpm exec oxfmt --check packages/vue3/src/useHttp.ts packages/react/src/useHttp.ts packages/svelte/src/useHttp.svelte.ts tests/use-http.spec.ts packages/vue3/test-app/Pages/UseHttp/Index.vue packages/react/test-app/Pages/UseHttp/Index.tsx packages/svelte/test-app/Pages/UseHttp/Index.svelte`
- `pnpm build:all`
- `PACKAGE=vue3 pnpm exec playwright test tests/use-http.spec.ts -g "it handles 422 validation errors" --project chromium`
- `PACKAGE=react pnpm exec playwright test tests/use-http.spec.ts -g "it handles 422 validation errors" --project chromium`
- `PACKAGE=svelte pnpm exec playwright test tests/use-http.spec.ts -g "it handles 422 validation errors" --project chromium`
- `pnpm type-check:test-app`
- `pnpm lint:test-app`
- `git diff --check`


pascalbaljet:
Thanks!

bretto36:
This one got me, i'd coded it up to use the catch on the useHttp in a few places in my code, but was sending 422. 

But a good change in the long run

PR #3151, opened 2026-06-10
[3.x] fix(core): ensure robust SSR error classification

### LINKED ISSUE #3150: Uncaught TypeErrors in ssrErrors (classifySSRError) and history (isValidState)
I noticed two minor robustness issues in `@inertiajs/core` that can lead to uncaught runtime TypeErrors under certain conditions:

1. **Unsafe SSR error classification (`ssrErrors.ts`)**:
When rendering components server-side (SSR), if a non-Error type is thrown (like a raw string, e.g. `throw "Render failed"` or a custom object), `classifySSRError` attempts to access `error.message.toLowerCase()`. Since `error.message` is `undefined`, this crashes the SSR server format handler itself with a `TypeError: Cannot read properties of undefined (reading 'toLowerCase')` instead of displaying the original error safely.

2. **Unsafe history state verification (`history.ts`)**:
In `history.isValidState(state)`, the check `!!state.page` is performed. However, if the browser history `state` is null (which can happen under various navigation conditions or when other scripts push/replace state), this throws `TypeError: Cannot read properties of null (reading 'page')`.

I have a branch ready fixing both of these issues safely, with unit tests added to verify the fixes. I will submit a PR linking to this issue shortly.

hamdyelbatal122:
### Description
This PR addresses two runtime robustness issues in `@inertiajs/core` (fixes #3150):

1. **Unsafe SSR error classification (`ssrErrors.ts`)**:
Under SSR, throwing non-Error types (such as raw strings or objects) would crash the error handler itself when it tried to read `error.message.toLowerCase()`. We now normalize the thrown value into an `Error` instance first.

2. **Unsafe history state verification (`history.ts`)**:
Accessing `state.page` in `isValidState` throws an error if the browser history `state` is null. We guarded this with `typeof state === 'object' && state !== null && 'page' in state`.

3. **VisitId Type Alignment**:
Updates `VisitId` type definition in `types.ts` from `unknown` to `string` to match the recent refactoring where visit IDs are strings.

### Verification
- Added a new unit test suite `history.test.ts` for history state verification.
- Added test cases in `ssrErrors.test.ts` for non-Error throwables.
- Verified that all unit tests (`pnpm --filter @inertiajs/core test`) and end-to-end playwright tests (`pnpm test:react -g "history.spec.ts"`) pass.


pascalbaljet:
Thanks! I've reverted the changes to `history.ts` because `isValidState` is only called by `handlePopstateEvent`, which already returns early when `state` is undefined.

PR #3156, opened 2026-06-12
fix(core): prevent error modal from flashing empty before content renders

### LINKED ISSUE #3130: Development error modal can flash empty before iframe srcdoc paints
### Summary

The development error modal can briefly show as an empty modal shell before the error HTML paints.

This is most noticeable when the server returns a rich HTML error page, such as Youch, Rails-style exception pages, Laravel Ignition-like pages, or any other non-Inertia HTML response with enough CSS/JS to take a moment to render inside the modal iframe.

### Current behavior

In `@inertiajs/core`, the dialog is opened before the iframe receives `srcdoc`:

```js
dialog.appendChild(iframe)
document.body.prepend(dialog)
dialog.showModal()
dialog.focus()
iframe.srcdoc = page.outerHTML
```

Because `showModal()` happens first, browsers can paint the dialog/backdrop before the iframe has parsed and rendered the error document. The result is a small but visible flash of an empty modal.

### Expected behavior

The modal should become visible only after the iframe has content ready to paint, so the developer sees the error page directly instead of an empty shell first.

### Suggested fix

Either set `iframe.srcdoc` before `showModal()`:

```js
dialog.appendChild(iframe)
document.body.prepend(dialog)
iframe.srcdoc = page.outerHTML
dialog.showModal()
dialog.focus()
```

Or, for the most robust version, keep the dialog hidden until the iframe has fired `load`, then call `showModal()` or reveal it.

### Why this matters

The modal itself is excellent because it lets developers see non-Inertia HTML error responses for XHR requests without digging through the Network tab. The small timing flash makes rich error pages feel less polished, especially when using framework development error renderers.

We hit this while integrating Youch-powered development errors into `inertia-sails` for Sails.js apps. We can work around it with app-level CSS or a small client observer, but the clean fix seems to belong in Inertia's modal implementation.

### Environment

- Inertia v3
- Browser: Chrome
- Adapter tested through `@inertiajs/vue3`
- Error response: HTML rendered into Inertia's development modal iframe

sawirricardo:
## Problem

Fixes #3130

The development error modal briefly shows as an empty dialog before the iframe error page content renders. This happens because `dialog.showModal()` is called before `iframe.srcdoc` is set.

## Current code (dialog.ts)

```ts
dialog.appendChild(iframe)
document.body.prepend(dialog)
dialog.showModal()          // ← dialog visible, but iframe is empty

dialog.focus()

iframe.srcdoc = page.outerHTML  // ← content set after dialog is already shown
```

## Fix

Move `srcdoc` assignment before `showModal()` so content is ready before the dialog becomes visible:

```ts
iframe.srcdoc = page.outerHTML   // ← content set first

dialog.appendChild(iframe)
document.body.prepend(dialog)
dialog.showModal()               // ← dialog shown with content ready
```

## Testing

Tested by triggering server errors in a Laravel + Inertia app with rich error pages (Ignition). The empty flash no longer occurs.

### DIFF
--- a/packages/core/src/dialog.ts
+++ b/packages/core/src/dialog.ts
@@ -70,12 +70,12 @@ export default {
       dialog.remove()
     })
 
+    iframe.srcdoc = page.outerHTML
+
     dialog.appendChild(iframe)
     document.body.prepend(dialog)
     dialog.showModal()
 
     dialog.focus()
-
-    iframe.srcdoc = page.outerHTML
   },
 }


pascalbaljet:
I'm closing this PR because it doesn't change anything in real-world usage. As with #3130, feel free to open a new issue that includes a real-world scenario and clear reproduction steps.

Issue #3143, opened 2026-06-07
Issues with Inertia `async` visit option


aalhoura:
### Inertia adapter(s) affected

- [x] React
- [x] Vue 3
- [ ] Svelte
- [ ] Not Applicable

### JS package version

3.3.1

### Backend stack (optional)

Laravel 13.14

### Describe the problem

I've identified several issues with the Inertia  `async` visit option when used under different scenarios.

**Issue 1:** 

If a `<Link>` with `async` points to a different endpoint than the one the user is currently on, then the asynchronous request cannot be triggered more than once. If the user clicks the asynchronous `Link` more than once, the previous asynchronous requests are cancelled and are replaced by the latest one. This does not happen if the `<Link>` with `async` points to the _same_ endpoint the user is currently on.

For example, if the user is currently on `/page1` and clicks an asynchronous `<Link>` pointing to `/page2` multiple times, only the latest asynchronous request is processed, and all the previous asynchronous requests are cancelled.

Clicking the `<Link>` to `/page2` multiple times does not work ❌
```tsx
export default function Page1() {
    return (
        <Link href="/page2" async>
            Go to page 2
        </Link>
    );
}
```
However, if the asynchronous `<Link>` points to `/page1` (the same page/endpoint the user is currently on), then the user can click the asynchronous `<Link>` multiple times and all of the asynchronous requests are processed without any cancellations.

Clicking the `<Link>` to `/page1` multiple times works ✅
```tsx
export default function Page1() {
    return (
        <Link href="/page1" async>
            Go to page 1
        </Link>
    );
}
```

**Issue 2:**

If a `<Link>` with `async` is clicked and the user navigates to a different page while the asynchronous request is still in flight, then the asynchronous request is cancelled. This does not happen if the user navigates to the _same_ page from which the asynchronous request was triggered. 

For example, if the user is currently on `/page1` and clicks an asynchronous `<Link>` pointing to `/page2`, then, while the asynchronous request is in flight, the user navigates to `/page3`, the asynchronous request to `/page2` is cancelled.

Clicking the `<Link>` to `/page3` cancels the asynchronous request ❌
```tsx
export default function Page1() {
    return (
        <>
            <Link href="/page2" async>
                Go to page 2
            </Link>

            <Link href="/page3">
                Go to page 3
            </Link>
        </>
    );
}
```
[...]

PR #3157, opened 2026-06-12
[3.x] fix(core): add async option to FormComponentOptions

sawirricardo:
## Problem

Related to #3143 (sub-issue 3)

The `async` visit option was missing from `FormComponentOptions`, causing `<Form options={{ async: true }}>` to be silently ignored. Form submissions were always synchronous regardless of the `async` prop.

## Fix

Add `"async"` to the `Pick<VisitOptions, ...>` in `FormComponentOptions`:

```ts
// Before
"preserveScroll" | "preserveState" | "preserveUrl" | "replace" | "only" | "except" | "reset" | "viewTransition"

// After  
"preserveScroll" | "preserveState" | "preserveUrl" | "replace" | "only" | "except" | "reset" | "viewTransition" | "async"
```

Both React (`...options` spread in `Form.ts:257`) and Vue3 (`...props.options` spread in `form.ts:304`) pass options directly to `router.visit()`, so no adapter changes needed — the type fix is sufficient.

## Usage after fix

```tsx
<Form action="/endpoint" method="post" options={{ async: true }}>
  <button type="submit">Submit asynchronously</button>
</Form>
```

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -772,7 +772,7 @@ export type UseHttpSubmitArguments<TResponse = unknown, TForm = unknown> =
 
 export type FormComponentOptions = Pick<
   VisitOptions,
-  'preserveScroll' | 'preserveState' | 'preserveUrl' | 'replace' | 'only' | 'except' | 'reset' | 'viewTransition'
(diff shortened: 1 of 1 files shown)


pascalbaljet:
Thanks!

PR #3158, opened 2026-06-12
fix(core): preserve async requests when navigating to a different page

sawirricardo:
## Problem

Related to #3143 (sub-issues 1 and 2)

When navigating to a different page, the router unconditionally cancelled **all** in-flight async requests via `cancelInFlight()`, including explicitly user-initiated async visits. This caused two issues:

1. **Async `<Link>` to a different endpoint**: clicking multiple times cancelled previous async requests instead of allowing concurrent requests
2. **Navigating away while async in-flight**: navigating to any different page cancelled pending async requests

Same-endpoint async requests worked because the `isSamePage` check skipped the cancellation block entirely.

## Root Cause

`router.ts:284-287` — `cancelInFlight({ prefetch: false, optimistic: false })` had no filter for `async`, so all non-prefetch, non-optimistic requests were cancelled on page change, including user-initiated async visits.

## Fix

Add an `async` filter to `cancelInFlight()` and pass `async: false` when cancelling on page navigation:

**requestParams.ts** — add `isAsync()` method
**request.ts** — add `isAsync()` delegating to requestParams  
**requestStream.ts** — add `async` filter to `cancelInFlight()`
**router.ts** — pass `{ prefetch: false, optimistic: false, async: false }` to preserve explicitly async requests

### Before
```ts
this.asyncRequestStream.cancelInFlight({ prefetch: false, optimistic: false })
// ↑ cancels ALL async requests including user-initiated async visits
```

### After
```ts
this.asyncRequestStream.cancelInFlight({ prefetch: false, optimistic: false, async: false })
// ↑ only cancels internal async requests (deferred props, partial reloads)
//   preserves explicitly async visits
```

## Note

This does not address sub-issue 3 (`<Form>` ignoring `async` prop) — that is a separate type-only fix in #3157.

Also, `response.ts` `shouldSetPage()` still discards async responses when the user has navigated to a different component/pathname. This is an intentional staleness guard but may need revisiting as a design decision (should async responses still update props in the background after navigation?).

### DIFF
--- a/packages/core/src/request.ts
+++ b/packages/core/src/request.ts
@@ -45,6 +45,10 @@ export class Request {
     return this.optimistic
   }
 
+  public isAsync(): boolean {
+    return this.requestParams.isAsync()
+  }
+
   public isPendingOptimistic(): boolean {
     return this.isOptimistic() && (!this.response || !this.response.isProcessed())
   }
--- a/packages/core/src/requestParams.ts
+++ b/packages/core/src/requestParams.ts
@@ -71,6 +71,10 @@ export class RequestParams {
     return this.params.deferredProps === true
   }
 
+  public isAsync(): boolean {
+    return this.params.async === true
+  }
+
   public onCancelToken(cb: VoidFunction) {
     this.params.onCancelToken({
       cancel: cb,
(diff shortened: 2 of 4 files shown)


pascalbaljet:
Thanks for the PR! I'm closing it because this has been fixed in #3198, which takes a slightly stricter approach: instead of exempting all async requests from cancellation, it only cancels the ones targeting the page you're navigating away from. That way async polls and reloads on the page you're leaving are still cancelled, while background async visits to other pages complete.

Issue #1096, opened 2022-02-11
Link new tab not working in vue


ayophanz:
I'm trying to add a link using the inertiajs build it link components

`<Link href="/blog/uncategorized/test-slug" target="_blank">Preview</Link>`

instead of redirect to new tab the line will remain in the current tab.

Please help
thanks

ajnsn:
Hey @ayophanz 
For this use case, you could just use a normal anchor-tag instead.

reinink:
Yup, exactly @ajnsn, thanks for helping out here 👍

amirmasoud:
If you have to do it inside of JS for whatever reason, one solution can be something like this:

```vue
<script setup>
const newTab = $ref(true);
const openInNewTab = (href) => {
  Object.assign(document.createElement("a"), {
    target: "_blank",
    rel: "noopener noreferrer",
    href: href,
  }).click();
};
</script>
<template>
  <button
    @click="newTab ? openInNewTab('/open-in-new-tab') : $inertia.visit('/inertia-link')"
  >
    click
  </button>
</template>
```
Reference\
[https://stackoverflow.com/a/28374344/2747661](https://stackoverflow.com/a/28374344/2747661)

vitalijalbu:
Neither in react works

andreapozza:
A cleaner solution: `const openInNewTab = href => window.open(href, '_blank')`
https://github.com/inertiajs/inertia/issues/1096#issuecomment-1419872292

amirmasoud:
`window.open` might be blocked by browsers as popup if it's not happening on the actual click event.

andreapozza:
Hi @amirmasoud, thank you for the explanation. I didn't think about that.

vedmant:
That's unfortunate, I have a component that optionallity opens in a new tab, in this case I have to conditionally use `<Component :is=""` to switch between just `a` and `Link` which is making the code more cumbersome. Btw, `NuxtLink` implementation supports `target` attribute

PR #3159, opened 2026-06-15
[3.x] Support anchor target attribute

devhammed:
First of all,  thanks for the work on Inertia!

The reason for this pull request is that I was working on a navigation link in my company's project, and some of the links require opening in a new tab. Of course, I added `target="_blank"` to the `Link` component as a concept I'm used to in libraries like React Router, Nuxt, and the like, but I was surprised that this isn't supported, judging from the comments in https://github.com/inertiajs/inertia/issues/1096 and after diving into the source code myself.

I will be open to backporting this to 2.x if it is merged.


pascalbaljet:
Thanks!

PR #3160, opened 2026-06-16
[3.x] feat: Add page to titleCallback

mrleblanc101:
Fix #3155

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,5 @@
   },
   "engines": {
     "pnpm": ">=11.1.1"
-  },
-  "packageManager": "pnpm@11.7.0+sha512.19cc852c120c7125760f2443ee6be0ca5b40f9f50598de1a09a1f177503e010e57c23c77646e01e761de59bf874fb22a3398c33ab9691fc13eb946b6f0f4d620"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -45,5 +45,6 @@
   },
   "engines": {
     "pnpm": ">=11.1.1"
-  }
+  },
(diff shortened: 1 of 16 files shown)


pascalbaljet:
Thanks!

mrleblanc101:
@pascalbaljet I think I forgot to update so types.
```
No overload matches this call.
  Overload 3 of 3, '(options?: InertiaAppOptionsAuto<PageProps & { [key: string]: unknown; name: string; locale: string; auth: Auth; currentTeam: Team; teams: Team[]; tenant: Tenant; }> | undefined): Promise<...>', gave the following error.
    Type '(title: any, page: any) => any' is not assignable to type 'HeadManagerTitleCallback'.
      Target signature provides too few arguments. Expected 2 or more, but got 1.
```

The code is:
```js
createInertiaApp({
    title: (title, page) => (title ? `${title} | ${page.props.tenant.name}` : page.props.tenant.name),
});
```

Capture d’écran, le 2026-06-25 à 11 28 38

And the types are defined inside `global.d.ts`:
```js
    export interface InertiaConfig {
        sharedPageProps: {
            name: string;
            locale: string;
            auth: Auth;
            currentTeam: Team;
            teams: Team[];
            tenant: Tenant;
            [key: string]: unknown;
        };
    }
}
```
And it work properly elsewhere in the app, for example in my header.vue (no type error):
```html
<img
    class="size-full object-cover object-center"
    :src="$page.props.tenant.logo_square"
    alt=""
/>
```

mrleblanc101:
@pascalbaljet Actually, after updating the runtime was working and running 3.5.0, but the types still resolved to 3.4.0 in the pnpm cache or something 🤔  I had to `rm -rf node_modules`

I still have an error where `page.props` is of type unknown, and I think it's a regression because it breaks everywhere in my app, not just the titleCallback.
Reverting to 3.4.0 solve the issue.

I'll open a new issue.

Capture d’écran, le 2026-06-25 à 11 59 41

PR #3163, opened 2026-06-18
[3.x] Fix invalid ARIA roles on default progress bar

### LINKED ISSUE #3162: Default progress bar uses invalid ARIA roles (role="bar" / role="spinner")
### Summary

The built-in progress bar renders with `role="bar"` and `role="spinner"`. Neither is a valid ARIA role (ARIA 1.2), so axe-core flags it as a critical `aria-roles` violation on any page where the bar shows during a visit.

### Source

`packages/core/src/progress-component.ts`, still present on `master` (`@inertiajs/core@3.4.0`) and in `2.3.x`:

```js
barSelector: '[role="bar"]',
spinnerSelector: '[role="spinner"]',
template: [
  '<div class="bar" role="bar">',
  '<div class="peg"></div>',
  '</div>',
  '<div class="spinner" role="spinner">',
  '<div class="spinner-icon"></div>',
  '</div>',
].join(''),
```

Looks like `role="bar"` and `role="spinner"` are only used as internal `querySelector` hooks, not for any AT purpose.

### Impact

This shows up on every Inertia app as a critical axe-core violation (SC 4.1.2 Name, Role, Value). It's a problem for anyone who has to pass automated a11y checks or produce a WCAG/VPAT report. The bar is decorative, so the invalid roles add no value but do create a reportable failure.

### Suggested fix

Two options:

1. Treat it as decorative. Add `aria-hidden="true"` to the wrapper and switch the internal selectors to a `data-*` attribute or class instead of `role`. This avoids any behaviour change.
2. Use the valid `role="progressbar"` on the bar element.

Option 1 is the cleaner one.

### Repro

Run axe-core (e.g. `@axe-core/playwright`) against any Inertia page while a visit is in flight. You get an `aria-roles` violation on `.bar` / `.spinner`.

Reproduced on `@inertiajs/core@2.3.18`, confirmed present in `master` / `3.4.0`.

atymic:
The default progress bar rendered with role="bar" and role="spinner", neither of which is a valid ARIA role, so axe-core flags it as a critical aria-roles violation on any page where the bar shows during a visit.

These roles were only used as internal querySelector hooks, not for any AT purpose. Switch the selectors to the existing .bar/.spinner classes and mark the wrapper aria-hidden, since the bar is decorative. No behaviour change.

Fixes #3162


pascalbaljet:
Thanks! I've broadened the selectors so it won't break for users passing a custom template with the old `role` attributes.

atymic:
Thanks Pascal :)

PR #3165, opened 2026-06-19
[3.x] Prevent unhandled rejection when a view transition is superseded

mortenhauberg:
When a new view transition starts before an in-flight one finishes (rapid navigation or an interrupted visit), the browser skips the active transition and rejects its `ViewTransition.ready` promise with an `AbortError`. Inertia never attached a handler to that promise, so the rejection surfaced as an `unhandledrejection` event and was reported to error trackers (e.g. Sentry).

This is distinct from the visibility guard added in #2957: that only prevents starting a transition while the document is hidden, it cannot catch a running transition being aborted by a newer one. Attach a no-op `.catch()` to `ready`, since being superseded is expected behaviour, not an error.

Note the rejection only surfaces as an unhandled rejection in WebKit (Safari); Chromium does not emit the event for a bare `ready` rejection, which is why the regression test asserts under WebKit.

[Spec](https://www.w3.org/TR/css-view-transitions-1/#ViewTransition-prepare):
> If document's active view transition is not null, then skip that view transition with an 'AbortError' DOMException.

As in #2957, Claude helped me write this fix.

### DIFF
--- a/packages/core/src/page.ts
+++ b/packages/core/src/page.ts
@@ -251,6 +251,13 @@ class CurrentPage {
     return new Promise((resolve) => {
       const transitionResult = document.startViewTransition(() => doSwap().then(resolve))
 
+      // Starting a new view transition before this one finishes (e.g. rapid
+      // navigation or an interrupted visit) causes the browser to skip the active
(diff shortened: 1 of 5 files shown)


pascalbaljet:
Thanks!

PR #3188, opened 2026-07-03
[3.x] Change Svelte render call to be asynchronous

### LINKED ISSUE #3187: Runtime error during Svelte SSR using top-level awaits
### @inertiajs/svelte Version

3.6.0

### Backend stack (optional)

_No response_

### Describe the problem

Using `await` in a Svelte component currently results in the https://svelte.dev/docs/svelte/runtime-errors#Server-errors-await_invalid runtime error. The simple fix is to call `render` asynchronously.

### Steps to reproduce

The bug can be reproduces by rendering a simple component:

```html
<script>
  const text = await Promise.resolve("hello world")
</script>

{text}
```

buhrmi:
fixes #3187 

this PR also enables the `async: true` compiler option, which will be the [new default in Svelte 6](https://svelte.jp/docs/svelte/runtime-errors#Client-errors-flush_sync_in_effect)

### DIFF
--- a/.github/workflows/playwright-chromium.yml
+++ b/.github/workflows/playwright-chromium.yml
@@ -7,13 +7,22 @@ permissions:
 jobs:
   test-chromium:
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.repository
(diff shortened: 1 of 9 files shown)


pascalbaljet:
Thanks!

PR #3194, opened 2026-07-08
[3.x] Preserve identity of untouched props in replaceProp

skryukov:
`router.replaceProp()` builds the new props object with `set(cloneDeep(currentProps), name, newValue)`, deep-cloning every prop to change one. Every prop gets a new object identity on each call, so memoized components that consume unrelated props re-render on every `replaceProp`, `appendToProp`, and `prependToProp`, and the clone itself costs O(all props) per call. Partial reloads already preserve the identity of untouched props when merging the response into the current page, so `replaceProp` was the odd one out.

This PR replaces the deep clone with a path-aware immutable set: only the containers along the path are copied, and everything untouched keeps its identity. The input props object is still never mutated, and the path syntax matches the previous behavior, including dot notation, array indices, and creating missing intermediate containers.

Includes a test-app page with a memoized child component that fails on the current behavior (the memoized component re-renders when an unrelated prop is replaced) and passes with this change, plus unit tests asserting parity with the previous `set` semantics and the new identity guarantees.

### DIFF
--- a/packages/core/src/objectUtils.ts
+++ b/packages/core/src/objectUtils.ts
@@ -63,36 +63,37 @@ const compareValues = (value1: any, value2: any): boolean => {
   }
 }
 
-// Path-aware copy-on-write "set": copies only the containers along `path` and
-// preserves the identity of everything untouched, matching how partial reloads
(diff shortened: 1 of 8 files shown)


pascalbaljet:
Thanks!

PR #3203, opened 2026-07-18
[3.x] Guard Form getFormData against a missing form element

### LINKED ISSUE #3202: Form: uncaught TypeError 'Failed to construct FormData' when form unmounts during a deferred dirty-state update
### Version

- `@inertiajs/react` 3.0.3 (also confirmed present in the latest `3.6.1` source)

### Describe the problem

`<Form>` can throw an uncaught `TypeError: Failed to construct 'FormData': parameter 1 is not of type 'HTMLFormElement'` when an `input`/`change`/`reset` event fires shortly before the form unmounts (e.g. the user navigates away right after typing).

The dirty-state tracker defers a read of the form element through `React.startTransition`, but `startTransition` callbacks are not cancelled on unmount. If the `<form>` is torn down before the transition flushes, `formElement.current` is `null` and `new FormData(null)` throws.

Relevant source (`dist/index.js`, same in 3.0.3 and 3.6.1):

```js
const getFormData = (submitter) => new FormData(formElement.current, submitter); // throws when formElement.current === null
const getData = (submitter) => formDataToObject(getFormData(submitter));

const updateDirtyState = (event) => {
  if (event.type === "reset" && event.detail?.[FormComponentResetSymbol]) {
    event.preventDefault();
  }
  deferStateUpdate(
    () => setIsDirty(event.type === "reset" ? false : !isEqual(getData(), formDataToObject(defaultData.current)))
  );
};

// deferStateUpdate:
var deferStateUpdate = (callback) =>
  typeof React.startTransition === "function" ? React.startTransition(callback) : setTimeout(callback, 0);
```

`updateDirtyState` is registered for `input`, `change`, and `reset` events. The effect cleanup removes the listeners, but a transition already scheduled by a prior event still runs after unmount, at which point `formElement.current` is `null`.

Observed in production (Chrome Mobile / Android). Stack trace (source-mapped) terminates in `getFormData` → `getData` → the `deferStateUpdate` callback in `updateDirtyState`, invoked from the form event listener.

### Steps to reproduce

1. Render an Inertia `<Form>`.
2. Fire an `input`/`change` on a field (or trigger a `reset`).
3. Unmount the form (client-side navigation) before the scheduled `startTransition` flushes.
4. The deferred `setIsDirty` runs, calls `getData()` → `new FormData(null)` → `TypeError`.

It is timing-dependent, so it reproduces most reliably by navigating away immediately after a keystroke/change on a page under some CPU load.

### Suggested fix

Guard `getFormData` against a missing form element so the deferred dirty-check safely no-ops after unmount:

```js
const getFormData = (submitter) =>
  formElement.current ? new FormData(formElement.current, submitter) : new FormData();
```

(Alternatively, skip the deferred update when `formElement.current` is null inside `updateDirtyState`.)

NiekNijland:
Fixes #3202.

### Problem

`<Form>` can throw an uncaught `TypeError: Failed to construct 'FormData': parameter 1 is not of type 'HTMLFormElement'` when an `input`/`change`/`reset` event fires shortly before the form unmounts (e.g. the user navigates away right after typing).

`updateDirtyState` (registered for `input`/`change`/`reset`) defers its `getData()` read through `deferStateUpdate`, which is `React.startTransition`. `startTransition` callbacks are **not** cancelled on unmount, and the effect cleanup only removes the listeners. So a transition scheduled by a prior event still runs after the `<form>` has been torn down, at which point `formElement.current` is `null` and `new FormData(null)` throws.

```ts
const getFormData = (submitter?: FormSubmitter): FormData => new FormData(formElement.current, submitter)

const updateDirtyState = (event: Event) => {
  // ...
  deferStateUpdate(() =>
    setIsDirty(event.type === 'reset' ? false : !isEqual(getData(), formDataToObject(defaultData.current))),
  )
}

const deferStateUpdate = (callback: () => void) => {
  typeof React.startTransition === 'function' ? React.startTransition(callback) : setTimeout(callback, 0)
}
```

Seen in production on Chrome Mobile / Android. Source-mapped stack terminates in `getFormData` → `getData` → the `deferStateUpdate` callback in `updateDirtyState`.

### Fix

Guard `getFormData` so it returns an empty `FormData` when the form element is gone. At that point the deferred dirty-state check is cosmetic (the form has already unmounted), so a no-op is safe.

```ts
const getFormData = (submitter?: FormSubmitter): FormData =>
  formElement.current ? new FormData(formElement.current, submitter) : new FormData()
```

### Scope

React only. The Vue 3 adapter's equivalent (`onFormUpdate`) updates `isDirty` **synchronously** (no `startTransition`), so its listener cannot run after unmount and it is not affected.

### Notes

`getFormData` is unchanged in every 3.x release through 3.6.1, so the issue affects all of them.

### DIFF
--- a/packages/react/src/Form.ts
+++ b/packages/react/src/Form.ts
@@ -128,7 +128,8 @@ const Form = forwardRef<FormComponentRef, FormProps>(
     const [isDirty, setIsDirty] = useState(false)
     const defaultData = useRef<FormData>(new FormData())
 
-    const getFormData = (submitter?: FormSubmitter): FormData => new FormData(formElement.current, submitter)
+    const getFormData = (submitter?: FormSubmitter): FormData =>
+      formElement.current ? new FormData(formElement.current, submitter) : new FormData()
 
     // Convert the FormData to an object because we can't compare two FormData
     // instances directly (which is needed for isDirty), mergeDataIntoQueryString()
--- a/packages/react/test-app/Pages/FormComponent/UnmountRace.tsx
+++ b/packages/react/test-app/Pages/FormComponent/UnmountRace.tsx
@@ -1,0 +1,22 @@
+import { Form } from '@inertiajs/react'
+import { useState } from 'react'
+
+export default function UnmountRace() {
+  const [show, setShow] = useState(true)
+
+  return (
+    <div>
+      <h1>Form Unmount Race</h1>
+
+      {show && (
+        <Form action="/dump/post" method="post">
+          <input type="text" name="name" id="name" defaultValue="John" />
+        </Form>
+      )}
+
+      <button id="hide" onClick={() => setShow(false)}>
+        Hide Form
+      </button>
+    </div>
+  )
+}
--- a/packages/svelte/test-app/Pages/FormComponent/UnmountRace.svelte
+++ b/packages/svelte/test-app/Pages/FormComponent/UnmountRace.svelte
@@ -1,0 +1,17 @@
+<script lang="ts">
+  import { Form } from '@inertiajs/svelte'
+
+  let show = $state(true)
+</script>
+
+<div>
+  <h1>Form Unmount Race</h1>
+
+  {#if show}
+    <Form action="/dump/post" method="post">
+      <input type="text" name="name" id="name" value="John" />
+    </Form>
+  {/if}
+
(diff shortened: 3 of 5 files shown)


pascalbaljet:
Thanks!

PR #3214, opened 2026-08-16
Add official Angular adapter

gerardp:
## Summary

This adds an official `@inertiajs/angular` adapter targeting Angular 22.1+.

- Implements the complete adapter surface, including `createInertiaApp`, `App`, `Head`, `Link`, `Form`, `Deferred`, `InfiniteScroll`, `WhenVisible`, `useForm`, `useHttp`, `usePage`, `usePoll`, `usePrefetch`, `useRemember`, and layout props.
- Adds Angular SSR support through the `@inertiajs/angular/server` entry point.
- Uses Angular's signal-based APIs and standalone components/directives. Inertia remains the sole owner of navigation, history, scroll, cache, and visits; Angular Router is intentionally not installed.
- Preserves Inertia callback return contracts with function-valued inputs while also exposing notification outputs for Angular templates.
- Adds an Angular test application implementing the shared adapter fixtures and connects it to the existing Playwright, quality, compatibility, build, and publish workflows.
- Documents the new adapter and development commands.

## Why

Angular users currently need a community integration or a custom bridge to use Inertia. Shipping the adapter in this monorepo gives Angular the same core-owned navigation semantics, shared compatibility suite, SSR support, and release process as the React, Vue, and Svelte adapters.

The implementation reuses the public contracts from `@inertiajs/core` instead of duplicating them, and it does not change the behavior of the existing adapters. Framework-specific test selectors and server handling were adjusted only where Angular renders equivalent markup differently.

## Validation

- `pnpm --filter @inertiajs/angular test` — 35 tests passed
- `pnpm --filter @inertiajs/angular lint`
- `pnpm --filter @inertiajs/angular publint`
- `pnpm --filter @inertiajs/angular es2022-check`
- `pnpm --filter @inertiajs/angular-test-app type-check`
- `PACKAGE=angular npx playwright test --project=chromium` — passed
- `PACKAGE=angular SSR=true npx playwright test --project=chromium` — 16 passed, 7 skipped Vite-transform-only tests
- `pnpm install --frozen-lockfile`
- `pnpm audit --audit-level=high` — no known vulnerabilities
- `pnpm publish --dry-run --no-git-checks`

Follow-up playground: #3215. It is submitted immediately so the adapter can be evaluated in a complete Laravel application without mixing that application into this already substantial adapter diff.


pascalbaljet:
Thanks for the PR, and for the amount of work that went into this. We don't want to add more adapters to this repo for now. That said, I'd love to see this live on as a community-driven adapter!

I'll close this one along with #3215.

Issue #1089, opened 2022-02-07
ReferenceError: document is not defined inertia ssr 


samehdoush:
### Versions:

- `@inertiajs/inertia` version: 0.11.0
- `@inertiajs/inertia-vue` version: 0.8.0

### Describe the problem:
laravel 8

inertia js : 0.5.4

package.json
```
       "@inertiajs/inertia": "^0.11.0",
        "@inertiajs/inertia-vue": "^0.8.0",
   "laravel-mix": "^6.0",
   "sass": "~1.32",
        "sass-loader": "^12.2.0",
        "vue-cli-plugin-vuetify": "~2.3.1",
        "vue-loader": "^15.9.6",
        "vue-template-compiler": "^2.6.10",
        "vuetify-loader": "^1.7.0",
        "webpack": "^5.59.1",
 "vue": "^2.6.14",
     "vue-server-renderer": "^2.6.14",
   "vuetify": "^2.5.5",
        "vuetifyjs-mix-extension": "0.0.20",
        "vuex": "^3.4.0",
        "webpack-node-externals": "^3.0.0"

```

webpack.ssr.mix.js
```
const path = require('path')
const mix = require('laravel-mix')
const webpackNodeExternals = require('webpack-node-externals');


mix
    .options({ manifest: false })
    .js('resources/js/ssr.js', 'public/js')
 
    .vue({
        version: 2, options: { optimizeSSR: true }
    })
    .webpackConfig({
        resolve: {
            alias: {

                '@resources': path.resolve('resources'),
                '{Template}': path.resolve('resources/js/Themes/default'),
                '@themeConfig': path.resolve('themeConfig.js'),
                '@core': path.resolve('resources/@core'),
                '@axios': path.resolve('resources/js/plugins/axios.js'),
                '@user-variables': path.resolve('resources/sass/variables.scss'),
                '@sass': path.resolve('resources/sass/'),
                'apexcharts': path.resolve(__dirname, 'node_modules/apexcharts-clevision'),
                '@': path.resolve('resources/js'),
                // @resources/assets/images
            },
        },
   
        target: 'node',
        devtool: 'source-map',
        externals: [webpackNodeExternals({
            // this WILL include `jquery` and `webpack/hot/dev-server` in the bundle, as well as `lodash/*`
[...]

PR #3218, opened 2026-08-17
[3.x] Add ClientOnly component

evan-burrell:
Adds a `ClientOnly` component to the React, Vue and Svelte adapters, addressing the request in discussion #1722 (open since 2023, 9 upvotes).

## The problem

Under SSR, any component that touches `window` or `document` while rendering crashes the render the recurring "`document is not defined`" / "`window is not defined`" class of report (#1089, and #3021 was this same problem inside `WhenVisible`). Plenty of third-party widgets charts, maps, editors, anything measuring the viewport simply cannot run on the server.

Today the options are:

- **Disable SSR for the whole page** via `inertia.ssr.enabled` (the recommendation in discussion #1429). Works, but it is per-endpoint and all-or-nothing you lose server rendering for the entire page to accommodate one widget.
- **Lean on `Deferred`.** A deferred prop is absent during SSR, so the fallback renders on the server and the children render once the prop arrives. This produces the right visible shape, but it requires inventing a deferred prop you may not otherwise want and it costs an HTTP round-trip to render something that needs no server data at all.

Neither lets you keep SSR for the page and carve out one subtree.

## The component

```jsx
// React
<ClientOnly fallback={<p>Loading chart...</p>}>
  {() => <Chart width={window.innerWidth} />}
</ClientOnly>
```

```vue
<!-- Vue -->
<ClientOnly>
  <Chart />
  <template #fallback><p>Loading chart...</p></template>
</ClientOnly>
```

```svelte
<!-- Svelte -->
<ClientOnly>
  <Chart />
  {#snippet fallback()}<p>Loading chart...</p>{/snippet}
</ClientOnly>
```

The children render only after the component has mounted in the browser; the `fallback` renders during SSR and during the initial hydration render.

The one design point worth calling out: this is a **mount flag, not a `typeof window` check**. On the client the first render is a hydration of the server's HTML, so gating on `window` alone would render the children immediately and mismatch. Waiting for mount keeps both renders identical. The tests assert no hydration errors to hold that in place.

## API

I followed the existing adapter conventions rather than Nuxt's naming:

- **`fallback`** - matches `Deferred` and `WhenVisible`, rather than introducing `placeholder` as a third name for the same concept.
- **No `fallbackTag` / `placeholderTag`.** Nuxt needs those because its `placeholder` is a plain string; a slot/snippet already covers it here.

In React, `children` accepts either a node or a function. The function form matters because JSX children are built eagerly, so inline `window` access would still execute on the server. Vue slots and Svelte snippets are already lazy, so they need no equivalent.

## Scope - what this does not do

Nuxt's `ClientOnly` also **tree-shakes the wrapped component out of the server build**, so its code never reaches the server. This implementation is runtime-only: the children's module is still imported server-side, it just isn't rendered. That is sufficient for the `window`/`document` access case in #1722, but it will **not** strip a heavy or server-incompatible dependency from the SSR bundle. Worth being explicit about, since the name invites the comparison.

## Tests

Per CONTRIBUTING, the same pages exist in all three test apps and the Playwright specs are shared:

- `tests/client-only.spec.ts` - client-rendered path: children render after mount, and again after back-navigation.
- `tests/ssr.spec.ts` - SSR path: the server HTML contains the fallback and **not** the children, then the children replace it after hydration with no hydration errors and no console errors.

Both cover React, Vue and Svelte. The React non-SSR page uses the plain-node `children` form and the SSR page uses the function form, so both branches of the React API are exercised.

Verified locally:

| Adapter | `pnpm test:*` | `pnpm test:ssr:*` |
| --- | --- | --- |
| React | 1177 passed | 69 passed |
| Vue | 1183 passed | 69 passed |
| Svelte | 1142 passed | 63 passed |

`oxfmt --check`, `oxlint`, `tsc` / `vue-tsc` / `svelte-check`, the ES2022 check and all package builds are clean.

One note on the Svelte suite: `infinite-scroll.spec.ts:885` ("Remember state › ... dataset verification") is flaky on my machine it failed twice and then passed, and it also passes on an unmodified `3.x` checkout. It is scroll-restore-after-refresh timing and unrelated to this change, but flagging it in case CI trips on it.

## Why it does not break anything

Purely additive a new component per adapter plus one export line each. No existing file's behaviour changes, and nothing in `packages/core` is touched. Apps that never import `ClientOnly` are unaffected.

## Docs

Happy to open a PR against `inertiajs/docs` if you want this. My initial thought would be a section in `v3/advanced/server-side-rendering.mdx` rather than a new page but I held off until you have had a chance to weigh in on the component itself and which version it should target.

### DIFF
--- a/packages/core/src/hydrationBoot.ts
+++ b/packages/core/src/hydrationBoot.ts
@@ -1,0 +1,22 @@
+let isHydrationBoot = true
+let clientRendered = false
+
+// The `typeof window` check in `canRenderClientOnly` is what actually keeps this module
+// inert on the server -- a client boot never runs inside the SSR process, so the state
+// below is never touched there regardless. The guards on the mutators are defence in
+// depth: these are exported functions, so nothing stops a caller from invoking them
+// outside the client boot path they're meant for.
+export function setHydrationBoot(value: boolean): void {
+  if (typeof window === 'undefined') return
+  isHydrationBoot = value
+  clientRendered = false
+}
+
+export function markClientRendered(): void {
+  if (typeof window === 'undefined') return
+  clientRendered = true
+}
+
+export function canRenderClientOnly(): boolean {
+  return typeof window !== 'undefined' && (clientRendered || !isHydrationBoot)
+}
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -10,6 +10,8 @@ export { hasFiles } from './files'
 export { objectToFormData } from './formData'
 export { formDataToObject } from './formObject'
 export { default as createHeadManager, resolveServerHead } from './head'
+/** @internal Not part of the public API. May change or be removed without notice. */
+export { canRenderClientOnly, markClientRendered, setHydrationBoot } from './hydrationBoot'
 export { http } from './http'
 export { HttpCancelledError, HttpError, HttpNetworkError, HttpResponseError } from './httpErrors'
 export { default as useInfiniteScroll } from './infiniteScroll'
--- a/packages/core/tests/hydrationBoot.test.ts
+++ b/packages/core/tests/hydrationBoot.test.ts
@@ -1,0 +1,51 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+// The module holds mutable top-level state, so each test needs a fresh instance.
+const loadFresh = async () => {
+  vi.resetModules()
+  return import('../src/hydrationBoot')
+}
+
+describe('hydrationBoot', () => {
+  afterEach(() => {
+    vi.unstubAllGlobals()
+  })
+
+  it('blocks rendering until markClientRendered runs, on a hydration boot', async () => {
+    vi.stubGlobal('window', {})
+    const { canRenderClientOnly, markClientRendered } = await loadFresh()
+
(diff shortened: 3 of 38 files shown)


pascalbaljet:
Thanks for this! I'm not against it, but I have to think a bit about the naming and the mechanics once the client is loaded.

The fallback should really only render during the hydration render. After that, any subsequent navigation should render the default slot, and maybe in apps without SSR it shouldn't render at all. I think right now every visit remounts the component, so you get a flash of the fallback even when there's no server render involved.

evan-burrell:
Thanks for the feedback @pascalbaljet! I've been waiting to find something to contribute back since we met at Laravel Live UK back in 2024! This is something we find ourselves reaching for on projects where they have pieces that break the server side rendering with client side pieces. I've given your feedback a first pass, let me know if I can help or clarify anything.

pascalbaljet:
@evan-burrell I refactored it a bit and renamed the component to `<WhenMounted>` so it pairs well with `<WhenVisible>`, naming by intent. I also moved the hydration state out of `packages/core` into the Vue and Svelte adapters, since React doesn't need it.

evan-burrell:
Looks great! Thanks @pascalbaljet

PR #3230, opened 2026-08-24
[3.x] Prevent unhandled rejection when a view transition is aborted mid-flight

### LINKED ISSUE #3229: View transition: `finished` rejection is unhandled in Chromium (follow-up to #3165)
### @inertiajs/vue3 Version

3.6.1 — the relevant code is unchanged in 3.7.0, which I checked against the published package

### Backend stack (optional)

Laravel 13.19.0 / PHP 8.5.7 / `inertiajs/inertia-laravel` 3.1.1 — not backend related, listed for completeness

### Describe the problem

#3165 attached a no-op catch to `ViewTransition.ready`, which fixed the unhandled rejection in WebKit — that PR's description notes it "only manifests as an unhandled rejection in WebKit/Safari, not in Chromium".

Chromium rejects a different promise. When a transition is aborted, `finished` rejects as well, and that one still has no handler, so it reaches `onunhandledrejection`:

```
InvalidStateError: Transition was aborted because of invalid state
```

In `packages/core/src/page.ts`, `swap()`:

```ts
const transitionResult = document.startViewTransition(() => doSwap().then(resolve))
transitionResult.ready.catch(() => {})
viewTransitionCallback(transitionResult)
```

What we see in our production error tracker over 30 days, with `viewTransition` enabled for all real navigations:

- circa 100 events from a few dozen distinct users, all via `onunhandledrejection`
- Browsers: roughly an even split between Chrome and Edge, plus a handful from mobile Chromium and Opera — nothing from Gecko
- Spread across every kind of page (list pages, detail pages, calendar, login, register, password reset), so it is not tied to a specific component

Two sibling messages appear in the same project and look like the same root cause, at circa 30 events each, though we did not break those down by browser:

- `View transition was skipped because document visibility state is hidden.`
- `Skipped ViewTransition due to document being hidden`

The guard added in #2957 checks `document.visibilityState` before starting the transition, but a tab can also become hidden *during* one.

**Expected behaviour:** an aborted or skipped view transition is expected — the page simply swaps without the cross-fade — so it should not produce an unhandled rejection, the same way #3165 established for `ready`.

**Suggested fix:** treat the other two promises the same way as `ready`:

```ts
transitionResult.ready.catch(() => {})
transitionResult.finished.catch(() => {})
transitionResult.updateCallbackDone.catch(() => {})
```

A no-op catch does not stop `viewTransitionCallback` consumers from attaching their own handlers, so this should not change public behaviour. As a workaround we wrap `document.startViewTransition` in our app entry and attach no-op catches to all three promises.

### Steps to reproduce

1. In a Vue app on `@inertiajs/vue3` 3.6.1 or 3.7.0, enable view transitions for navigations:

```ts
createInertiaApp({
    defaults: {
        visitOptions: () => ({ viewTransition: true }),
    },
    // ...
})
```

2. Open the app in Chrome or Edge with the console visible.
3. Click a link, then click a second link before the cross-fade has finished.
4. Repeat a few times.

`InvalidStateError: Transition was aborted because of invalid state` shows up as an unhandled rejection in the console, and in any error tracker listening for `unhandledrejection`. The behaviour is not app specific — it follows from `finished` having no rejection handler in `swap()`.

lazerg:
Follow-up to #3165, which attached a no-op catch to `ready` to stop a superseded view transition from surfacing as an unhandled rejection. Chromium and Edge have the same problem with two other promises: when a transition is aborted (a newer transition supersedes it, or the tab goes hidden mid-flight), `finished` and `updateCallbackDone` also reject with an `InvalidStateError`, and neither had a handler, so it reaches `onunhandledrejection` and gets picked up by error trackers.

This attaches the same no-op catch to both, matching how `ready` is already handled in `swap()`, and adds a regression test that backgrounds the tab while a transition is in flight to trigger the abort.

Fixes #3229.

### DIFF
--- a/packages/core/src/page.ts
+++ b/packages/core/src/page.ts
@@ -256,8 +256,8 @@ class CurrentPage {
     return new Promise((resolve) => {
       const transitionResult = document.startViewTransition(() => doSwap().then(resolve))
 
-      // A newer transition aborts this one, rejecting `ready` with an AbortError.
-      // That's expected, so swallow it to avoid an unhandled rejection.
+      // A newer transition, or the tab going hidden mid-flight, aborts this one,
+      // rejecting these promises. That's expected, so swallow it to avoid an unhandled rejection.
       transitionResult.ready.catch(() => {})
       transitionResult.finished.catch(() => {})
       transitionResult.updateCallbackDone.catch(() => {})
--- a/packages/core/src/page.ts
+++ b/packages/core/src/page.ts
@@ -253,14 +253,23 @@ class CurrentPage {
 
     const viewTransitionCallback = typeof viewTransition === 'boolean' ? () => null : viewTransition
 
+    // The browser skips this transition when a newer one supersedes it, when the tab goes hidden
+    // mid-flight, or when the swap times out, always rejecting with a DOMException. That's
+    // expected, so swallow it and let a failing swap reject as it normally would.
+    const ignoreSkippedTransition = (promise: Promise<unknown>) => {
+      promise.catch((error) => {
+        if (!(error instanceof DOMException)) {
+          throw error
+        }
+      })
+    }
+
     return new Promise((resolve) => {
       const transitionResult = document.startViewTransition(() => doSwap().then(resolve))
 
-      // A newer transition, or the tab going hidden mid-flight, aborts this one,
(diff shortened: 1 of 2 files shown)


pascalbaljet:
Thanks!

PR #3232, opened 2026-08-25
[3.x] Support for Live props

pascalbaljet:
This PR introduces *live props*: props that automatically refresh whenever a broadcast event is received, keeping the page in sync with the server without any user interaction.

### Basic usage

On the server-side, mark a prop as live and pass the event that should refresh it. Inertia reads the channels from the event itself, so you don't have to repeat them.

```php
Inertia::render('Orders/Show', [
    'order' => Inertia::live(fn () => $order->fresh(), on: new OrderUpdated($order)),
]);
```

On the front-end, all you have to do is pass a transport to `createInertiaApp`. The `echo()` helper uses the Echo instance that `configureEcho()` set up, which is all a Laravel app needs.

```js
import { echo } from '@inertiajs/vue3/echo'
import { configureEcho } from '@laravel/echo-vue'

configureEcho({ broadcaster: 'reverb' })

createInertiaApp({
  live: echo(),
})
```

That's it. Whenever `OrderUpdated` is broadcast, Inertia performs a partial reload for the `order` prop. Props that share a channel and event are refreshed together in a single request.

`@laravel/echo-{vue,react,svelte}` is an optional peer dependency, so nothing changes for apps that don't use live props.

### Events and channels

You may pass an array of events, and you may pass the `channel` argument to override the channels the events broadcast on.

```php
Inertia::live($value, on: [new OrderUpdated, new OrderCancelled]);

Inertia::live($value, on: 'order.updated', channel: new PrivateChannel('orders.1'));
```

Public, private, presence and encrypted private channels are all supported.

### Throttling

Props that change more often than you want to fetch them may be throttled. You may pass the number of milliseconds per prop, or set a default for the whole app.

```php
Inertia::live($value, on: new ActivityLogged, throttle: 4000);
```

```js
createInertiaApp({
  live: echo({ throttle: 2000 }),
})
```

Events that arrive within a few milliseconds of each other are collected into one partial reload, and a prop that a request in flight already claims is skipped instead of being fetched twice. Reloads are held while the tab is hidden and flushed when it comes back. You may disable that with the `pauseWhenHidden` option.

```js
createInertiaApp({
  live: echo({ pauseWhenHidden: false }),
})
```

### Deferred, optional and always props

`Inertia::defer()`, `Inertia::optional()` and `Inertia::always()` all accept `->live()`, so a prop may be deferred on the initial load and live from then on.

```php
Inertia::render('Orders/Show', [
    'activity' => Inertia::defer(fn () => $order->activity)->live(on: new ActivityLogged($order)),
]);
```

### Loading state

The new `useProp` hook tells you whether a prop is currently being fetched, which is nicer than a full loading state for something that refreshes in the background. It tracks in-flight partial requests, so it works for deferred props as well.

```vue
<script setup>
import { useProp } from '@inertiajs/vue3'

const { value, loading, loaded } = useProp('order')
</script>

<template>
  <span v-if="loading">Refreshing...</span>
</template>
```

### Socket IDs

Inertia sends the socket id of the connection along as the `X-Socket-Id` header on every request, so `broadcast(...)->toOthers()` works just like it does anywhere else in Laravel. You may also register a resolver yourself.

```js
import { socketId } from '@inertiajs/vue3'

socketId.resolveUsing(() => mySocket.id)
```

### Custom transports

Live props are not tied to Laravel Echo. The `echo()` helper is just one *transport*, and a transport is any object that turns a channel and an event name into a subscription. Anything you can subscribe to works, whether that's a raw WebSocket, an SSE stream, Ably, or a broadcaster from another ecosystem.

```js
createInertiaApp({
  live: {
    transport: {
      subscribe(channel, event, handler) {
        const stop = mySocket.on(`${channel.name}:${event}`, handler)

        return () => stop()
      },
      socketId: () => mySocket.id,
      onStatusChange: (callback) => mySocket.onStatus(callback),
    },
  },
})
```

Only `subscribe` is required. A transport that reports a `socketId` gets it sent along with every request, and a transport that reports connection status gets every live prop refreshed after a reconnect, since the events that fired while the connection was down never arrived.

### Reacting to events

Every incoming event fires a cancelable `live` event with the props it feeds, along with the channel, event name and payload. Returning `false` prevents the reload, so you may decide per event whether a refresh is worth a request.

```js
router.on('live', (event) => {
  if (event.detail.payload.status === 'draft') {
    return false
  }
})
```


pascalbaljet:
### Broadcasting prop data

Live props refresh with a partial reload, so the value is whatever your controller resolves at that moment. That is the right default, though it does mean one event turns into one request per open tab.

An event may now carry the new values with it. Pass them through `Inertia::broadcastProps()` from `broadcastWith()`:

```php
public function broadcastWith(): array
{
    return Inertia::broadcastProps([
        'order' => OrderResource::make($this->order),
    ]);
}
```

Inertia writes those values straight into the page, so no request is made at all. You may spread the result alongside anything else the event already sends.

A value only ever lands on a prop that declared this event, so a page that names the prop differently, or does not mark it live, ignores what it cannot use and refreshes as it always did.

*One thing worth knowing before you carry a value: a broadcast delivers the same data to everyone on the channel, while a refreshed prop is resolved per request with that user's session and policies. So only carry a value that is the same for every subscriber.*

### Pusher

Echo is not required. You may use `pusher-js` directly by passing it to `pusher()`.

```js
import { pusher } from '@inertiajs/core/pusher'
import Pusher from 'pusher-js'

createInertiaApp({
  live: pusher({ client: new Pusher(key, options) }),
})
```

Configuring Pusher, including channel authorization, stays with the application. `throttle` and `pauseWhenHidden` work just like they do with `echo()`.

### Events as class names

You may pass a class name to `on`, not just an instance. Inertia builds the event without its payload to read its channels and its broadcast name, which works whenever neither depends on one.

```php
Inertia::live(fn () => $order->fresh(), on: OrderUpdated::class);
```

An event that builds its channel from the payload, such as `new PrivateChannel('orders.'.$this->order->id)`, has nothing to read from and throws with a message telling you to pass an instance. The same goes for a `broadcastAs()` built from the payload.

skryukov:
Hey @pascalbaljet awesome addition! Couple questions:

1. Do you think we can add an Action Cable adapter for Rails to core?
2. Do you plan to support live merge props (I see there's an error raised there atm)? When I think about live props, I see chat, and chat is a paginated resource, and as far as I understand we can't use live props with this kind of stuff. I know this requires a 'deleted' signal implemented, and would love to hear your thoughts on this.

PR #3234, opened 2026-08-26
[3.x] Clear the prefetch hover timer when a link is clicked

### LINKED ISSUE #3233: Link prefetch: hover timer is not cleared on click, so a late prefetch remounts the page just navigated to
### Version:

- `@inertiajs/react` version: 3.6.1
- `@inertiajs/core` version: 3.6.1

### Describe the problem:

`<Link prefetch>` clears its hover timer on `onMouseLeave` and on unmount, but **not on click**. If you
click inside `prefetch.hoverDelay` (75ms by default), the timer fires *after* the visit has started, and
the resulting prefetch is applied to the page you have just navigated to — remounting the component tree
and destroying any open dropdown or dialog and all page-level state.

This is not the same as [#2066](https://github.com/inertiajs/inertia/issues/2066) (fixed by
[#2097](https://github.com/inertiajs/inertia/pull/2097)). That fix added the guard in `router.prefetch`:

```js
if (visitUrl === currentUrl) { return; }
```

The gap is that this guard is evaluated at **request** time, while `Response.handlePrefetch()` decides
whether to apply at **response** time — and the URL changes in between. Concretely: the timer fires
while `window.location` is still the *old* URL, so the guard passes and the request goes out; ~60ms
later the response lands, `window.location` is now the target URL, and `handlePrefetch` applies it:

```js
async handlePrefetch() {
    if (isSameUrlWithoutHash(this.requestParams.all().url, window.location)) {
        this.handle();
    }
}
```

`Response.process()` has no `processed` guard, so this second pass falls through to `setPage()`.

Three behaviours compose into it; none is wrong alone:

1. `Link`'s hover timer is not cleared on click:

```js
const prefetchHoverEvents = {
    onMouseEnter: () => { hoverTimeout.current = window.setTimeout(
        () => { doPrefetch(); }, config.get("prefetch.hoverDelay")); },
    onMouseLeave: () => { clearTimeout(hoverTimeout.current); },
    onClick: regularEvents.onClick          // no clearTimeout
};
```

2. `router.prefetch`'s same-URL guard is request-time only, as above.
3. `handlePrefetch` applying a same-URL response is intentional, and correct for the case it was written
   for.

An **in-flight** prefetch is not affected — the visit consumes it cleanly. I measured that specifically,
because it changed how large the fix needed to be. Sweeping the hover-to-click delay, two runs each:

| hover → click | page remounted |
|---|---|
| 0ms | yes |
| 40ms | yes |
| 90ms | no |
| 200ms | no |

At 90ms the prefetch was still in flight when the click landed, and nothing remounted. Only a prefetch
*issued* after the click is harmful, which suggests clearing the timer on click is a complete fix rather
than a mitigation.

### Steps to reproduce:

1. A list page whose rows are `<Link prefetch>` to a detail page.
2. The detail page renders a dropdown (or any component holding local state) — a Radix
   `DropdownMenu` in my case.
3. Hover a row and click it within 75ms — a fast pointer click, or trivially reliable under automation
   where the hover-to-click gap is ~0ms.
4. Open the dropdown as soon as the detail page renders.

The dropdown closes on its own, or never opens. Under automation it presents as
`expect(locator).toBeVisible() failed / element(s) not found` on an element that did exist a moment
earlier.

A cleaner observation than watching the dropdown: React's `useId` runs once per component instance, so
Radix's generated trigger id changes when the page remounts. Collect the id with a `MutationObserver`
installed before page scripts run — reading it after the element is actionable is too late, because the
remount has already completed by then.

### Measured trace

From a Playwright trace, times in ms on the trace's own clock. Requests distinguished by the
`purpose: prefetch` header:

```
83935.5  GET /projects/3          (real visit)               67.5ms
84003    ↳ response arrives
84018.1  GET /projects/3          purpose: prefetch           62.9ms   ← timer fires; location still /projects
84044    trigger id = radix-_r_12_
84052    click on the dropdown trigger begins
84081    ↳ prefetch response → handlePrefetch → setPage()
84101    click reported done
84105    trigger id = radix-_r_2u_                                     ← new instance: remount
89112    the menu item never appeared; assertion times out
```

### Suggested fix

Adding a `clearTimeout` to `prefetchHoverEvents.onClick`, matching what `onMouseLeave` already does,
closes the window — the click always precedes it. That leaves the intentional behaviour of
`handlePrefetch` untouched.

A defence in depth would be for `handlePrefetch` to decline a response whose URL was not the current URL
when the request started, but the one-line timer fix appears sufficient given the in-flight measurements
above.

### Notes

Possibly adjacent: [#3210](https://github.com/inertiajs/inertia/issues/3210) also concerns a click
interacting with an in-flight prefetch, though the symptom there is a lost `once` prop rather than a
remount.

Happy to open a PR for the `clearTimeout` if that shape is agreeable. I have worked around it locally
with a wrapper component that passes `prefetch={false}` and owns the hover timer itself, so I am not
blocked — filing because the composition seemed worth recording upstream.

lazerg:
A `<Link prefetch>` clears its hover timer on mouse leave and on unmount, but not on click. Clicking within the 75ms hover delay lets the timer fire after the visit has already started. At that point `window.location` is still the old URL, so `router.prefetch` doesn't recognize the target as the current page and sends the request anyway. The response lands after the visit has swapped the page, `handlePrefetch` sees a matching URL and applies it, remounting the whole component tree and wiping any local state on the page you just navigated to.

The hover handler now clears the timer on click, the same way it already does on mouse leave. React, Vue and Svelte all had the same gap, so all three are fixed here.

Fixes #3233.

### DIFF
--- a/packages/react/src/Link.ts
+++ b/packages/react/src/Link.ts
@@ -247,7 +247,11 @@ const Link = forwardRef<unknown, InertiaLinkProps>(
       onMouseLeave: () => {
         clearTimeout(hoverTimeout.current)
       },
-      onClick: regularEvents.onClick,
+      onClick: (event: React.MouseEvent) => {
+        clearTimeout(hoverTimeout.current)
+
+        regularEvents.onClick(event)
+      },
     }
 
     const prefetchClickEvents = {
--- a/packages/svelte/src/link.ts
+++ b/packages/svelte/src/link.ts
@@ -63,22 +63,24 @@ function link(
   let baseParams: VisitOptions
   let visitParams: VisitOptions
 
+  const handleClick = (event: MouseEvent) => {
+    if (shouldIntercept(event)) {
+      event.preventDefault()
+      router.visit(href, visitParams)
+    }
+  }
+
   const regularEvents: ActionEventHandlers = {
-    click: (event: MouseEvent) => {
-      if (shouldIntercept(event)) {
-        event.preventDefault()
-        router.visit(href, visitParams)
-      }
-    },
+    click: handleClick,
   }
 
   const prefetchHoverEvents: ActionEventHandlers = {
     mouseenter: () => (hoverTimeout = setTimeout(() => prefetch(), config.get('prefetch.hoverDelay'))),
     mouseleave: () => clearTimeout(hoverTimeout),
-    click: (event: PointerEvent) => {
+    click: (event: MouseEvent) => {
       clearTimeout(hoverTimeout)
 
-      regularEvents.click?.(event)
+      handleClick(event)
     },
   }
 
--- a/packages/svelte/src/link.ts
+++ b/packages/svelte/src/link.ts
@@ -75,7 +75,11 @@ function link(
   const prefetchHoverEvents: ActionEventHandlers = {
     mouseenter: () => (hoverTimeout = setTimeout(() => prefetch(), config.get('prefetch.hoverDelay'))),
     mouseleave: () => clearTimeout(hoverTimeout),
-    click: regularEvents.click,
+    click: (event: PointerEvent) => {
+      clearTimeout(hoverTimeout)
+
+      regularEvents.click?.(event)
+    },
   }
 
   const prefetchClickEvents: ActionEventHandlers = {
--- a/packages/vue3/src/link.ts
+++ b/packages/vue3/src/link.ts
@@ -306,7 +306,11 @@ const Link: InertiaLink = defineComponent({
       onMouseleave: () => {
         clearTimeout(hoverTimeout.value)
       },
-      onClick: regularEvents.onClick,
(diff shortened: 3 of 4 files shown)


SuperDJ:
@lazerg adding this to all 3 adapters might increase the change of getting it merged. Also will increase the speed at which other adapters get the fix.

lazerg:
Good call, done in 9c68de7b. Vue and Svelte had the identical gap, their hover handlers passed `regularEvents.click` straight through with no `clearTimeout`, so the fix is the same shape in all three. I also dropped the react-only guard on the test, it now runs against every adapter and I checked it fails before the change and passes after on each one.

pascalbaljet:
Thanks!

PR #3239, opened 2026-08-31
[3.x] Avoid top-level await in the generated SSR bundle

ramonmalcolm10:
# Avoid top-level await in the generated SSR bundle

## Problem

The SSR templates in `packages/vite/src/frameworks/*` wrap the user's `createInertiaApp()` call in a **top-level await**:

```js
const render = await createInertiaApp({ ... })

const renderPage = (page) => render(page, renderToString)
```

Top-level await is ESM-only. Any bundle containing it can never be emitted as CommonJS, which rules out bundlers and runtimes that require CJS output.

The concrete case that led here: compiling the SSR bundle into a single-file executable with `bun build --compile --bytecode` (bytecode caching requires CJS). It fails outright:

```
$ bun build --compile --bytecode bootstrap/ssr/app.js --outfile ssr-server
6 | var render = await createInertiaApp({
                     ^
error: "await" can only be used inside an "async" function
    at bootstrap/ssr/app.js:6:20
```

The only workaround today is patching the generated bundle in the build pipeline before compiling it, which couples the deployment to the exact shape of the plugin's codegen.

## Change

Await the configure promise inside `renderPage` instead of at module scope:

```js
const renderPromise = createInertiaApp({ ... })

const renderPage = async (page) => (await renderPromise)(page, renderToString)
```

This is a drop-in change because both consumers already await the render result:

- dev server — `packages/vite/src/ssr.ts`: `const result = await render(page)`
- production — `@inertiajs/core`'s `createServer`: `const result = await render(page)`

Applied to all three framework configs (react, vue, svelte).

### Preserving fail-fast

With top-level await, a rejection from `createInertiaApp()` fails module evaluation and takes the process down at boot. Moving the await into `renderPage` would defer that to the first request, so the production branch attaches an explicit handler:

```js
if (import.meta.env.PROD) {
  renderPromise.catch((error) => {
    console.error(error)
    process.exit(1)
  })

  createServer(renderPage)
}
```

Happy to drop this if you'd rather keep the template minimal — the rejection would still surface, just on the first render instead of at boot.

## Verification

- `packages/vite`: all 53 tests pass; inline snapshots updated and `toContain` assertions adjusted to the new shape. (`tests/pages.test.ts` and `tests/ssr.test.ts` fail to collect on a clean checkout in my environment — unbuilt `@inertiajs/core/ssrErrors` — identical before and after this change.)
- End-to-end on a real Laravel + Inertia React 3.7 app (274 lazily-imported page chunks): built the plugin from this branch, ran `vite build --ssr`, then compiled the output with no patching:

  ```
  bun build --production --minify --bytecode --sourcemap --compile --target=bun \
    --define 'import.meta.env={"DEV":false,"PROD":true,"SSR":true}' \
    bootstrap/ssr/app.js --outfile ssr-server
  ```

  Bytecode generation succeeds, and the resulting standalone binary serves `/health` and renders pages correctly (verified against several routes, including lazily imported ones).

## Note on `--define`

That `--define` is needed for a second, unrelated ESM-only construct: `packages/react/src/createInertiaApp.ts` defaults `dev` to `!!import.meta.env?.DEV`, and the `import.meta` token alone forces ESM output. Since `@inertiajs/react` is published as ESM-only, that seems intentional, and `--define` is a fine consumer-side workaround — so this PR does not touch it. Happy to open a separate discussion if a dual ESM/CJS build is of interest.

### DIFF
--- a/packages/vite/src/frameworks/react.ts
+++ b/packages/vite/src/frameworks/react.ts
@@ -46,9 +46,7 @@ export const config: FrameworkConfig = {
 
   // SSR template that wraps the createInertiaApp call with server bootstrap code
   // Uses import.meta.env.PROD to skip the standalone server in dev mode
-  // The configure call is awaited per render instead of at the top level, which keeps the
-  // generated bundle free of top-level await and therefore compilable to CommonJS. A rejection
-  // is handled so it never takes the server down, and surfaces per render instead
+  // Awaited per render, not at the top level, so the bundle can compile to CommonJS
   ssr: (configureCall, options) => `
 import createServer from '@inertiajs/react/server'
 import { renderToString } from 'react-dom/server'
--- a/packages/vite/src/frameworks/react.ts
+++ b/packages/vite/src/frameworks/react.ts
@@ -46,22 +46,21 @@ export const config: FrameworkConfig = {
 
   // SSR template that wraps the createInertiaApp call with server bootstrap code
   // Uses import.meta.env.PROD to skip the standalone server in dev mode
-  // The configure call is awaited per render rather than at the top level, which keeps
-  // the generated bundle free of top-level await and therefore compilable to CommonJS
+  // The configure call is awaited per render instead of at the top level, which keeps the
+  // generated bundle free of top-level await and therefore compilable to CommonJS. A rejection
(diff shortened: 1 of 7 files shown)


pascalbaljet:
Thanks!

PR #3242, opened 2026-09-01
3.x] Add preventNavigationWhenDirty to useForm and <Form>

cwray-tech:
This PR adds an opt-in guard that warns users before they navigate away from a page with unsaved form changes.

Inertia already tracks `isDirty` on `useForm()` and `<Form>`, but nothing stopped a user from clicking a `<Link>` and losing their work. This wires dirty state into navigation blocking so apps get a standard “unsaved changes” flow without every team reimplementing it.

The guard lives in `@inertiajs/core` (`dirtyFormGuard`) and is exposed consistently across React, Vue, and Svelte.

## What end users get

- **Fewer accidental data loss moments.** If someone edits a profile, starts a long form, or tweaks settings and then clicks away, they get a chance to stay on the page instead of silently losing input.
- **A clear, consistent prompt.** Inertia visits show a confirm dialog with a sensible default message: *“You have unsaved changes. Are you sure you want to navigate away?”*
- **Protection when leaving the site entirely.** Tab close, refresh, and external navigations trigger the browser’s native `beforeunload` prompt when a guarded form is dirty (browsers show their own generic text there - custom copy isn’t supported on that path).
- **No friction when saving.** Submitting a dirty form does not show the confirmation dialog - the guard treats an in-flight submit as intentional navigation.

## API

**Per form (recommended):**

```js
// useForm
const form = useForm({ name: '' }).preventNavigationWhenDirty()
// optional custom message:
const form = useForm({ name: '' }).preventNavigationWhenDirty('Your draft will be lost.')

// <Form> component
<Form prevent-navigation-when-dirty ... />
<Form prevent-navigation-when-dirty unsaved-changes-message="..." ... />
```

**Globally (optional):**

```js
config.form.preventNavigationWhenDirty = true
config.form.unsavedChangesMessage = '...'
```

Per-form settings override the global defaults. The fluent `preventNavigationWhenDirty()` method on `useForm()` is the primary opt-in - same pattern as other form configuration helpers.

## Why this doesn’t break existing apps

- **Opt-in by default.** `config.form.preventNavigationWhenDirty` defaults to `false`. No form is guarded unless you explicitly enable it on that form or turn it on globally.
- **Listeners are lazy.** The `inertia:before` and `beforeunload` handlers are only registered while at least one guarded form is mounted. Apps that don’t use the feature pay no runtime cost.
- **Prefetch visits are ignored.** Hover/prefetch navigation is not blocked - only real navigations prompt the user.
- **Submitting still works.** While a form is submitting, it is not considered “dirty” for guard purposes, so save flows aren’t interrupted by a confirmation dialog.
- **No changes to existing form behavior.** Dirty tracking, validation, submission, reset, and error handling are untouched. This only adds an optional layer on top of existing `isDirty` state.
- **Adapter parity.** React, Vue, and Svelte all use the same core guard, so behavior is consistent and covered by shared Playwright tests.

## How this makes building web apps easier

Before this, blocking navigation on dirty forms meant wiring `router.on('before')`, tracking which forms were dirty, handling `beforeunload` separately, making sure submits were exempt, and duplicating that logic in every adapter. Easy to get wrong, and easy for one framework to behave differently from another.

## Implementation notes

- **Core:** `dirtyFormGuard` registers forms in a lightweight registry and hooks `inertia:before` to call `window.confirm()` when any registered form is dirty and not submitting.
- **Adapters:** `useForm` and `<Form>` register/unregister with the guard on mount/unmount and pass through optional custom messages.
- **Tests:** unit tests in `packages/core/tests/dirtyFormGuard.test.ts`; E2E coverage in `tests/form-helper.spec.ts` (dismiss/accept dialog, submit exemption, `<Form>` component).
- **Playgrounds:** demo pages added at `/form/prevent-navigation-when-dirty` in the Vue, React, and Svelte Laravel playgrounds for manual testing.

## Known limitations

| Scenario | Covered? |
|----------|----------|
| `<Link>` clicks, `router.visit()`, programmatic Inertia visits | Yes - via `inertia:before` + `window.confirm()` |
| Tab close / refresh / leaving the origin | Yes - via `beforeunload` (browser-controlled message) |
| Browser back / forward buttons | No - not reliably interceptable in SPAs |
| Client-side `router.push()` / `router.replace()` | No - these manipulate history without an Inertia visit |

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -75,6 +75,8 @@ export const config = new Config<InertiaAppConfig>({
     recentlySuccessfulDuration: 2_000,
     forceIndicesArrayFormatInFormData: true,
     withAllErrors: false,
+    preventNavigationWhenDirty: false,
+    unsavedChangesMessage: 'Changes you made may not be saved.',
   },
   prefetch: {
     cacheFor: 30_000,
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -76,7 +76,7 @@ export const config = new Config<InertiaAppConfig>({
     forceIndicesArrayFormatInFormData: true,
     withAllErrors: false,
     preventNavigationWhenDirty: false,
-    unsavedChangesMessage: 'Changes you made may not be saved.',
+    unsavedChangesMessage: 'You have unsaved changes. Are you sure you want to navigate away?',
   },
   prefetch: {
     cacheFor: 30_000,
--- a/packages/core/src/dirtyFormGuard.ts
+++ b/packages/core/src/dirtyFormGuard.ts
@@ -1,0 +1,84 @@
+import { config } from './config'
+import { eventHandler } from './eventHandler'
+import { GlobalEvent } from './types'
+
+export type DirtyFormEntry = {
+  isDirty: () => boolean
+  isSubmitting: () => boolean
+  message?: string
+}
+
+class DirtyFormGuard {
+  protected forms = new Set<DirtyFormEntry>()
+  protected removeBeforeListener: VoidFunction | null = null
+  protected beforeUnloadListener: ((event: BeforeUnloadEvent) => void) | null = null
+
+  public register(entry: DirtyFormEntry): VoidFunction {
+    this.forms.add(entry)
+    this.setupListeners()
+
+    return () => {
+      this.forms.delete(entry)
+      this.teardownListeners()
+    }
+  }
+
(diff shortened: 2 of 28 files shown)


pascalbaljet:
Thanks for the PR, but I don't think this belongs in the library though. The `before` event is already cancelable and `isDirty` / `processing` are public, so a dirty-form guard can be implemented in userland.

cwray-tech:
Appreciate it!

Issue #1915, opened 2024-07-11
Router only checks that `window.history.state` isn't empty, breaks when used with other frameworks


ariel-codes:
### Versions:

- `@inertiajs/core` version: 1.2.0
- `@inertiajs/vue2` version: #.#.#
- `@inertiajs/vue3` version: #.#.#
- `@inertiajs/react` version: 1.2.0
- `@inertiajs/svelte` version: #.#.#

### Describe the problem:

When navigating back/forward from pages, the router only checks that `window.history.state` is _not empty_, but this can conflict with other navigation frameworks (in my case Turbo/Hotwire, which uses `window.history.state.turbo`). This in turn leads to the `resolve` function passed to `createInertiaApp` being called with an undefined arg.

### Steps to reproduce:

- Set `window.history.state` to any truthy value before initializing Inertia

pedroborges:
@ariel-codes I'm curious about how you're integrating Inertia.js with Turbo/Hotwire, as they generally aim to solve similar problems but in different ways. It's my understanding that Turbo drives navigation and partial updates without the need for a front-end framework, while Inertia integrates tightly with frameworks like Vue, React, and Svelte.

Since both manipulate the `history.state`, it's understandable that there would be conflicts in scenarios where both are being used. As for Inertia, we might need a safeguard to confirm the presence of Inertia-specific data in `history.state` before continuing execution.

pascalbaljet:
For reference, see also PR #2238. I don't think we'll support other frameworks/libraries interfering with the history state.

pascalbaljet:
I'm closing this issue due to lack of response. If you're still encountering this and can provide additional details or steps to reproduce, feel free to open a new issue, preferably with a GitHub demo repo with just the essentials to trigger the problem. Thanks!

PR #3244, opened 2026-09-02
[3.x] Preserve URL changes made while the page is mounting

### LINKED ISSUE #3243: Initial page load's queued history write reverts URL changes made while the page is mounting

- [x] React
- [x] Vue 3
- [x] Svelte

### Adapter version

3.7.0 (present since 3.0.0; also reproducible on every 3.x release we tested, starting from 3.1.1)

### Describe the problem

During the initial page load, Inertia queues a history write for the current entry (`InitialVisit` → `page.set()` → `history.replaceState()`), and that write carries the **server-provided page URL** captured at enqueue time. The queue flushes asynchronously (component resolution, `structuredClone` of props, optional history encryption), so there is a window — the entire mount of the initial page — during which any URL change made through the History API is silently reverted once the queued write flushes.

This breaks the common pattern of an app (or a third-party library) syncing state into the query string while the page mounts:

```js
// e.g. an onboarding page canonicalizing its step param on mount
window.history.replaceState(window.history.state, '', '/onboarding?step=1')
```

Two things go wrong when the queued initial write flushes:

1. The address bar is reverted to the server URL (`/onboarding`), dropping the query param.
2. Because the modified URL no longer matches `page.url`, `isSameUrlWithoutHash()` is `false`, so `page.set()` takes the **`pushState`** branch — the initial page load adds a *second* history entry, breaking `history.back()`.

The race is timing-dependent (it hinges on how long the mount-time queue takes to drain), so in real apps it often shows up as flaky behavior — e.g. it reliably reproduces for us on loaded CI machines and intermittently on slower devices.

This is the same bug class as #2233, where Inertia's *scroll-position* writes reverted externally-added query params. That was fixed in #2280 by omitting the URL from those writes. The initial-visit write is the remaining URL-carrying async write that can clobber the address bar.

### Steps to reproduce

1. Create a page component that adds a query param via the History API while mounting (before Inertia's initial visit settles):

```vue
<script setup>
window.history.replaceState(window.history.state, '', '/example?step=1')
document.addEventListener('inertia:navigate', () => {
  console.log(location.search, history.length)
}, { once: true })
</script>
```

2. Load `/example` fresh (full page load).
3. When `inertia:navigate` fires (initial visit settled), observe:
   - **Expected:** `location.search` is `?step=1`, no extra history entry.
   - **Actual:** `location.search` is empty (URL reverted to `/example`), and `history.length` grew by one.

The synchronous-mount timing (Vue/Svelte `script setup` / component body) makes this reproduce deterministically; with `useEffect`-style timing it becomes the intermittent flake described above.

I have a fix ready — the initial render's history write always replaces and omits the URL (mirroring the #2280 approach), so it only stores the page state on the existing entry. PR incoming with a regression test across the three adapters.

mochetts:
Fixes #3243

## Problem

During the initial page load, `InitialVisit` → `page.set()` queues a history write that carries the **server-provided page URL** captured at enqueue time. The queue flushes asynchronously (component resolution, `structuredClone` of the props, optional history encryption), so any URL change made through the History API while the page is mounting — a query param synced by the app on mount, or one added by a third-party library — is silently reverted once the write flushes.

It gets worse: because the modified URL no longer matches `page.url`, `isSameUrlWithoutHash()` returns `false` and `page.set()` takes the `pushState` branch, so the initial page load also adds a **duplicate history entry**, breaking `history.back()`.

Since the race depends on how long the mount-time queue takes to drain, in real apps this surfaces as flaky behavior — for us it reliably reproduced on loaded CI machines (an onboarding page normalizing `?step=1` on mount) and intermittently on slow devices.

This is the same bug class as #2233, where the *scroll-position* writes reverted externally-added query params, fixed in #2280 by omitting the URL from those writes. The initial-visit write is the remaining URL-carrying async write that can revert the address bar.

## Fix

Two changes to `page.set()`, applied only to the initial render (`initialRender: true`, passed by `InitialVisit.handleDefault` / `handleLocation`):

1. **Always replace.** The browser already has a history entry for the requested URL, so pushing can only ever create a duplicate entry. (Previously `replace` was inferred from URL equality, which is exactly what broke when the URL had legitimately changed during mount.)
2. **Omit the URL from the write.** The initial write's purpose is to store the page object on the *existing* entry so back/forward restores work — not to navigate. `history.replaceState()` now accepts a `preserveCurrentUrl` option that skips passing a URL to `window.history.replaceState()`, keeping whatever is in the address bar by the time the queued write flushes. This mirrors what #2280 did for the scroll-position writes.

On a normal initial load (nobody touched the URL) the written entry is byte-for-byte what it was before — the address bar already shows `page.url` — so behavior is unchanged. Only when the URL was deliberately changed during mount does the outcome differ: the change now survives, and no duplicate entry is pushed.

## End-user benefit

Apps can keep using the documented escape hatch of the History API for cosmetic query-string updates (analytics params, step/wizard params, filter state) without their URL being nondeterministically reverted on page load, and without a stray extra history entry breaking the back button. This class of bug is particularly nasty to diagnose because it only loses the race on slow machines — several of us have chased it as CI-only test flakiness.

## Tests

Added a `UrlOnMount` page to all three test apps plus a shared Playwright test. The page adds `?step=1` via `history.replaceState` synchronously while mounting, then reports `location.search` and the `history.length` delta once the initial visit settles (`inertia:navigate`). The test asserts the param survives and no extra entry was pushed.

Before the fix (Vue/Svelte, where mount timing is synchronous and the race is lost deterministically):

```
1) [chromium] › tests/initial-visit.spec.ts › it preserves URL changes made via the History API while the page is mounting
   Expected string: "?step=1"
   Received string: ""
```

After the fix: passes on all three adapters, and the full suite is green (`pnpm test:vue`: 1190 passed / 18 skipped; the history/scroll/manual-visits/client-side-visits/events suites also pass on React and Svelte).

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -163,7 +163,11 @@ class History {
     return window.history.state?.documentScrollPosition || { top: 0, left: 0 }
   }
 
-  public replaceState(page: Page, cb: (() => void) | null = null): void {
+  public replaceState(
+    page: Page,
+    cb: (() => void) | null = null,
+    { preserveCurrentUrl = false }: { preserveCurrentUrl?: boolean } = {},
+  ): void {
     if (isEqual(this.current, page)) {
       cb && cb()
       return
@@ -189,7 +193,14 @@ class History {
       return this.getPageData(page).then((data) => {
         // Defer history.replaceState to the next event loop tick to prevent timing conflicts.
         // Ensure any previous history.pushState completes before replaceState is executed.
-        const doReplace = () => this.doReplaceState({ page: data }, page.url).then(() => cb?.())
+        //
+        // With `preserveCurrentUrl` the write only stores the page state on the
+        // existing entry: omitting the URL keeps whatever is in the address bar by
+        // the time this queued write flushes, so URL changes made via the History
+        // API while the page was mounting (query params added by the app or a
+        // third-party library) are not reverted to the server-provided URL.
+        const doReplace = () =>
+          this.doReplaceState({ page: data }, preserveCurrentUrl ? undefined : page.url).then(() => cb?.())
 
         if (isChromeIOS) {
           return new Promise((resolve) => {
--- a/packages/core/src/page.ts
+++ b/packages/core/src/page.ts
@@ -96,13 +96,17 @@ class CurrentPage {
       const isServer = typeof window === 'undefined'
       const location = !isServer ? window.location : new URL(page.url)
       const scrollRegions = !isServer && preserveScroll ? Scroll.getScrollRegions() : []
-      replace = replace || isSameUrlWithoutHash(hrefToUrl(page.url), location)
+      // The initial render must always replace: the browser already has an
+      // entry for the requested URL, so pushing would add a duplicate one.
+      replace = replace || initialRender || isSameUrlWithoutHash(hrefToUrl(page.url), location)
 
       // Clear flash data from the page object, we don't want it when navigating back/forward...
       const pageForHistory = { ...page, flash: {} }
 
       return new Promise<void>((resolve) =>
-        replace ? history.replaceState(pageForHistory, resolve) : history.pushState(pageForHistory, resolve),
+        replace
+          ? history.replaceState(pageForHistory, resolve, { preserveCurrentUrl: initialRender })
+          : history.pushState(pageForHistory, resolve),
       ).then(() => {
         const isNewComponent = !this.isTheSame(page)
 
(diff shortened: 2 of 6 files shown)


pascalbaljet:
Inertia assumes it's the only thing managing the browser history state, and I've decided against opening that up before in #2238 and #1915. Manual `history.replaceState()` calls also don't update `page.url`, so the param would survive the initial load but disappear again on the next `router.reload()` or partial reload, and the address bar and `usePage().url` disagree in the meantime.

For syncing state into the query string, use a client-side visit instead. That updates the address bar and `page.url` together:

```js
router.replace({
    url: '/onboarding?step=1',
    preserveScroll: true,
    preserveState: true,
})
```

PR #3248, opened 2026-09-07
[3.x] Fix multi-byte UTF-8 corruption in SSR request body parsing

onk:
Both the Vite SSR dev endpoint (packages/vite/src/ssr.ts) and the production SSR server (packages/core/src/server.ts) read the incoming request body with `data += chunk`. Since `chunk` is a Buffer, this decodes each chunk independently, so a multi-byte UTF-8 character split across a chunk boundary gets replaced with U+FFFD on both sides of the split.

~~Fixed by buffering the raw chunks and decoding once via Buffer.concat(...) .toString('utf8') after the stream ends.~~

Fixed by setting the request stream encoding to UTF-8 via `setEncoding('utf8')`. Node.js then handles multi-byte characters that span chunk boundaries correctly while the body can continue to be accumulated as a string.

## Repro

I split this into two commits. The first commit contains the fix, and the second commit contains tests that reproduce the bug through both public SSR entry points.

The tests need to mock `http.createServer` and the Vite dev middleware to reach `readRequestBody` / `readableToString`, so they are a bit complicated. I mainly added them to make the bug easy to reproduce and verify.

If the tests feel too heavy to keep, feel free to merge only 480508416e .

### DIFF
--- a/packages/core/src/server.ts
+++ b/packages/core/src/server.ts
@@ -65,6 +65,7 @@ type Port = number
 const readableToString: (readable: IncomingMessage) => Promise<string> = (readable) =>
   new Promise((resolve, reject) => {
     let data = ''
+    readable.setEncoding('utf8')
     readable.on('data', (chunk) => (data += chunk))
     readable.on('end', () => resolve(data))
     readable.on('error', (err) => reject(err))
--- a/packages/core/src/server.ts
+++ b/packages/core/src/server.ts
@@ -64,9 +64,9 @@ type Port = number
 
 const readableToString: (readable: IncomingMessage) => Promise<string> = (readable) =>
   new Promise((resolve, reject) => {
-    let data = ''
-    readable.on('data', (chunk) => (data += chunk))
-    readable.on('end', () => resolve(data))
+    const chunks: Buffer[] = []
+    readable.on('data', (chunk: Buffer) => chunks.push(chunk))
+    readable.on('end', () => resolve(Buffer.concat(chunks).toString('utf8')))
     readable.on('error', (err) => reject(err))
   })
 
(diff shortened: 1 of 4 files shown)


mokevnin:
We hit this too (Rails + `inertia_rails`, SSR enabled), and can confirm it still reproduces on the latest release, 3.7.1.

Minimal repro against an unpatched `@inertiajs/core@3.7.1`: start the SSR server, then POST a ~24KB body containing multi-byte characters, splitting it across two TCP writes in the middle of a multi-byte sequence. The rendered output comes back with `U+FFFD` on both sides of the split:

```
split at byte: 12032 of 24061
U+FFFD in SSR response: 2
```

What made this annoying to track down is that it only shows up on longer pages — short bodies arrive in a single chunk, so everything looks fine in development and in tests.

In the meantime we're carrying a local patch through pnpm's `patchedDependencies`, on both `@inertiajs/core` (`dist/server.js`) and `@inertiajs/vite` (`dist/index.js`). Since the patch keys are pinned to an exact version, it has to be regenerated on every bump of `@inertiajs/*`, which is exactly the kind of thing that eventually gets dropped by accident — and the breakage is silent.

We went with `readable.setEncoding('utf8')` rather than buffering the chunks, since Node's `StringDecoder` then holds the partial sequence across chunk boundaries; functionally equivalent to what this PR does. Either way, it'd be great to see this land so we can drop the patch. Thanks for writing it up with tests.

onk:
Oh, I think `readable.setEncoding('utf8')` would be the correct fix here.

[`readable.setEncoding(encoding)`](https://nodejs.org/api/stream.html#stream_readable_setencoding_encoding) is specifically designed to decode a readable stream while correctly handling multibyte characters that span chunk boundaries.

It also avoids having to buffer all chunks as `Buffer`s and concatenate them before decoding.

onk:
To maintainers:

You only need to cherry-pick 480508416e .

pascalbaljet:
Thanks, all!

PR #3259, opened 2026-09-14
[3.x] Support externally managed navigation and app disposal

danharrin:
This adds an optional way to use Inertia in an application that already handles its own navigation. It would let applications migrate individual pages to Inertia without first replacing their router or application shell.

For example, an application could start using React, Vue, or Svelte for a page, with Inertia handling server-driven props and forms, while keeping its existing navigation between pages. More pages could move to Inertia over time.

The immediate use case is Filament, where Livewire handles the application shell and navigation. This would let developers build a Filament page with Inertia while retaining that shell. There is no dependency on Filament or Livewire in the implementation.

### How it works

The optional `externalNavigation` option accepts three hooks:

- `navigate(url)` hands navigation outside the current Inertia page to the host.
- `remember(data, key)` and `restore(key)` optionally let the host manage remembered state.

The host handles document navigation, browser history, and document scrolling. Inertia still handles requests within the page, including forms, validation, deferred and optional props, partial reloads, and client-side prop updates.

Normal links and responses that navigate to another page are handed to the host. Same-page updates stay within Inertia. Same-path GET partial reloads can use query parameters without changing the host URL, allowing features such as infinite scrolling to work within the page.

For ordinary destination responses, flash and success/error callbacks run before the handoff. The host receives a URL rather than the response props. A server location response is appropriate when transient data needs to remain available for the destination's subsequent request.

Automatic external mounts return a disposer. Custom Vue and Svelte mounts use their normal unmount lifecycle. Custom React setup also receives a disposer for the router scope.

### Lifecycle changes

The host needs to be able to remove an Inertia page and mount another without reloading the document. That requires cleaning up the outgoing page's work as well as delegating navigation.

Requests, queued responses and client visits, prefetches, polling, and asynchronous component resolution are tied to the active router lifetime. Checks after asynchronous work and callbacks prevent an outgoing operation from updating a replacement mount. Disposal also removes owned head elements and prevents pending head updates from restoring them. Fresh external mounts reset dynamic layout props.

React needs additional handling because the router initializes before effects commit. A suspended or abandoned render may never run effect cleanup, so the external router scope is owned outside the effects. Effect attachment and detachment are separate from disposal so StrictMode can detach and reattach without destroying the router.

The implementation keeps Inertia's single-active-router model. It does not add support for multiple independent routers in one document.

### SSR root IDs

There is also a small, separate correction to `CreateInertiaAppOptionsForSSR`: its `id` option now accepts a string rather than only `undefined`.

React and Vue already use a supplied ID when rendering the SSR root, but the declaration rejected it. This became relevant when using a host-specific root instead of the default `app` element. The correction changes no runtime behavior and includes checks in both adapters' existing TypeScript fixtures. It could be split into a separate PR if preferred.

### Compatibility and tests

External navigation is opt-in. The intention is to preserve existing Inertia navigation without requiring a major version.

The browser tests use the existing adapter test applications. They cover navigation handoffs, forms, props, repeated mounts, cleanup, and React suspension. The focused suites pass across all three adapters, as does the combined React native-polling and external-navigation suite. The React and Vue TypeScript checks also pass with custom SSR root IDs.

### A note on implementation

I'm not familiar with Inertia's core internals, and AI helped substantially with the implementation, tests, and review iterations. I wanted to disclose that given the breadth of the lifecycle changes.

I'd appreciate feedback on whether this approach fits Inertia, whether the API belongs at this level, and whether I've missed existing mechanisms that should be used instead. I'm happy to revise the implementation or split the changes differently to make them easier to review and maintain.


packages/svelte/src/components/Render.svelte
danharrin, line 1: I checked the built package in a strict Svelte consumer with `skipLibCheck: false`: it failed with the duplicate `Render` declaration before this change and passed afterward.

PR #3262, opened 2026-09-16
[3.x] Forward missing Form visit callbacks

daleweaver777:
`Form` does not forward `onHttpException`, `onNetworkError`, `onBeforeUpdate`, or `onFlash` to the underlying visit. These callbacks are part of `FormComponentProps`, but are missing from the adapter implementations. For example, an HTTP 500 or a failed network request never invokes the callback passed to `<Form>`, so applications cannot handle these failures locally as they can with `useForm`.

This forwards all four callbacks in React, Vue, and Svelte, preserving their arguments and return values. Returning `false` from either error callback now suppresses the corresponding global event and default handling. Existing success, validation-error, reset, and completion behavior is preserved.

Adds five shared regression tests covering HTTP and network error arguments, allowing/suppressing global error events, incoming page data, flash data, and completion. All five reproduce the missing callbacks against the original React implementation and pass with the fix across all three adapters.

Validation:

- `pnpm build:all`
- `pnpm type-check:test-app`
- `pnpm lint:test-app`
- Formatting checked for all changed files.
- Full `tests/form-component.spec.ts` Chromium suites: React 104 passed; Vue 103 passed, 1 skipped; Svelte 103 passed, 1 skipped.


pascalbaljet:
Thanks!

